PR #479, opened 2012-07-01
Mailing lists

t-8ch:
This allows nice list replies. It assumes, that the mailing list sets the `Reply-To` header.
The Mail is sent to the value of `Reply-To` and `From` is added to `Cc`.

There is also early list autodetection and a hook for fine grained control. And of course documentation.

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -73,26 +73,30 @@ def recipient_to_from(mail, my_accounts):
 
 
 @registerCommand(MODE, 'reply', arguments=[
-    (['--all'], {'action':'store_true', 'help':'reply to all'}),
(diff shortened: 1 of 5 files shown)


GuillaumeSeren:
Hi there,
I find this branch, and rebased it, it work well for my use,
and I will be happy to clean it up a bit more, and submit it if some of you are interested.

I agree the feature might be simple and keep the (L) mutt behavior,
in the rare case of adding the sender in copy, (which is seen as bad by most ML, but the sender can ask for it if he is not a subscriber) we can 'reply --all'.

pazz:
@GuillaumeSeren : sure, feel free to send a cleaned up PR.

GuillaumeSeren:
@pazz Ok I will clean up and then submit, thank you for reply.

PR #533, opened 2012-11-06
show threads and unread in tag list

k00n:
The tag list widget now shows the number of threads and number of unread
threads of each tag next to its name.


pazz:
I'm retiring this issue.

PR #612, opened 2013-05-23
Color html rendering

cagprado:
Put some colors in mail body. This should work in case you use color dump with elinks or other program in mailcap for html rendering. It translates the ANSI escapes to Urwid attributes object and applies them while in thread mode. Also it strips them out on replies, forwards, bounces, write as new and pipeto (--format=decoded) commands.

It assumes no extra configuration in alot files: i.e. user would decide if he wants it by simply adding color dump in mailcap.

Works with 8 color escapes or 256 color escapes. Also bold, underline and reverse attributes. Ignores all other escapes. It also honors the cumulative behaviour of ANSI escapes which is different from how Urwid works (i.e. turn on bold… it will be turned on until it is turned off =P).

Invalid escapes like '\e[38;233' instead of '\e[38;5;233' will probably raise an exception (out of bounds indexing), however I don’t know if it is something that 'alot' would care about. ;)

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -627,7 +627,8 @@ class PipeCommand(Command):
                     pipestrings.append(mail.as_string())
                 elif self.output_format == 'decoded':
                     headertext = extract_headers(mail)
-                    bodytext = extract_body(mail)
+                    # Get bodytext from accumulate_body which strips ANSI esc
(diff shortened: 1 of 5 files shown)


pazz:
closing this in favour of #1015. thanks!

PR #617, opened 2013-05-26
make tag commands undoable

ignatz:
First of all, i do not consider this patch ready to merge, however I wanted to kick of some discussion about an undo mechanism.

Tackling this issue in general is far from trivial and would also require rethinking the whole command mechanism. Inverse operations aren't always obvious/possible and the concept of command lines makes it even harder (e.g. a long command line could include a single non-reversible command).
But what can easily be realized is the undoing of _individual_ tag commands, which doesn't handle whole command lines. Yes, this feels like a hack, but the possibility to at least undo tagging is something I desperately like to have.

Apart from that, the implemented mechanism  does offend existing code, by registering the very same command in both search and thread mode, albeit implemented only in thread. One would need to duplicate the implementation to match the existing code/policy. 
Currently, similar but independent TagCommands exist in search as well as tag mode. The reason for that -as far as I understand it- is, that each command is individually responsible for choosing it's scope to act on. If modes would be responsible for scoping, rather than commands, they could also handle ui refreshes, etc... and all commands could drop the corresponding code. Furthermore, commands like TagCommand could be reused in search and thread mode or maybe I got it all wrong ;-). Even though, beyond this pull request, this is more about kicking  off some discussion. Did you ever consider such a change and is this something you would be interested in? What do you think about undoing in general and in particular? What is your opinion on reusing the UndoTagCommand in thread and search mode?

### DIFF
--- a/alot/commands/search.py
+++ b/alot/commands/search.py
@@ -7,6 +7,7 @@ import logging
 from alot.commands import Command, registerCommand
 from alot.commands.globals import PromptCommand
 from alot.commands.globals import MoveCommand
+from alot.commands.thread import UndoTagCommand
 
 from alot.db.errors import DatabaseROError
 from alot import commands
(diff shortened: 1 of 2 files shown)


np:
Can this be considered again? Undo would be really nice to have.

pazz:
I'm putting this one on hold for now

Issue #560, opened 2013-02-14
regularly poll for new messages


josch:
Hi,

from the documentation I can't make out how to regularly poll for new messages.

I'd like to run offlineimap and notmuch new every X minutes. Is this possible from within alot or do I have to set up a cron job for it?

cheers, josch

PR #633, opened 2013-07-03
Retain message position when refreshing search buffer, and regularly poll for new messages (#560)

witten:
This change prevents the message cursor from jumping to the top when refreshing the search buffer (unless the message disappears during the refresh), and adds periodic refreshing of search buffers so that new threads show up.

Not sure if you want to take this approach, but I figure this pull request will at least get the discussion started.


pazz:
I'm retiring this PR for the above discussed shortcomings of the lib. Feel free to re-open if you can think of a possible workaround that is stable enough for the main branch.

PR #648, opened 2013-08-13
Handle multiple directories when searching for the theme

laarmen:
This patch introduces support for the "/foo/bar:/foo/baz" syntax for the
configuration key "themes_dir", and make the search go through thoses
first, then try the default dir, and only then fails if no theme has
been found.
## 

The $PATH-like syntax is a bit extra, I figured that since I was using lists — which are needed for an Debian-specific patch to also look up theme in a system-wide dir.


pazz:
closing this for lack of activity. The "problem" remains unsolved: I'd be nice to have a way to keep theme files system wide.

Issue #661, opened 2013-10-20
crypto: encrypt email by default if is a reply to an encrypted email


meskio:
I think that if the email I'm replying to was encrypted alot should encrypt the reply by default. Too often I send by mistake the body of an encrypted emails as plain text when I reply to them.

Other clients like mutt or thunderbird-enigmail do that.

lamarpavel:
According to #774 this has been implemented and merged into master, I think this issue could be closed.

pazz:
thanks for pointing this out!

PR #675, opened 2013-11-19
Encrypt on reply to encrypted email

meskio:
For issue #661.

It fix the issues of the first pull-request (#669). Now the check for keys is done at soon as it enters on the envelop buffer.

I don't know much of urwid, I hope I'm not doing any mistake.


meskio:
I'm closing this pr, as there is a follow up in #766.

I think your comment about signatures should be a separated pull-req.

PR #696, opened 2014-03-06
On exit command, check if there are any unsent messages

foobacca:
This commit doesn't actually work, though it looks like it should.  I've stepped through the debugger, and it does get to the line with:

```
if (yield ui.choice('quit without sending message?',
                    select='yes', cancel='no',
                    msg_position='left') == 'no'):
```

but the message never actually appears.  When I get to this line it goes off somewhere into some defer code and eventually the application quits without ever showing the 'quit without sending message?' prompt.

Any pointers as to how to get this to work appreciated - I've lost a few messages due to quitting when I have an unsent message.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -42,6 +42,16 @@ class ExitCommand(Command):
     @inlineCallbacks
     def apply(self, ui):
         msg = 'index not fully synced. ' if ui.db_was_locked else ''
+
+        # check if there are any unsent messages
+        for buffer in ui.buffers:
+            if (isinstance(buffer, buffers.EnvelopeBuffer) and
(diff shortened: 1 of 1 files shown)


geier:
I can confirm foobacca's observation, jump to some defer in twisted an exists from there (without querying the user).

geier:
I believe I found the error:
the diff should be
```python
                if (yield ui.choice('quit without sending message?',
                                    select='yes', cancel='no',
                                    msg_position='left')) == 'no':
                    raise CommandCanceled()
```
note the closing parenthesis is before the ``==``.

I believe this would make a good addition to alot.

dcbaker:
#948 Is a rebased version of this, and has been merged

PR #709, opened 2014-06-29
use setuptools

posativ:
Hi,

I made some changes so a `python setup.py install` will now install all required dependencies (except for the broken PyGPGME). An installation will no longer replace alot executable, because it's now been generated by setuptools.

With setuptools you may also install alot in develop mode, which install all dependencies and links the package for local development:

``` sh
$ python setup.py develop
```

Preferably in a virtual environment.


pazz:
fyi: a rebased version of this is in `testing` now

PR #724, opened 2014-08-22
Exchange & Outlook display blank emails

krakensden:
For whatever reason, it looks like Exchange Webmail and Outlook don't like plain-text emails with \r\n newlines- they will display them as 'blank' messages.

I don't really expect you to merge this- it subverts 90% of the `RFC3156_canonicalize` function- but I'd like you to be aware of the issue, and make it easy for other people to find the problem.

I don't really know if there's a good solution without pulling SMTP functionality inside of alot, unfortunately.

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -610,9 +610,9 @@ def RFC3156_canonicalize(text):
     :param text: text to canonicalize (already encoded as quoted-printable)
     :rtype: str
     """
-    text = re.sub("\r?\n", "\r\n", text)
-    if not text.endswith("\r\n"):
-        text += "\r\n"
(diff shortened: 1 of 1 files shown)


pazz:
thanks for mentioning this issue (ans sorry for my late reply).
As you say, I can't really merge this in as it would break the encrypting/signing feature for sure.
I doubt that this has much to do with smtp and instead suspect that there is either an error in outlooks mail viewer (not so unlikely) or some fishy translation going on in your sendmail setup.
Either way, I'll close this issue here, but it will surely pop up if someone searches for this issue.
Cheers!

krakensden:
That's fair :).

PR #728, opened 2014-09-28
Templates improvements

laarmen:
I started out just wanting to send out a git patch on a mailing list. The end result is a patchset allowing for in-place template files, and a parser that is probably more robust (at least it doesn't choke on git patches :-) )

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -10,6 +10,7 @@ import email
 import tempfile
 from twisted.internet.defer import inlineCallbacks
 import datetime
+from StringIO import StringIO
 
 from alot.account import SendingMailFailed, StoreMailError
 from alot.errors import GPGProblem, GPGCode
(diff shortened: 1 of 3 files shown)


pazz:
@laarmen could you clean up the commit messages and maybe add some more inline docs to explain what the change is? If you like you can document your use case separately in the wiki or markdown docs. If you're quick I'll add this to the upcoming release :)

pazz:
How do we stand on this? I think generally is a good idea and its easy to parse at least the top level headers into an envelope object. Where it gets tricky itemplte is a more complex multipart mail with attachments etc: what if you want to still change the "body" text? We could extract a body text as we do in threads but then the msg later generated for snding might have a completely different structure..

lucc:
@pazz if we can display the mime tree of the imported message in the envelope buffer the user could select which part they want to edit. Then the original structure could be preserved.

There are two problems with this though:
1. we need a way to display the mime tree of a message and select a line from that in the user interface. This is where I failed in #894.
2. We need a message object in the envelope. As far as I remember the envelope only holds the text and some metadata at the moment.

pazz:
Closing this here as part of my pre-release spring cleaning, based on a general lack of activity..

PR #729, opened 2014-10-02
Allow configuring what does alot consider to be the thread subject.

elenril:


pazz:
this was merged long ago and I forgot to close the issue

PR #732, opened 2014-11-10
Honor notmuch-config search.exclude_tags value

nicoe:


pazz:
there is a much cleaner solution available now: the notmuch bindings
now (e13d557bfdf3f07f16e051430056c57e574b650d) expose the
`notmuch_query_add_tag_exclude` method directly. So we can just
use it to add exclude tags whenever we instanciate a quary object,
and it should guarantee the same behaviour as notmuch cli..

pazz:
@bremner 
yes, as mentioned in my comment above :)
There is actually only one position where alot reads notmuch's config directly, and it is to find out the path to the database. IIRC, the bindings do not default to the path specified in the (notmuch)config
but hardcode `~/.notmuch-config`.
I agree that it'd be nicer if alot didn't read the notmuch config but i don't see a nice (calling `notmuch config` is not nice) alternative..

pazz:
closing in favour of #1165

PR #733, opened 2014-11-18
New option `thread_authors_order_by` to control author order in thread view

tomyl:
Default value 'first_message' lists authors in the order they joined the
conversation. Value 'latest_message' order authors by their latest
message, which makes it easier to see which authors who wrote the most
recent messages.

Note that authors with duplicate emails were previously filtered. We now
keep all authors where name + email is unique. This is behavior is
desired for some email notification services. For example, Jira will
set the From header to "Joe User jira@company.com" meaning you will
only see one author if you only keep unique email addresses.


pazz:
pushed

Issue #695, opened 2014-03-05
add options to improve message indenting in long threads


xunam:
In a long thread, the display of messages and headers in thread view becomes problematic: say you have a terminal with standard width of 80 characters and a thread with 30 successive replies, with 2 characters of indenting for each message you end up with a width of only 20 characters for actually displaying a message. Over 40 messages, the display gets messed up as Urwid does what it can to fit long lines in its display (worst case: a whole message displayed with a width of 1 character).

I don't have definitive solutions to suggest, but a few (not mutually exclusive) possiblities might improve things:
- add an option that would allow message headers and bodies to not be indented at all (with a suitable theme, the distinction between thread structure and message bodies would still be visible)
- allow for horizontal scrolling (or shifting everything to the left, when focusing on late messages in a long thread)
- do not indent systematically but only when a given message has at least two replies: this way a long thread with no "branching" would appear linear, which would be more readable (in my opinion)

Issue #694, opened 2014-02-27
High CPU usage and slow UI performance


ghost:
Hi,
I noticed very high cpu usage when I scroll mail list or thread. Cpu usage peaks up to 70-100% (especially in thread view). Tested on release 0.3.5 and newest revision from git.

How to reproduce: open some thread, press _j_ and look at cpu usage...

GuillaumeSeren:
Hi there,
I agree I get the same behavior here on Debian SID,
on big threads maybe it come from the tree view but I didn't see any difference folded or not.

It can come from this version, I will try to upgrade from git.

josch:
I uploaded an xz-compressed cprofile output from me scrolling through a very long thread: https://mister-muffin.de/p/_83G

Scrolling was very sluggish and slow and the Python process consumed 100% of my CPU during scrolling.

Here is the top of the pstats output in case it helps:

```
         406159648 function calls (391120742 primitive calls) in 393.638 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.006    0.006  393.640  393.640 alot/__main__.py:4(<module>)
        1    0.000    0.000  393.086  393.086 alot/__main__.py:20(main)
        1    0.000    0.000  380.642  380.642 alot/ui.py:33(__init__)
        1    0.000    0.000  380.583  380.583 /usr/lib/python2.7/dist-packages/urwid/main_loop.py:268(run)
        1    0.000    0.000  380.583  380.583 /usr/lib/python2.7/dist-packages/urwid/main_loop.py:366(_run)
        1    0.000    0.000  380.582  380.582 /usr/lib/python2.7/dist-packages/urwid/main_loop.py:1174(run)
        1    0.000    0.000  380.582  380.582 /usr/lib/python2.7/dist-packages/twisted/internet/base.py:1197(run)
        1    0.290    0.290  380.581  380.581 /usr/lib/python2.7/dist-packages/twisted/internet/base.py:1202(mainLoop)
     5718    0.019    0.000  365.720    0.064 /usr/lib/python2.7/dist-packages/urwid/main_loop.py:1197(wrapper)
   231835    0.283    0.000  319.094    0.001 /usr/lib/python2.7/dist-packages/twisted/internet/base.py:790(runUntilCurrent)
     2516    0.191    0.000  318.749    0.127 /usr/lib/python2.7/dist-packages/urwid/main_loop.py:566(draw_screen)
[...]

PR #734, opened 2014-11-30
Retro Thread viewer!

andresmrm:
Hi, Pazz! You've made a quite impressive MUA! Thanks!
I'm comming from Mutt and found in Alot a lot of features that I missed in Mutt. ;)

But I found the current thread viewer a bit problematic with big threads, and saw that I'm not the only one:
https://github.com/pazz/alot/issues/694
https://github.com/pazz/alot/issues/695
So I made a "Retro Thread" (RT) viewer, that somehow replicates Mutt behavior: with a small viewer for the thread tree and a big message viewer. 

The code is a big mess and not finished, so I'm not sure I should be sumitting a pull request, but it seems usable here and faster in big threads than the current one.
I tried to place most of the code related with the RT in a sepate file, to don't "corrupt" your code. =P
But had to add a command (in `alot/commands/search.py`) to enter this "alternative" viewer, and a binding group (in `alot/commands/__init__.py`) for the bindings of the new viewer.
(Besides that there are 2 typo-corrections in this pull)

To use the code, you'll probably need to add a few binding in the config file.
In **[[search]]** add:

```
<some key> = RTselect
```

And than the new bindings group, something like:

```
[[RTthread]]
    l = move next
    h = move previous
    j = move down
    k = move up
    z = move switch
    r = reply
```

Best regards!


andresmrm:
Most of this functionality is added by #896, and I think that code is better written. So closing this PR in favor of that one.

PR #738, opened 2014-12-09
Option to show all tags of a message, even if they are commom to thread

andresmrm:
Setting ´msg_summary_hides_threadwide_tags´ to False in the configs allows to show all tags of a message in the end of the message´s summary, in Thread Buffer.


pazz:
merged

kirelagin:
This behaviour was driving me completely crazy until I noticed the pattern. I’m wondering if `False` would be a better default for this option as less surprising for new users :/.

PR #739, opened 2014-12-09
config to allow specific command for when editing a mail the first time

andresmrm:
New config ´new_file_editor_cmd´ that allows to set the command to be used when editing a mail for the first time (reply, forward, ...). The ´editor_cmd´ will be used only in the Envelope Buffer, after you openned and closed the mail a first time.
If ´new_file_editor_cmd´ is unset (default) uses the old behavior.

Why all this?

For exemple:

```
new_file_editor_cmd = vim -c 'exe "norm }O\n\n\<esc>O"' -c 'startinsert'
editor_cmd = vim -c 'exe "norm }"'
```

Will open some space between the headers and the quoting part, and enter insertion mode in Vim. That's something I always have to do when editing a new mail. But I don't want it to happen twice in the same mail (for example, when I'm in Enevelop Buffer, forgot something in my mail, and I'm going to edit it for the second time).

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -285,7 +285,7 @@ class SendCommand(Command):
                      'help': 'refocus envelope after editing'})])
 class EditCommand(Command):
     """edit mail"""
-    def __init__(self, envelope=None, spawn=None, refocus=True, **kwargs):
+    def __init__(self, envelope=None, spawn=None, refocus=True, new_file=False, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


lucc:
I think this is a very special corner case that is better handled in your vim config than in alot. You can have a look at my mail filetype code for ideas. I implemented something similar: https://github.com/lucc/vim-config/blob/develop/ftplugin/mail.vim#L11-L18

PR #740, opened 2014-12-09
Retro thread buffer

andresmrm:
Still needs more refactoring...


binaryplease:
Hi, 
@pazz whats the current status here?
Seems like @andresmrm added some commits cleaning up.
 Is this ready to merge, what is missing?

pazz:
@binaryplease you are welcome to use this, but what keeps me from merging this are two things:
1. there are two many "back and forth" patches, where things are introduced and in a later patch removed. This makes it unreasonably hard to review this and I'm not comfortable with pushing the bunch without a proper review (not necessarily by me!). Patches need to be squashed here.

2. I'd rather first push something related but simpler: A buffer that displays one message only.
and a search buffer that displays messages. If we had those, then most of the use cases for this split-window buffer here would be covered IMO.

But anyway: feel free to use or polish this patchset.

andresmrm:
Sorry for not answering before...

The code *works*. I've been using it for +1 year now and, at least here, it's much faster with big threads. But the code itself is a mess.
The two points suggested by @pazz seems a good path to organize this feature. And I think it would be better to start from scratch, from a more recent **alot** version (this branch is based on an old one). I just don't know if I'll have time to do it...

andresmrm:
Hello again, @pazz!
I cloned a fresh master to try the directions you pointed.
Duplicated the SearchBuffer and created a SearchMessagesBuffer, that lists messages instead of threads. Now searching for a thread id we have a simple list of the thread messages. Great!

The other part following your directions would be the MessageBuffer. But if the search and the message are in different buffers, how will I display them at the same time? And how to change the MessageBuffer message when the focused message change in the SearchBuffer?

andresmrm:
Followed your suggestion and created #896.
But instead of a MessageBuffer I made a MessageViewer widget that is used by the SearchMessagesBuffer, this way we don't have to open another buffer to display a message. The SearchMessagesBuffer is a frame that displays a list of messages as the header and a MessageViewer as the body.
The code from that new PR should be much better than the one in this retro thread PR.

pazz:
thanks, I'll close this one then. (sorry for the late reply)

PR #752, opened 2015-02-11
Allow specifing the mailcap field key for `pipeto`

jkoelker:
When decoding the message, use the mailcap field specified for command
lookup.

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -556,6 +556,8 @@ class ChangeDisplaymodeCommand(Command):
                    'help': 'let the shell interpret the command'}),
     (['--notify_stdout'], {'action': 'store_true',
                            'help': 'display cmd\'s stdout as notification'}),
+    (['--field_key'], {'help': 'mailcap field key for decoding',
+                       'default': 'copiousoutput'}),
 ],
 )
 class PipeCommand(Command):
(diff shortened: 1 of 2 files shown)


pazz:
hi Jason, thanks for the patch.
Can you outline a possible use case for this?
@all: Does anyone else consider this useful enough for inclusion?

laarmen:
This might be useful if the mail isn't properly formatted and doesn't include the right mailcap, I guess.

jkoelker:
Sure! Unfortunately I have to deal with Exchange at work. It does some interesting things with email. ;)

One of those is the way calendar invites are sent out. There are two base64 encoded mime parts. One text/html and the other text/calendar. This patch allows for setting the mailcap as such:

```
text/html;  w3m -dump -o document_charset=%{charset} '%s'; nametemplate=%s.html; copiousoutput
text/html;  w3m -dump -o document_charset=%{charset} '%s'; nametemplate=%s.html; x-gmail                                    
text/calendar; gcalcli import -d -v; copiousoutput                   
text/calendar; cat -; x-gmail
```

The `copiousoutput` ones are for display in the thread, it will show a summary of the inline `.ics` that is in the text/calendar part, while allowing the base64 decoded version to be sent to the command with `pipeto`. I use the `x-gmail` key just grab the decoded `ics` and then shuttle that into `gcalcli` to import it into my calendar via a key binding.

pazz:
Ok, i see what you're doing. I'm not convinced that this is the way i'd solve the use case but hey: one selling point of alot is that it's supposed to be hackable. 
I don't see much harm in the patch code, particularly because `extract_body` is messy already anyway :)

Just for the record: I think I'd not set copiousoutput i nthe mailcap and instead handle  `text/calendar`
using a script that first displays the files content and then asks me if i want to import it via gcalcli.
Another solution would be to only use `copiousoutput` and display the file inline as you do,
and then write a short hook that extracts the attachment and pipes it to gcalcli. Then you can bind the hook to a key (instead of `pipeto`)..

PR #753, opened 2015-02-20
Port to Py3k

chuwy:
Changes to enable support of Py3k:

- Change all `map`, `filter`, `.keys` etc calls which return values considered
  to be subscriptable to list comprehensions or wrap them with `list`. Left as
  is same functions and methcds which values considered to be only iterated.
- Switch to explicit relative imports. Also minor rearrange imports to be more
  PEP8fy.
- Add reserve imports for new Py3k stdlib module structure: urllib, StringIO,
  etc
- Switch to setuptools
- Consider interpreter version on string/unicode/bytes conversions
- Minor PEP8 changes
- Switch to new exceptions syntax

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ docs/build
 docs/source/configuration/*table.rst
 /alot/VERSION
 tags
+.idea
+alot.egg-info
--- a/alot/account.py
+++ b/alot/account.py
@@ -1,10 +1,11 @@
 # Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
 # This file is released under the GNU GPL, version 3 or a later revision.
 # For further details see the COPYING file
-import mailbox
-import logging
-import os
+
 import glob
+import logging
+import mailbox
+import os
 
 from alot.helper import call_cmd_async
 from alot.helper import split_commandstring
--- a/alot/addressbooks.py
+++ b/alot/addressbooks.py
@@ -1,12 +1,12 @@
 # Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
 # This file is released under the GNU GPL, version 3 or a later revision.
 # For further details see the COPYING file
-import re
-import os
 
+import os
+import re
+
+from alot.helper import call_cmd, split_commandstring
 from alot.settings.utils import read_config
-from helper import call_cmd
-from alot.helper import split_commandstring
 
 
 class AddressbookError(Exception):
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -1,24 +1,23 @@
 # Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
 # This file is released under the GNU GPL, version 3 or a later revision.
 # For further details see the COPYING file
-import urwid
-import os
-from notmuch import NotmuchError
+
 import logging
+import os
 
-from settings import settings
-import commands
-from walker import PipeWalker
-from helper import shorten_author_string
-from db.errors import NonexistantObjectError
+from notmuch import NotmuchError
+import urwid
 
-from alot.widgets.globals import TagWidget
-from alot.widgets.globals import HeadersList
-from alot.widgets.globals import AttachmentWidget
+from alot import commands
+from alot.foreign.urwidtrees import ArrowTree, TreeBox, NestedTree
+from alot.db.errors import NonexistantObjectError
+from alot.helper import shorten_author_string
+from alot.settings import settings
+from alot.walker import PipeWalker
+from alot.widgets.globals import (AttachmentWidget, HeadersList, TagWidget)
 from alot.widgets.bufferlist import BufferlineWidget
 from alot.widgets.search import ThreadlineWidget
 from alot.widgets.thread import ThreadTree
-from alot.foreign.urwidtrees import ArrowTree, TreeBox, NestedTree
 
 
 class Buffer(object):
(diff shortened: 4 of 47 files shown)


pazz:
thanks for your efforts. i don't actually know all that much about python 2 vs 3,
so I hope some other people step in to review this and give you some feedback.
Generally, It'd help if you kept the patches small instead of one big commit.
Question: does your version run properly on both v2 and 3?
I'm asking because at least two of the dependencies (twisted and urwid) are not officially supporting py3k yet.

pazz:
about setuptools: there is another issue: https://github.com/pazz/alot/pull/709
care to comment about how this differs from your patch.
Also, I now had a glance at your patch and it's *really* difficult to review due to its size.
Would you mind breaking it down into a list of smaller, more specific patches? thanks!

chuwy:
Ok. I'll push non-squashed PR tomorrow. Biggest part of this port is switch to explicit relative imports, and rearrange of imports (I think I haven't to do that?). And yes, it work on both versions (as for me, I didn't tested everything yet). I use Twisted==15.0.0 and urwid==1.3.0 on both Python 3.4 and 2.7 and uwrid actually compatible with Py3k (https://pypi.python.org/pypi/urwid/).

On setuptools vs distutils. AFAIK distutils are broken and not maintained, setuptools is current "default option" (afew uses setuptools for e.g). #709 is more accomplished, I've just changed one line in setup.py for `setup.py develop` (which is very handy) and leaved it accidentally.

pazz:
ok, so you're supporting #709? I don't recall why exactly i closed that issue and i'm considering merging this instead. thoughts?
Just send another pull-req about the py3k stuff. I'm happy to consider this when its ready :)

chuwy:
I actually don't see anything wrong with #709. I can merge it and test with my py3k port. Besides now we have wrong dependencies: magic and gpgme which now are python-magic and pygpgme respectively, and #709 fix it.

Ok, I'll send it when will sure everything works properly. But I think we need also to increase requirements: Twisted 15 seems to work, but I not sure which lowest version is appropriate. Same for pygpgme: I didn't tested anything GPG-related and pygpgme seems pretty ancient.

pazz:
i merged #709 into testing now.
regardless, please rebase all branches you want merged to `master`. I don't mind small pull requests
and I'd prefer if they came in individual branches for each change: you can group small fixes like your last commit into one pull request, as long as they are separate commits. But these should be in separate PR's (and thus separate branches in your repo) from the py3k stuff.
thanks!

chuwy:
All changes in this branch are related to python3 port or they fix bugs introduced by me due porting (like the last one), so they are just reminders for me to not introduce them again when I'll split the first commit into parts.

I planning to split the first big commit today. Should I do `reset --hard` and `push --force` to this branch or just make another Py3k PR?

chuwy:
All work is here now: https://github.com/chuwy/alot/tree/python3

pazz:
before i forget: you need not touch anything withing `alot.foreign`. It will ge away soon (already is on the testing branch) because i detached the urwidtrees into a separate project. If you have changes for this, please send a PR here: https://github.com/pazz/urwidtrees

chuwy:
Ok. It seems the only py3k-incompatible thing inside `alot.foreign` is relative imports, and in urwidtrees project they're fixed.

pazz:
I will close this PR now, and you can send a new one for your new py3k cleanup branch.
If that's ready, then I will first merge it into testing for a week or so, then master.
This means I will have to rebase it to testing first (please keep your branch rebased to master)
and for this it really helps if commits are small in case there are conflicts.
Sorry to be so pedantic about small commits again :)

chuwy:
It's understandable :) Well, actually we couldn't avoid at least one big commit with imports. May be I souldn't rearrange the imports, because it's doesn't affect anything beside someone's (mine) PEP8-obsession. But anyway I had to change almost all files in order to other parts of imports-related work.

All the following commits will be much shorter.

lfos:
Any progress here? I started working on a Python 3 support, too, before I found this ticket.

@chuwy Do you plan on finishing this off? If you don't have time right now, maybe we can at least polish the existing patches on your py3k/python3 branches and merge them...?

chuwy:
Hi @lfos 
Well, actually all the work already done, at least for version for end of Feb. I started to split commits to make history more clean and ran out of time. But it worked! You may checkout another [branch](https://github.com/chuwy/alot/tree/python3) in my fork where I split commits. Now I have really no time for this, but you need very little efforts to merge it.

lfos:
@chuwy Yeah, I had a quick look at it looks mostly good. I wonder whether we should use six.moves instead of the try-except approach, though? Do you remember anything that didn't work or was still missing? I might be able to review it in detail and test it later.

@pazz Any comments on that series?

chuwy:
On py3k branch everything worked! On python3 branch sadly I don't remember what I haven't done (may be you can find some way to compare them and you'll find out).
As about six, I decided it doesn't worth it to include another dependency, since for me everything more clear with try-except approach.

pazz:
FTR: I am also opposed to another dependency that is not absolutely necessary
here. Can one of you comment on issue #779 ?
In particular: wasn't here some argument discussed here for or against relative
imports in python3?
Thanks!

lfos:
That additional dependency consists of a single file with <1000 SLOC. I am fine with the exception handlers as well, though.

I don't think relative imports were discussed here. They are the most straightforward way of porting the existing code, though.

pazz:
FTR: this PR is abandones in favour of #800

PR #755, opened 2015-02-21
Fix call without hooks

### LINKED ISSUE #754: `call` didn't work without hooks
It seems we must to have the hooks file to run **any** python code inside alot.

chuwy:
Fix #754


pazz:
good catch! thx

PR #756, opened 2015-03-07
make running alot from git easier (for development)

TomasTomecek:
So we can run alot like:

```shell
python alot/init.py
```

Or is there an easier way?

### DIFF
--- a/alot/init.py
+++ b/alot/init.py
@@ -184,3 +184,6 @@ def main():
 
     # set up and start interface
     UI(dbman, cmdstring)
+
+if __name__ == '__main__':
+    main()


pazz:
i usually just use `./setup.py install --user && alot ...` during development. I'm not sure about this 'develop mode' of setuptools though? I think this patch should be harmless though

TomasTomecek:
> I'm not sure about this 'develop mode'

Me neither. Never used it. I prefer changing my `PYTHONPATH`.

To be honest I do this in every python project with CLI I work on. It makes it so much easier to develop and test.

The only thing this can break is if someone imports `init.py` directly in interactive shell. But, is there anyone who does that actually?

pazz:
ok, pushed..

Issue #331, opened 2012-02-16
Saved searches/folder list with count


quite:
Saved searches, like a folder list that Emacs-notmuch has.
Has it been discussed? How could it work, as a buffer?

I would like the ability to have such a list, and one which
also displays the number of messages/threads for each
entry. For emulating folders with unread indicators...

TomasTomecek:
@pazz can I file a PR? Is it possible that you'll merge it at some point?

pazz:
sure thing! send a pull request. Before you do its hard for me to comment on the details of your implementation anyway. It is of course likely that your first PR does not get pushed directly,
but this shouldn't discourage you from submitting it!
Just remember to send one PR per feature (you implement both a saved-search list and a folders-tree as new buffers?) and split your contribution into manageable chunks.

ibotty:
Just FYI: Notmuch 0.23 added "Support for named queries". So "saved searches" are in notmuch proper.

pazz:
Is that documented somewhere?
thx,
p

Quoting Tobias Florek (2016-11-18 20:10:00)

ibotty:
Yes. It's documented in notmuch-search-terms(7) and notmuch-config(1).

pazz:
great! it seems to work for me:

1. check that `notmuch config get built_with.field_processor` returns `true`
2. `notmuch config set query.test "from:santa"
3. `notmuch search query:test`

search queries also work within alot as it's done in the lib \o/
Now it is a matter of adjusting alot's query string completer and perhaps build some small buffer to list saved searches.

TomasTomecek:
I just tested what @pazz wrote and it works just fine for me (Fedora rawhide). I also agree that having a dedicated buffer with list of queries would be awesome! Unfortunately I won't have time to implement anything like that in near future.

lucc:
On what system are you or how did you build xapian/notmuch with the field processor enabled? I think it is a compile time option of xapian but have trouble finding the right switch for it.

TomasTomecek:
As I said, Fedora rawhide. This is [the build](https://koji.fedoraproject.org/koji/buildinfo?buildID=828767) and here is the [build log](https://kojipkgs.fedoraproject.org//packages/xapian-core/1.4.1/1.fc26/data/logs/x86_64/build.log). I haven't built it myself.

lucc:
(I can confirm that the query:foo stuff now also works on Arch Linux (they are now shipping notmuch with the field processor enabled)).

[...]

PR #757, opened 2015-03-07
RFE: implement folder tree buffer

TomasTomecek:
Related to #331 
## Contains
- discover maildir structure
- query notmuch for # of unread msgs in folders
- display as tree
- theming:
  - empty folders
  - not empty folders
  - focused folders
- configuration
  - path to maildir
  - blacklisted folders
- enable (un)folding
- implement refresh
## Bugs

(these should be addressed **before** merging)
- [x] ~~collapse/expand all doesn't work~~ (IMO doesn't make much sense)
- [x] function next/previous unread folder behaves oddly on subfolders
- [x] get rid of TODOs and FIXMEs from the patch
- [x] docs
  - [x] written docs
  - [x] docstrings
## RFEs

(these may be addressed in subsequent PRs)
- completion for folders when searching
- new command: pre-complete searching for selected folder
- figure out something better for status line of folders buffer (e.g. # of new messages in favorite folders)

### DIFF
--- a/alot/__init__.py
+++ b/alot/__init__.py
@@ -1,5 +1,5 @@
 __productname__ = 'alot'
-__version__ = '0.3.7'
+__version__ = '0.3.8.dev'
 __copyright__ = "Copyright (C) 2012-15 Patrick Totzke"
 __author__ = "Patrick Totzke"
 __author_email__ = "patricktotzke@gmail.com"
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -5,6 +5,7 @@ import urwid
 import os
 from notmuch import NotmuchError
 import logging
+from alot.widgets.account import AccountTree
 
 from settings import settings
 import commands
@@ -18,7 +19,7 @@ from alot.widgets.globals import AttachmentWidget
 from alot.widgets.bufferlist import BufferlineWidget
 from alot.widgets.search import ThreadlineWidget
 from alot.widgets.thread import ThreadTree
-from urwidtrees import ArrowTree, TreeBox, NestedTree
+from urwidtrees import ArrowTree, TreeBox, NestedTree, CollapsibleArrowTree
 
 
 class Buffer(object):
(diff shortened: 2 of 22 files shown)


TomasTomecek:
@pazz could I get some feedback here?

Would be awesome to have this merged so I don't have to run alot from my custom branch.

pazz:
the biggest problem i have with this is that its difficult to review because its one large commit.
Also, I wonder if we cannot get rid of this extra os.walk replacement and rather use the standard lib somehow. it will be difficult to support this when the stdlib changes.
The way the tree looks should be configurable (thx for amending the theme files btw)

TomasTomecek:
> the biggest problem i have with this is that its difficult to review because its one large commit.

I could probably break it a bit. The thing is that most basic implementation will have several hundreds lines of code anyway.

> Also, I wonder if we cannot get rid of this extra os.walk replacement and rather use the standard lib somehow. it will be difficult to support this when the stdlib changes.

The custom walk replacement solves this issue:

https://www.python.org/dev/peps/pep-0471/

pazz:
about the custom walks and path explorations:
I think there is a much nicer solution which is to let the notmuch database provide the visible folders.
You can define for instance a new `Tree` subclass (that you later display) that gets its information from
notmuch:

```python
db=notmuch.Database()
notmuch_path  = db.get_path()
root = db.get_directory(notmuch_path)  # this is a notmuch.Directory
children = root.get_child_directories
```

You can for instance preproces this, instanciating a `SimpleTree`... no os.path voodoo needed.

TomasTomecek:
Oh my... When I wrote this back then, that api of notmuch wasn't available yet. It will make the code _way_ more simple. Will rebase and rewrite (it will take a couple of weeks since I'm pretty busy lately).

pazz:
Quoting Tomas Tomecek (2015-11-04 11:59:53)
>    Oh my... When I wrote this back then, that api of notmuch wasn't available
>    yet. It will make the code way more simple. Will rebase and rewrite (it
>    will take a couple of weeks since I'm pretty busy lately).

no problem, take your time :)

TomasTomecek:
Rewritten (using notmuch folder api) and rebased against master. Removed also the blacklist folder functionality.

pazz:
looks good at first glance. please split this into separate patches, even if the functionality only makes sense as a whole.

TomasTomecek:
hope I'll manage to do this around Christmas

TomasTomecek:
rebased against testing (can't change base branch of this PR, that's why your commit is in there), updated docstrings and some code a tiny bit, split into multiple commits (not sure about the granularity)

let me know if this is fine

pazz:
I don't understand why you have a class "Account": notmuch has no concept of accounts and
as far as i understand, this new buffer type allows to list those folders seen and indexed by notmuch.
So you should not need this. Also, please check but i think libnotmuch already has a class "Folder". why do you need another wrapper?

TomasTomecek:
> I don't understand why you have a class "Account": notmuch has no concept of accounts and
> as far as i understand, this new buffer type allows to list those folders seen and indexed by notmuch.
> So you should not need this.

We can rename it. Would `Folders` be fine?

> Also, please check but i think libnotmuch already has a class "Folder". why do you need another wrapper?

There's just `Filename` and `Directory` and those contain filesystem related methods, not stuff for doing queries.

https://notmuch.readthedocs.org/en/0.15/filesystem.html

The wrapper of mine is responsible for doing `unread` queries, e.g.:

```
$ notmuch count folder:"Inbox" tag:unread
```

TomasTomecek:
Updated (please, don't get scared by the number of commits, will squash in the end -- I think that squashing it now would make the review process more complicated).

ibotty:
Any news?

TomasTomecek:
I am using this ever since I wrote it and it works just fine.

mschaaf:
@TomasTomecek are still working on your pull request? I will like to use it but with the current master branch.

TomasTomecek:
@mschaaf I'm not working on it any more, hence closing

PR #758, opened 2015-03-07
settings: don't fail ugly when section for binding misses

TomasTomecek:
Took me some time to figure out when I started my work on the folder buffer — would like to make it easier for others in future.

### DIFF
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -313,6 +313,11 @@ class SettingsManager(object):
         # get all bindings in this mode
         globalmaps, modemaps = self.get_keybindings(mode)
         candidates = globalmaps.keys() + modemaps.keys()
+        if mode != 'global':
+            try:
+                candidates = candidates + self._bindings[mode].scalars
+            except KeyError:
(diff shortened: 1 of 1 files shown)


pazz:
i dont understand why you dont just ship an empty section in the default condig as part of the patchset that adds a new buffer?

TomasTomecek:
Well, first you need to know you have to create that section in the default config. I didn't (and was really puzzled by that weirdly looking `KeyError`). I also prefer seeing
```
RuntimeError: "Mode 'my-great-mode' is not in default bindings."
```
rather than
```
KeyError: 'my-great-mode'
```

lucc:
@TomasTomecek sorry to bother you with this old stuff but do you mind rebasing this? And if you remember how the actual problem manifested do you think we can add a test to exercise it? 

As there was no list access with the mode name as a key here previously where did the original KeyError come from?

TomasTomecek:
I'm sorry but I have now idea about this PR any more, hence closing it. Feel free to carry it if you wish.

PR #766, opened 2015-04-25
Encrypt on reply to encrypted email (3)

meskio:
For issue #661.

A third try, following up the comments in #669. Sorry, it took me more than a year to get back on this. I hope I will be faster in next iterations.

I'm not really happy with this solution, as the code in `alot.commans.thread.ReplyCommand._set_encrypt` it's a copy if the code in `alot.commands.envelop.EncryptCommand.apply`. I'm sending this pull-req more to open a discussion than to get this solution merged.

This code could be moved to `alot.db.envelop.Envelop`, but that will mean to pass `ui` to the method there to be able to do `ui.choice`, it might not be a clean solution. Another choice could be to convert it into a utility function and place it somewhere, but I don't know alot's code in depth enough to come with a proposal on where to place it. @pazz what is your opinion on that?

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -13,6 +13,8 @@ from email.message import Message
 import mailcap
 from cStringIO import StringIO
 
+from alot.errors import GPGProblem, GPGCode
+from alot import crypto
 from alot.commands import Command, registerCommand
 from alot.commands.globals import ExternalCommand
 from alot.commands.globals import FlushCommand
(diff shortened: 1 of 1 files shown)


pazz:
Hi Ruben, thanks for getting back to this.
I see at least two sensible options here:
1st being, as you suggested, a utility function with the key-lookup stuff (from
line 275 onwards) that is then used both by ReplyCommand and EncryptCommand.
The second approach would be to not look up key hashes but to just store hints
(email addresses) in the envelope and let envelope.SendCommand deal with all this.
I think I'd prefer the first option.

meskio:
I think I prefer the first option too.

The second option will be similar to how mutt does it, that it don't ask you what key you want until you hit the send button. But one nice thing of alot is that you can do `:encrypt 0xC732B1D1C28F4E2F`, if we only store email addresses we'll lose this feature. We'll have to store addresses and hashes to be able to maintain this feature. I'm in favor to keep it simple, and 1st looks simpler to me.

Do you have any suggestion on where can this utility function fit? Or should I try to come up with something myself?

pazz:
you could put it into `alot/commands/utils.py` or something like that. It will be used only by commands for now right?

Issue #771, opened 2015-06-01
Prefer matching "To:", "CC:", "From:" over "Delivered-To:" when selecting sender address.


MacGyverNL:
I have one mailbox where a number of different addresses are delivered. Some of these addresses I use to send mail with, and they are listed as accounts in the config. However, in the end, all mail carries a Delivered-To header for the "main" account, which is listed at the top of the config file. Because it seems to pick the top account, many replies I start formatting start out with the "wrong" sender address. Changing the order of the accounts in the config file, moving this account entirely to the bottom, seems to fix the problem. But I consider some headers more valuable than others in determining where a reply should come from, so it might be worth considering changing this behaviour.

pazz:
Hi!
I'm not sure what exactly you're proposing here, but go ahead and send a
pull request (doesn't have to be perfect the first time) to get the discussion
going..

josch:
I'm in the exact same situation as @MacGyverNL and was about to file a bug for this but was extremely pleased to not only already find a bug about this issue but also a patch which is even merged already! Thanks alot (pun intended) @pazz and @MacGyverNL !!

josch:
Note for my future self: applying the commit 34537d8323767281c0e9a5537de9faa12fc4be0e from @MacGyverNL's fork requires application of commit b70d1e1790a163aa92f6259a586ff61baf879458 from the same fork first.

MacGyverNL:
Note that I recently rebased that on top of testing so I could start running on the testing-branch. I spoke with @pazz on IRC and he prefers pull requests based on master, so in the near future I'll rebase it on master again and provide a merge-commit with testing instead, so that you shouldn't have to apply b70d1e1 if you're running master.

lucc:
I assume this is done, please reopen if not.

PR #772, opened 2015-06-01
Introduce header-based preference in sender select

MacGyverNL:
Introduce a header-based preference in sender selection, as proposed
in #771.

In the examples below, assume the accounts are ordered X-Y-Z in the
config file.

Suppose I'm in a conversation, sending mails from account Y. However,
I add my account X to the "CC:" header. If I do not receive any replies
before sending another mail, I want to send it from account Y again,
9 out of 10 times. The old version would prefer account X, because it
lumps all headers in one list and gets a match on account X first.

Suppose I have aliasing in my mailserver, but send mails from several
of these aliases. They are not "aliases" as such, but accounts which
happen to share a mailbox. However, account X is the account where the
mail is ultimately sent to so will always be present in the
"Delivered-To:" header, even if the sender only knows my address for
account Y, which is in the "To:". Obviously, I want any replies to be
sent from account Y, not account X. This is fixed by prioritizing
matches on the "To:" over "Delivered-To:".

The only case which this patch doesn't fix is when somebody replies
in the first situation, adding *both* addresses to the "To:" header.
In this case, however, I would be fine manually editing the proposed
sender address.


MacGyverNL:
Not sure whether I did it entirely right, but that should make it configurable. Seems to work for me, anyway.

MacGyverNL:
Closing because I don't want to directly merge into master, I'll run this as rebase from testing for a while to ensure I don't introduce any problems, testing had diverged a bit.

PR #767, opened 2015-04-30
Encrypt on reply to encrypted email (4)

meskio:
For issue #661.

A fourth try, following up the comments in #766.

PR #774, opened 2015-07-12
Encrypt on reply to encrypted email and add 'encrypt_by_default' config

meskio:
For issue #661.

A fifth try, following up the comments in #767 .

I added an `encrypt_by_default` config option. Also I moved the reply key fetch to `alot.commands.global.ComposeCommand` as @pazz suggested.

Now it blocks the error message only when is a reply to an encrypted message to make sure the user realizes that the email will not be sent encrypted. It don't block in case of `encrypt_by_default` config option neither in case of `:encrypt` command.

How does it look now?

### DIFF
--- a/docs/source/usage/crypto.rst
+++ b/docs/source/usage/crypto.rst
@@ -39,3 +39,6 @@ and :ref:`toggleencrypt <cmd.envelope.toggleencrypt>` and
 in envelope mode to ask alot to encrypt the mail before sending.
 The :ref:`encrypt <cmd.envelope.encrypt>` command accepts an optional
 hint string as argument to determine the key of the recipient.
+
+You can set the default to-encrypt bit for each :ref:`account <config.accounts>`
+individually using the option :ref:`sign_by_default <sign-by-default>`.
--- a/alot/account.py
+++ b/alot/account.py
@@ -51,7 +51,8 @@ class Account(object):
                  gpg_key=None, signature=None, signature_filename=None,
                  signature_as_attachment=False, sent_box=None,
                  sent_tags=['sent'], draft_box=None, draft_tags=['draft'],
-                 abook=None, sign_by_default=False, **rest):
+                 abook=None, sign_by_default=False, encrypt_by_default=False,
+                 **rest):
         self.address = address
         self.aliases = aliases
         self.realname = realname
@@ -60,6 +61,7 @@ class Account(object):
         self.signature_filename = signature_filename
         self.signature_as_attachment = signature_as_attachment
         self.sign_by_default = sign_by_default
+        self.encrypt_by_default = encrypt_by_default
         self.sent_box = sent_box
         self.sent_tags = sent_tags
         self.draft_box = draft_box
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -12,12 +12,13 @@ from twisted.internet.defer import inlineCallbacks
 import datetime
 
 from alot.account import SendingMailFailed, StoreMailError
-from alot.errors import GPGProblem, GPGCode
+from alot.errors import GPGProblem
 from alot import buffers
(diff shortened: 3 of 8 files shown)


pazz:
sorry to keep you waiting. this looks great indeed. i merged it into testing for now

docs/source/usage/crypto.rst
```diff
@@ -39,3 +39,6 @@ and :ref:`toggleencrypt <cmd.envelope.toggleencrypt>` and
 in envelope mode to ask alot to encrypt the mail before sending.
 The :ref:`encrypt <cmd.envelope.encrypt>` command accepts an optional
 hint string as argument to determine the key of the recipient.
+
+You can set the default to-encrypt bit for each :ref:`account <config.accounts>`
+individually using the option :ref:`sign_by_default <sign-by-default>`.
```
lucc: Shouldn't that be `encrypt_by_default`? Otherwise I don't understand it.

lucc:
Does this option work for you? If I run the alot testing branch with these modifications to my config I do not get automatically encrypted email even if the key of the recipient is known to gpg.
```diff
diff --git a/config b/config
index 2d034ab..ea300aa 100644
--- a/config
+++ b/config
@@ -12,6 +12,7 @@ theme = luc
     # TODO: make it possible to move this to the toplevel
     # This will use the default key from gpg.conf
     sign_by_default = True
+    encrypt_by_default = True
     [[[abook]]]
       type = abook
       abook_contacts_file = ~/.config/abook/addressbook
diff --git a/hooks.py b/hooks.py
index 4d94d01..e7b8c21 100644
--- a/hooks.py
+++ b/hooks.py
@@ -145,22 +145,6 @@ def find_gpg_key(query):
                         return kid
 
 
-def post_buffer_open(ui, dbm, buf):
-    """Enable automatic gpg encryption of mail if possible."""
-    #logging.debug('Opening buffer {} of type {}.'.format(buf, type(buf)))
-    if isinstance(buf, alot.buffers.EnvelopeBuffer):
-        envelope = buf.envelope
-        logging.debug('To fields: {}'.format(envelope.get_all('To')))
-        logging.debug('Cc fields: {}'.format(envelope.get_all('Cc')))
-        logging.debug('Bcc fields: {}'.format(envelope.get_all('Bcc')))
-        addresses = [addr.lower() for name, addr in getaddresses(
-            envelope.get_all('To') + envelope.get_all('Cc') +
-            envelope.get_all('Bcc') + envelope.get_all('From'))]
-        keys = [find_gpg_key(addr) for addr in addresses]
-        if all(keys):
-            ui.apply_commandline('encrypt '+' '.join(keys))
-
-
 def reply_subject(subject):
     """Format the subject for a reply message.
 
```
If I use the alot and the config master branch the autoencryption seems to work.

docs/source/usage/crypto.rst
pazz: yes, that's surely a typo

pazz:
@lucc if i understand the patch correctly, you should add that option *per account* and not globally..

pazz:
(and yes, it does work as expected for me)

pazz:
merged to master

lucc:
I do set the option in the account section. Do you see the normal encryption pseudo header in the envelope buffer? I only get it *after* I execute `encrypt` (which finds the correct gpg keys as expected).

pazz:
yes, i see the pseudo header with the correct recipient key as expected, directly after the editor exits.
Any chance your buffer is not refreshed after the encrypt-action? (does the pseudo header appear after just a `refresh` ?)

meskio:
@pazz thanks for merging it, sorry for my delays that took it like a year to get it in.

pazz:
@meskio: could you check issue #819 ? It appears this bug is related.

PR #780, opened 2015-07-30
add account option alias_regexp

vrs:
This adds wildcard support for + extensions and just about any extension scheme you can imagine.

I used the name alias_regexp to stay consistent with the regexp option under abook.


pazz:
this looks nice and clean. I pushed it to the testing branch for now. Thanks.

PR #785, opened 2015-09-15
implement msg_focused hook

ibotty:
This hook can be used, e.g. to automatically use call the
mark-github-notifications-as-read hook whenever a message would be
marked as read by the auto_remove_unread setting.

### DIFF
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -5,6 +5,7 @@ import urwid
 import os
 from notmuch import NotmuchError
 import logging
+from functools import partial
 
 from settings import settings
 import commands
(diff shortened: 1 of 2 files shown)


ibotty:
I am not sure about the right place to call `run_hooks`. It might be after every `body.refresh`, but there does not seem to be a place for it.

I don't think there are other commands that change the thread's active message, so this should be the right spot for this hook, but it might not be general enough.

ibotty:
The diff still applies. Is there some comment regarding the patch?

pazz:
as this code is not very well documented (my fault, sorry), could you explain what exactly happens here? In particular, It seems that you somehow buffer hook functions in `hooks_to_run` before eventually firing them. why?

I would push this to the testing branch but the new hook name and its exact semantics should be documented in this file
https://raw.githubusercontent.com/ibotty/alot/master/docs/source/configuration/hooks.rst

Thanks,
p

pazz:
I'll close this for lack of activity. Feel free to re-open or send another PR if you still think a feature like this is necessary. Thanks

PR #793, opened 2015-11-03
Add Mailing lists reply

GuillaumeSeren:
Hi there,
as I wanted a way to reply quite simply to mailing list, see that 'old' PR:
https://github.com/pazz/alot/pull/479#issuecomment-152750321

Then I started to brew it for my use (I read many mailing list, thanks to notmuch / alot),
and I works well in my case, so here a quick sum up:
* It add --list to ':reply' to change the recipient header directly.
* It also add an option to activate, auto detect to reply to avoid adding --list all the time.
* The auto detect option is off by default so may not break other user's work-flow.

I try to keep it simple, and stick to the Mutt-like (L) Reply to the list,
so I do not added copy sender as cc because, it can be see as bad practice,
but maybe later.

Guillaume.

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -117,7 +117,7 @@ class ReplyCommand(Command):
         :type message: `alot.db.message.Message`
         :param all: group reply; copies recipients from Bcc/Cc/To to the reply
         :type all: bool
-        :param listreply: reply to list
+        :param listreply: reply to list; autodetect if unset and enabled in config
         :type listreply: bool
         :param spawn: force spawning of editor in a new terminal
         :type spawn: bool
@@ -163,6 +163,11 @@ class ReplyCommand(Command):
             if not subject.lower().startswith(('re:', rsp.lower())):
                 subject = rsp + subject
         envelope.add('Subject', subject)
+
+        # Auto-detect ML
+        auto_replyto_mailinglist = settings.get('auto_replyto_mailinglist')
+        if auto_replyto_mailinglist and mail['List-Id']:
+            self.listreply = True
 
         # set From-header and sending account
         try:
@@ -227,7 +232,8 @@ class ReplyCommand(Command):
                 to = mail['To']
             logging.debug('mail list reply to: %s' % to)
             # Cleaning the 'To' in this case
-            envelope.__delitem__('To')
+            if envelope.get('To') is not None:
+                envelope.__delitem__('To')
 
         # Finally setup the 'To' header
         envelope.add('To', decode_header(to))
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -103,7 +103,8 @@ def determine_sender(mail, action='reply'):
 
 @registerCommand(MODE, 'reply', arguments=[
     (['--all'], {'action': 'store_true', 'help': 'reply to all'}),
-    (['--list'], {'action': 'store_true', 'dest': 'listreply', 'help': 'reply to list'}),
+    (['--list'], {'action': BooleanAction, 'default': None,
+                  'dest': 'listreply', 'help': 'reply to list'}),
     (['--spawn'], {'action': BooleanAction, 'default': None,
                    'help': 'open editor in new window'})])
 class ReplyCommand(Command):
@@ -111,7 +112,7 @@ class ReplyCommand(Command):
     """reply to message"""
     repeatable = True
 
-    def __init__(self, message=None, all=False, listreply=False, spawn=None, **kwargs):
+    def __init__(self, message=None, all=False, listreply=None, spawn=None, **kwargs):
         """
         :param message: message to reply to (defaults to selected message)
         :type message: `alot.db.message.Message`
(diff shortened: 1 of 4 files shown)


pazz:
very nice! did you have a look at my other comments for the previous patchset?
I think we should somehow specify (in the docs inline at least, but maybe in the help-sring for this option) what exactly the semantic of `--list` is. If it is the same as for some mutt feature, which is documented on the web, even a link could be helpful.
Apart from this I'd be happy to push this as it seems to be a recurring feature request.

GuillaumeSeren:
@pazz 
Yes I read the old PR and integrate nearly all feedback, except on the cc and groupreply stuff,
which I feel we can live without (maybe I'm wrong).

So about the L feature of mutt, despite to be very simple, what we know:
* The doc say "L	message is sent to a subscribed mailing list" http://dev.mutt.org/doc/manual.html#muttlists
* It need a mail to be setup in the muttrc to each account and mailing list (not easy to maintain), see http://www.jefftk.com/p/list-reply-in-mutt
* The message need the list-id param in the header, same idea here.

What I changed, I needed a reliable way to provide the list mail, because in some case it did not provide a Reply-To field which would be the perfect standard case, but not all mailing list provide it (and maybe my mail standard doc is a bit old -> http://people.dsv.su.se/~jpalme/ietf/mail-headers/mail-headers.html).

In my test it append, that some different software that provide the list or configuration, give different headers:
* 'X-BeenThere' is used at sourceforge and winehq ML.
* 'X-Mailing-List' is used by git-send-mail, and the git ML.

For upgrading the doc on --list I try to stay simple, but I can complete if you tell me what you need.

GuillaumeSeren:
I upgraded the branch with the new name for the list param, which now is called toList,
I think I've seen that one maybe in older thread, but I forget it.

pazz:
thanks for the followup info (see what i did there? :P) I must admit that I am still a bit confused about what exactly changes when i use `reply --list` instead of `reply` to a list mail.
Maybe we want to refer to http://dev.mutt.org/doc/manual.html#using-lists ?

Anyway, does anyone else have an opinion about this?

GuillaumeSeren:
Ok I upgraded the branch, I hope we're progressing !

GuillaumeSeren:
Hey I have started to implement the boolean but I am not sure to fully understand the action related to the different state, please explain me that:

State | Action
-----------------
None | Default action when no boolean is given, let the 'standard' behaviour like auto detect if enabled.
True  | Same action as the default let the configured behaviour append.
False | Override config behaviour, and so you can reply to the sender and not to the list.

Th logic can be inverted, but I think this go against the auto-detect mode, which can be disabled in the config, so I prefer the version above.

GuillaumeSeren:
Hey,
I finally assume we agree on the logic, and I add 2 more commits:
1775a6f Add force True / false option to --list$
28d4ec2 Add a boolean parameter to --list 

So while adding the boolean, to deactivate the auto-detection in the 28d4ec2,
I felt it would be better to also have a 'True' option to be able to fully control the feature.

Finally here is the logic:
* None: (default) Let the process as defined in config.                               
* True: Force 'autodetect_lists=True' and ignore config setup, Need list-id.          
* False: Force 'autodetect_lists=False' and just reply to the sender.

pazz:
that looks great!
Could you squash the last two commits, and maybe remove the debug logging?
Ah: and the line
```
if self.listreply == None or self.listreply == True:
```
can just be `if self.listreply` again after the last commit (because it will not be `None`).
I can do it myself if you don't mind waiting..
How about renaming the config option to something that contains the word `reply` ?
Otherwise people might think it has something to do with displaying list mails or so..

GuillaumeSeren:
I upgraded the branch with the following modifications:
* Squash last 2 commit into a single one.
* Delete the logging, still keeping the 'reply', each reply has one.
* Upgrade the 'if self.listreply == None or self.listreply == True:' into 'if self.listreply'.
* I renamed 'autodetect_lists' to 'auto_replyto_mailinglist', I hope it is better.

Hope it will be good ! :)

GuillaumeSeren:
Hey I come to check if you some time to review the PR ?

To me I think I get every point you give, feel free to send me some feedback.

Guillaume.

pazz:
sorry for the delay. i pushed it to testing for now. thanks

GuillaumeSeren:
Great !

GuillaumeSeren:
Hey,
Since you put those commit into the testing I can hope that they will be in the next stable right ?

When do you plan to tag it ?

pazz:
Quoting Guillaume Seren (2015-12-16 18:43:10)
>    Hey,
>    Since you put those commit into the testing I can hope that they will be
>    in the next stable right ?

It is in master already as of yesterday
> 
>    When do you plan to tag it ?

soon, there are only a handfull of small things I want to include..

GuillaumeSeren:
Oh great I didn't see that !

There is some other things I want to work on I will start some new branches \o/

PR #806, opened 2015-12-10
Add commands to jump to next message matching query

arekfu:
Hi,

This pull requests adds a `jump` command to thread mode to move to the next/previous message matching a specific query. I have added the following default bindings in thread mode:

```
/ = prompt 'jump '
n = jump
N = jump --backwards
```

This is intentionally similar to `vim`'s behaviour, but I had to change the binding for `editnew` to `'ctrl n'`.

Perhaps it might be useful to add default bindings for `jump tag:unread` and `jump --backwards tag:unread`.

Eventually I would like to add a similar feature to search buffers, but that needs a bit more work because search buffers do not have the convenient `focus_next_matching` and `focus_prev_matching` methods.


arekfu:
Oops, probably the last commit wasn't such a great idea. `?` is reserved for the help screen.

pazz:
Hi!
I like the `move next matching ..` commands and would actually prefer only those and no hardcoded
aliases `jump`. Since most of us use keybindings anyway i think those are totally sufficient.
How about you send a new PR containing the move-patch, one that cleans up the
`help move` accordingly and maybe one that introduces some default bindings? thanks!

pazz:
I believe that this functionality is now provided by the `move next QUERY` thread command.

PR #809, opened 2015-12-10
include some files when building a dist

geier:


pazz:
why should this be part of the distributed module?

geier:
This makes the assumption, that you want everything a user might want to use or have a look at in the distribution. If you believe users should work directly from a git checkout, this can be closed.

And users probably want to know what's new, have the zsh completion file, etc.

BTW uploading a release of alot to PyPI really is as simple as:
 1. [registering yourself on PyPI](https://pypi.python.org/pypi?%3Aaction=register_form)
 2. asking me or PyPI user *julian* to make you the owner of alot on PyPI
 2. running `python setup.py sdist`
 3. optional:  running `python setup.py bdist_wheel --universal`
 4. run `twine upload dist/*`

For 3. you need to `pip install wheel` and for 4. `pip install twine`.
For every other packaging related questions I refer to the [Python Packaging User Guide](https://python-packaging-user-guide.readthedocs.org) and of course to bug reports if I did something stupid.

But no pressure ;-)

pazz:
can I have some input from packagers about this one?

regarding PYPI: thanks for the update @geier ! I would like to personally keep out of this because I'd rather spend my limited alot-maintenance time on other things, but feel free to go ahead and maintain alot un pypi!

I will see if i find the time to do a release soon (see issue #810)

zenhack:
I agree this patch is a good idea - it means that the source tarball on PyPI actually contains everything from the source tree that I might want to package. I maintain the AUR package, and without that in the tarball I have to fetch from Git regardless if I want to ship zsh completion or the .desktop file.

pazz:
but /extra also contains theme files and so on which i feel should not be part of a python module..

zenhack:
The fact that it's in `MANIFEST.in` doesn't mean it will get installed with the module, just included in the source tarball. The usual expectation is that `python setup.py sdist` builds a source tarball, so I feel it should be included for consistency with that expectation, if nothing else.

pazz:
ok, so from that perspective, it also makes sense to include the theme files?

zenhack:
Quoting Patrick Totzke (2015-12-17 18:44:49)
>    ok, so from that perspective, it also makes sense to include the theme
>    files?

That would be my take, yeah.

TomasTomecek:
@zenhack 
> I have to fetch from Git

Fedora maintainer here. Am doing the exactly same thing -- getting git archive and building & installing from it.

lucc:
Maybe we have to find out how to override some of these checks on special PRs if we want to.

pazz:
@lucc: as an "administrator" i can just merge anyway, even though the checks don't pass (i think this was a QC bug there). I believe you have "collaborator" status. Did it not allow a merge overide for you?

lucc:
I don't know if I have override powers (and I don't need them). I just saw that github was waiting for travis but travis did not start the job. Either because there is no .travis file here or because travis was activated after this PR was started or I don't know. But the pending travis status was what I was referring to.

PR #813, opened 2015-12-22
Unthreaded Search Mode Results

adqm:
I'm not sure if this is something that is wanted in the main project (I can just keep it on my machine if not), but I figured I'd put it up here in case it is useful to others.

Just trying out alot for the first time, and one of the things I missed from mutt was the ability to turn threading off, and just have the search mode view show individual messages.

This changeset implements a couple of different behaviors in this vein, controlled by a config variable `threaded`:

* if `threaded` is `'yes'` (the default), then alot runs as before.
* if `threaded` is `'no'`, then alot will show individual messages instead of threads in search mode, and will only display the single selected message when entering thread mode.
* if `threaded` is `'semi'`, then alot will show individual messages in search mode, and when entering thread mode, it will open the entire thread, but it will focus on the particular message that was selected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.py[co]
 *.log
 *.swp
+*~
 *.egg-info/
 /bin
 /build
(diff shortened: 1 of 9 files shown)


andresmrm:
Hi, @adqm! I think #896 is related with this issue. Sorry, only saw your PR after coding the other one.

PR #814, opened 2015-12-24
Custom Headers, Ordered Headers, ask_to option

adqm:
Again, just a few more changes that I was interested in for myself, sharing in case they are useful.  This changeset:

* adds an option `ask_to`, for whether to ask for the `To` header when composing mail
* always adds `To` and `Subject` to the `Envelope`, regardless of the values of `ask_to` and `ask_subject` (so that they show up when editing the file)
* modifies `Envelope` to store headers in the order they were added, and uses this to ensure some kind of sane ordering in the file when editing
* allows adding custom headers, via a `custom_headers` section in the config file

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -343,7 +343,9 @@ class EditCommand(Command):
 
         # decode header
         headertext = u''
-        for key in edit_headers:
+        for key in self.envelope.headers:
+            if key not in edit_headers:
+                continue
             vlist = self.envelope.get_all(key)
             if not vlist:
                 # ensure editable headers are present in template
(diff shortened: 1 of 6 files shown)


pazz:
@lucc do you have an opinion on this? It'd have to be rebased but I'm generally not opposed.

lucc:
I don't have any specific objections against this PR. Why do you ask? (The PR has been inactive for a long time, are you still interested @adqm ?)

In general I vote for sane defaults and simple unified editing (the headers in the editor) in contrast to a myriad of options for all different aspects of whatever (in this case headers). We already have the editing of headers and the `honor_follow_up_to` and `followup_to` (btw maybe we can link to an explanation for these headers from our config docs?). `custom_headers` sounds like it can be done in a hook instead.

pazz:
I ask because it has been inactive and I wanted to clean up the RP queue :)
I agree that this can be done in hooks. So let's close it here.

Quoting Lucas Hoffmann (2018-02-21 20:51:18)

PR #815, opened 2015-12-24
automatically start gpg-agent if needed

adqm:
This changeset will make alot start a gpg-agent daemon when necessary during sending mail, and clean it up after sending.  This might be a bit extreme, and it is almost certainly not the way the gpg-agent daemon is intended to be used, but it eliminated some issues for me.

### DIFF
--- a/alot/db/envelope.py
+++ b/alot/db/envelope.py
@@ -5,6 +5,8 @@ import os
 import email
 import re
 import glob
+import signal
+import subprocess
 from email.encoders import encode_7or8bit
 from email.mime.text import MIMEText
 from email.mime.multipart import MIMEMultipart
@@ -24,6 +26,7 @@ from .utils import encode_header
 import email.charset as charset
 charset.add_charset('utf-8', charset.QP, charset.QP, 'utf-8')
 
+GPG_AGENT_RE = re.compile(r"GPG_AGENT_INFO=(?P<whole>.*?:(?P<pid>\d*):\d*;)")
 
 class Envelope(object):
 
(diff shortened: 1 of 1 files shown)


teythoon:
NACK. gpg-agent is supposed to be started on demand in gpg-modern (2.0.0+) and is a long-running process.  Your patch will break my setup.  If you have trouble with gpg-agent not being started on demand, please file a bug on bugs.gnupg.org.

lucc:
I agree with @teythoon that alot is not the place to implement this. You seem to use gpg 1.x because you use `$GPG_AGENT_INFO` in the patch. Have you tried to update to gpg 2?

Else you might just write a small wrapper shell script to start gpg-agent before starting alot. Or you could start gpg-agent at login or during shell init and keep it running. (Is there a reason to kill it early?)

adqm:
> You seem to use gpg 1.x because you use $GPG_AGENT_INFO in the patch. Have you tried to update to gpg 2?

I had no idea.  Switching to gpg 2 fixed the issue for me, thanks!  Closing.

> Your patch will break my setup.

Out of curiosity, is that actually true, @teythoon?  I understand it starts some extra processes, but would it really break anything?  If so, how?

teythoon:
@adqm it is.  I forward my desktops gpg-agent using ssh to my laptop running alot, so alot spawning a local one would prevent me from using the authentication token that is connected to the desktop machine.

lamarpavel:
>I forward my desktops gpg-agent using ssh to my laptop running alot

That sounds very interesting. If it's not too much effort, could you please provide a reference or documentation on how to do this?

teythoon:
Sure.  Not sure if I'm doing that the most elegant way, but I read the man pages of gpg-agent and ssh, and came up with:

    % grep extra ~/.gnupg/gpg-agent.conf 
    extra-socket /home/teythoon/.gnupg/S.gpg-agent.extra
    % ssh thinkbox.local "rm .gnupg/S.gpg-agent" ; ssh thinkbox.local -o "StreamLocalBindUnlink=yes" -R $HOME/.gnupg/S.gpg-agent:$HOME/.gnupg/S.gpg-agent.extra  -o ForwardAgent=no -t "/usr/bin/env PYTHONPATH=$HOME/.local/stow/alot/lib/python2.7/site-packages alot"

In theory (at least how I understand it), StreamLocalBindUnlink=yes should unlink the target socket first, but it doesn't (for me), that's why I remove it manually in the first invocation of ssh.  Don't do the 'env PYTHONPATH...' bit if you don't need it, but you get the idea.

lamarpavel:
Yeah, thanks a lot, I will try that out as soon as I get back home.

teythoon:
(Maybe StreamLocalBindUnlink doesn't do what I want b/c it is *Local* not, *Remote*...)

pazz:
(closing this)

meskio:
If I understand correctly this pull-req is being rejected. I think @pazz reopened when he wanted to close it.

pazz:
yes, my mistake

PR #816, opened 2015-12-24
don't indent threads if an option is set

adqm:
This small change allows the option to prevent indenting threads, via a new config option `indent_threads`.  The default is the current behavior.


pazz:
i am actually using a very similar patch myself locally :)
The reason i did not push it (yet) was that the messages in a thread are still ordered according to the thread structure, i.e., not by date or such..

pazz:
done #1173 
The option is called `thread_indent_replies` there and takes integers, so your version corresponds to values 2 and 0.

Issue #802, opened 2015-12-05
`arrow_head` display


genkimarshall:
First, thank you so much for this project!

I'm currently getting 'tofu' characters for the arrow heads on threads (as in, it can't render them) -- the ones that should look like [this](https://raw.githubusercontent.com/pazz/alot/master/extra/themes/screenshots/tomorrow.thread.png).

Looking into it, it looks like it traces back to this `urwid` library?
Because it's used [here](https://github.com/pazz/alot/blob/3988f5f74b154d47755da206bb0659b957e495fa/alot/buffers.py#L366) which then leads me to [this example](https://github.com/pazz/urwidtrees/blob/master/docs/examples/example2.arrows.py) which seems to use [this character](http://www.fileformat.info/info/unicode/char/search.htm?q=u25B7&preview=entity): ▷. But it does look like the function allows you to pass in your preference of character.
Well, I'm not sure I understand the problem. That unicode character can render fine in my terminal when SSH'd into my server, but when I run `alot` on my server, it shows up there as the tofu [].

Thanks for your time in advance. It's not a functionality problem and it seems no other issues have mentioned this problem so feel free to dismiss this, but I thought I would ask in case there is some clear solution I am missing.

pazz:
Hi!
Sorry but I cannot think of any good reason for your encoding issue.
But i agree that the feature you suggest, to make arrow heads etc configurable, would be nice.
The most simple solution would probably be to add config options for this
in the theme files, under `[[thread]]`,  (see for instance the non-urwid-attribute option `search.threadline.parts` for guidance).

PR #817, opened 2015-12-27
theme: Add option for custom thread arrow char

genkimarshall:
This is for #802.

Sorry for the delay on this! This is a simple solution for now, but:
- This adds a whole new method (with a new wrapper around it, emulating the
  other theme/setting methods).. Not sure if that is what you had in mind.
- There are of course [other characters that could be specified](https://github.com/pazz/urwidtrees/blob/master/urwidtrees/decoration.py#L303) as
  you know, but I wasn't sure how far you had in mind. I think it would be
  straightforward to add other options for these, and keep them all under the
  same method, passing in the string of the parameter desired? All of them
  would return the character to be used.
- I noticed there was a way to specify defaults in the spec, but I kept it as
  None and hard-coded the normal character into the code, because I couldn't
  get it to recognize the character in the spec as a unicode character (versus
  a string of chracters).
- There is no user-facing documentation yet, but I can definitely add them
  based on how this goes.

Let me know, and I can keep working on this PR.

Cheers.

### DIFF
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -363,11 +363,13 @@ class ThreadBuffer(Buffer):
 
         bars_att = settings.get_theming_attribute('thread', 'arrow_bars')
         heads_att = settings.get_theming_attribute('thread', 'arrow_heads')
+        arrow_tip_char = settings.get_theme_arrow_tip_char()
         A = ArrowTree(self._tree,
                       indent=2,
                       childbar_offset=0,
                       arrow_tip_att=heads_att,
                       arrow_att=bars_att,
+                      arrow_tip_char=arrow_tip_char
                       )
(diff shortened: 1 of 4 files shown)


pazz:
looks ok to me. why don't you put the default char into the theme spec file as default? or did you run into encoding issues?

genkimarshall:
Yeah, I couldn't get it to recognize it as a unicode character code, rather than a string of the characters "\u25B7". Is there some way I'm missing?

genkimarshall:
Going through PR's I still have open, I'll close this one to go along with the
recent clean-up efforts. It doesn't scratch an itch I have any more, and it
doesn't seem like anyone else wants it anyway :)

PR #825, opened 2016-01-06
Workaround for case-insensitivity of domain names

adqm:
This is my workaround for making sure that the "reply" action chooses the account associated with `user@example.com` even if the e-mail was sent to `user@EXAMPLE.COM`.  Since domain names are case insensitive, these should always be the same account (as should any combination of capital and lower-case letters in the domain), and so this prevents the user from having to add aliases for all of the different possibilities.

### DIFF
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -164,7 +164,9 @@ class SettingsManager(object):
             accountmap[acc.address] = acc
             for alias in acc.aliases:
                 accountmap[alias] = acc
-        return accountmap
+        return dict((self._lower_address(addr), acc)
+                    for (addr, acc)
+                    in accountmap.iteritems())
 
     def get(self, key, fallback=None):
         """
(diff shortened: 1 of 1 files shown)


pazz:
isnt the username also case insensitive?
In this case, one can simply replace the `myad` in `settings.Manager.get_account_by_address`
by `myad.lower()`..

adqm:
Not necessarily, I think.  Technically, the handling of the username is totally up to the discretion of the host system.  Practically, it is probably the case that most host systems will treat the username as case-insensitive, but there could be some domains whose usernames are case-sensitive.

From a practical perspective, if someone is receiving e-mails addressed to USER@example.com and user@example.com, it is probably the case that they want to use the same account to respond to both of those.  So I am happy to make that change if you think that makes sense.  It certainly makes for a simpler patch.

lucc:
As far as I remember user names are case sensitive by definition but insensitive in practice ([see wiki](https://en.wikipedia.org/wiki/Email_address)). So I would not apply this path to the username as well.

pazz:
could you also drop some inline docs and squash everything (old and new) into one commit?
thanks!

dcbaker:
I'm going to close this as this hasn't been updated in years, and a more complete solution is being disccused in issue #1107

PR #833, opened 2016-01-26
refresh current buffer on SIGUSR1

mpnordland:
Small change to add a signal handler to UI that refreshes the current refreshes the current buffer when the program receives the SIGUSR1 signal. Tested on my system for about a month now and I have seen no problems. This is used to notify alot from my mail downloader to refresh the mail view.

### DIFF
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -3,6 +3,7 @@
 # For further details see the COPYING file
 import urwid
 import logging
+import signal
 from twisted.internet import reactor, defer
 
 from settings import settings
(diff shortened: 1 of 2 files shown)


pazz:
I think this is a great feature. All this PR lacks is some documentation for it :)

mpnordland:
Added documentation to docs/source/usage/synopsis.rst and a docstring to the function

lucc:
About the use case described in the docs: doesn't it make more sense to send the signal from a notmuch hook as you probably run notmuch from offlineimap anyways.

mpnordland:
I hadn't realized notmuch had hooks.

On January 26, 2016 1:24:27 PM EST, Lucas Hoffmann <notifications@github.com> wrote:
>About the use case described in the docs: doesn't it make more sense to
>send the signal from a notmuch hook as you probably run notmuch from
>offlineimap anyways.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/pazz/alot/pull/833#issuecomment-175159119

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #834, opened 2016-01-29
Fix editnew error where account might be None in ComposeCommand

### LINKED ISSUE #819: editnew bug
```
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 588, in _runCallbacks
  File "/home/pazz/.local/lib/python2.7/site-packages/alot-0.3.8.dev0-py2.7.egg/alot/ui.py", line 633, in call_apply
    return defer.maybeDeferred(cmd.apply, self)ts
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1274, in unwindGenerator
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1128, in _inlineCallbacks
    result = g.send(result) is already due since 23 Dec 2015.
  File "/home/pazz/.local/lib/python2.7/site-packages/alot-0.3.8.dev0-py2.7.egg/alot/commands/globals.py", line 846, in apply
    if self.encrypt or account.encrypt_by_default:
exceptions.AttributeError: 'NoneType' object has no attribute 'encrypt_by_default'

```

meskio:
Fixes #819.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -770,55 +770,54 @@ class ComposeCommand(Command):
 
                 self.envelope.add('From', fromaddress)
 
-        # add signature
-        if not self.omit_signature:
-            name, addr = email.Utils.parseaddr(self.envelope['From'])
-            account = settings.get_account_by_address(addr)
-            if account is not None:
-                if account.signature:
-                    logging.debug('has signature')
-                    sig = os.path.expanduser(account.signature)
-                    if os.path.isfile(sig):
-                        logging.debug('is file')
-                        if account.signature_as_attachment:
-                            name = account.signature_filename or None
-                            self.envelope.attach(sig, filename=name)
-                            logging.debug('attached')
-                        else:
-                            sigcontent = open(sig).read()
-                            enc = helper.guess_encoding(sigcontent)
-                            mimetype = helper.guess_mimetype(sigcontent)
-                            if mimetype.startswith('text'):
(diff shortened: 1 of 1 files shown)


pazz:
how could `account` possibly be None?

pazz:
aha! https://github.com/pazz/alot/blob/master/alot/commands/globals.py#L798

Since subsequently, many of these checks if `account` is not None occur, I supose it'd be best if
this line was replaced with something that ensures that this is the case, for instance by defaulting to
the first account or so. This would make perfect sense for this command i think.

meskio:
I agree, it's a good idea to default to an account. Let's try to fix it this way.

meskio:
I force-pushed the fix by setting the first configured account if no address matches.

pazz:
It seems that from line 749 onwards we already determine the account to be used.
I think just before this point, the account should be determined, by looking up the From-header if exists, and by prompting otherwise, and if no accounts are set up we can raise an exception.
From that point onwards we can then safely use the `account` variable.

meskio:
In line 749 it determine the From, I just moved the account set below that, you are right that it makes sense to do it there.

meskio:
bump as requested in #819

pazz:
pushed to testing. thanks and sorry for the delay

PR #843, opened 2016-02-23
Add vcard addressbook feature.

lucc:
What do you think about this?

Docs are still missing, and one bug has to be fixed (see 1a0fcae, does anybody know how to do this?).

### DIFF
--- a/alot/addressbook/vcard.py
+++ b/alot/addressbook/vcard.py
@@ -1,3 +1,7 @@
+# Copyright (C) 2017  Lucas Hoffmann <l-m-h@web.de>
+# This file is released under the GNU GPL, version 3 or a later revision.
+# For further details see the COPYING file
+
 try:
     import vobject
 except ImportError:
@@ -32,7 +36,7 @@ class VcardAddressbook(AddressBook):
             paths = [path]
         elif os.path.isdir(path):
             paths = []
-            for dirpath, dirnames, filenames in os.walk(path):
+            for dirpath, _, filenames in os.walk(path):
                 for filename in filenames:
                     paths.append(os.path.join(dirpath, filename))
         else:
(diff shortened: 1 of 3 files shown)


pazz:
the code looks ok, but i'm not overly exited about introducing another dependency for this..

t-8ch:
I have written a small c program (for speed, python parsing was much to slow) for the same thing:
https://github.com/t-8ch/snippets/blob/master/contactquery.c

lucc:
@t-8ch that looks interesting but how do I compile it? I had no luck with `cc contactquery.c` or `cc -I /usr/include/glib-2 -l glib-2 contactquery.c`. What libraries are needed?

t-8ch:
@lucc there are a README and a Makefile (`make contactquery`) in the repo. If you have problems compiling feel free to open a ticket there.

pazz:
what's the status of this? Do you want to add this to the master branch still? 
If so, you could add some caching method to avoid parsing those files on every call..

lucc:
I did not work on this any more as you said that you don't like the extra dependency. At one point I did try to implement a differnt approach to the addressbook: To have `HookAddressbook` class that would ask a specific user defined hook function for the addresses instead of a specific user defined command. But I never finished and proposed it. I think I decided that it is saver to keep all these things (one can have a similar idea for sending mail) in different processes so exceptions are not tearing down the whole of alot.

(Background: When I first propsed this I had trouble with my addressbook https://github.com/scheibler/khard, because it is pretty slow. So completion was also slow. My idea here was to have some faster vcard parsing code in alot and use khard only to manage the addresses. Now I have written a shellscript/makefile which chaches the addresses and is much faster than khard itself.)

pazz:
OK thanks for the clarification. I'll close this issue for now, we can always reopen it later.

Issue #851, opened 2016-03-08
Ask the user before quitting


GuillaumeSeren:
Hey,
I just notice that, if I start to compose add recipient, then senders, then I changed my mind so I C-c to close that call, alot just quit (witch is not what i wanted). 
Maybe C-c in alot is not a good practice, and I know that <escape> works better, 
but I feel it could be great to have a confirmation to quitting, what do you think ?

pazz:
There is of course the "bug_on_exit" config flag, but that will only
trigger if you use the normal "exit" command (or close the last buffer).
I thing recently, someone added signal handling to `alot.ui.UI`,
so that `SIGUSR1` is intercepted and triggers a refresh.
It should be easy to add some code that intercepts `SIGKILL` and triggers `exit`
(which would then, if bug_on_exit=true, behave as you want).
Go for it! :)

Quoting Guillaume Seren (2016-03-08 10:52:28)

GuillaumeSeren:
@pazz:
Yes I can see that, I will try some test but maybe we can add that as a global behaviour,
so if you set it in you config you can prevent to quit alot by mistake.

GuillaumeSeren:
Oh nevermind, I did not test bug_on_exit flag it works on *normal* exit,
in mutt C-c trigger exit, and ask for quitting, and 'q' quit directly so I think,
we could 'force' bug_on_exit even if it is not set for that case.
It was some time I didn't read the source of alot so I am searching a bit to find a place for my hook.

GuillaumeSeren:
Ok I played a bit more with the source and added my case in the handle_signal() function,
it catch the signal and I can ask the use to confirm same way that bug_on_exit do.

Do you think we need something more ?

pazz:
I'd rather this signal triggers an exit command, because that would make sure
that all queued up changes to the index have been flushed before alot shuts
down.


Quoting Guillaume Seren (2016-03-09 22:40:13)

GuillaumeSeren:
@pazz :
I agree this way you can setup the bug_on_exit flag and get the expected behaviour,
in the other side (without the flag) it will quit in a cleaner fashion.
I am cleaning my patch, then I'll send it to you.

GuillaumeSeren:
Ok I prepared a PR here #852

PR #852, opened 2016-03-10
Add signal for SIGINT callback ExitCommand

GuillaumeSeren:
Hello,
this patch bind the SIGINT (C-c) signal to an exit command, 
that way is you configure the 'bug_on_exit' flag you can confirm before quitting,
if you don't you still get a cleaner way for quitting due to the ExitCommand().

### DIFF
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -8,6 +8,7 @@ from twisted.internet import reactor, defer
 
 from settings import settings
 from buffers import BufferlistBuffer
+from commands import globals
 from commands import commandfactory
 from commands import CommandCanceled
 from alot.commands import CommandParseError
@@ -69,6 +70,7 @@ class UI(object):
         mainframe = urwid.Frame(urwid.SolidFill())
         self.root_widget = urwid.AttrMap(mainframe, global_att)
 
+        signal.signal(signal.SIGINT, self.handle_signal)
         signal.signal(signal.SIGUSR1, self.handle_signal)
 
         # set up main loop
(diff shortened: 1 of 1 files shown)


pazz:
This looks nice!
The logging should rather say something like "shut down cleanly", but apart from that I'm happy.
I just skimmed the `exit` code and I cannot see how the case is dealt with, when the `dbman.writequeue` is not empty (because some flush failed due to a locked notmuch index).
In that case, `exit` should warn/wait/retry `dbman.flush()`.

GuillaumeSeren:
@pazz 
Ok I have upgrade the log message.
I feel a bit the same about the exit code, I mean I am not sure it cover every case possible.
It check the ui.db_was_locked, so maybe we can call the FlushCommand instead of (hard) checking db_was_locked, that way if the flush is not complete we can stop the quit.

This have the advantage of beeing good for all exit case IMO.

pazz:
did you forget to push the amended patch?
sidenote: i think it'd be nicer to check the signal for equality with `signal.SIGINT`, not just 2..

GuillaumeSeren:
No I was trying to find something better for the 'dbman.writequeue' check.

GuillaumeSeren:
From my understanding, the patches do the job described in the issue #851 
I feel that the flag db_was_locked populated by the FlushCommand, can be good enough,
and if we need a better check, it may be simpler to upgrade FlushCommand directly.

Are you agree ?

pazz:
yes. i will push this patch as it is. thank you.
we should open a new issue for this flushing stuff

GuillaumeSeren:
Great ! :beers:

Issue #849, opened 2016-03-01
feature/discussion: How to handle unknown/unvalidated/untrusted gpg keys?


lucc:
## Proposal 1
Indicate the validity of gpg keys in the summary in thread buffers (especially untrusted or unknown keys).
## Proposal 2
Skip keys depending on their validity when handling the `encrypt_by_default` option during envelope setup.

## Current situation for 1
Currently the relevant summary line in thread buffers looks like this: `PGP-Signature Valid: Foo Bar <foo@bar.net>` regardless of the validity of the key (field 2 of `gpg -k --with-colons`).
## Current situation for 2
Assume you have set `encrypt_by_default`. Imagine you write a person you don't know. For some reason you have a key that matches the email of that person in your keyring (but as you don't know the person it is of unknown validity). Now alot tries to encrypt this mail with a key of which you don't know if it belongs to the person in question.

lamarpavel:
Definitely agree on 1. Trust-level should be reported.

Not so sure about proposal 2. Instead of skipping the option  `encrypt_by_default` could accept {never, always, valid_only}.

An additonal problem is with multiple recipients, some of which might match valid keys while others do not. I have thought about this before but have not come to a satisfying conclusion yet.

lucc:
The idea to turn `encrypt_by_default` into a three value choice is nice. If we have this I think your last point does not need to be solved in `encrypt_by_default` as we still have `:toggleencrypt` and `:encrypt` so the user can handle all edge cases manually.

lucc:
@pazz are you interested in the three value thing for `encrypt_by_default`? If yes I will start it.

pazz:
So encrypt_by_default=always means mails would be encrypted to expired keys
if they are in the keychain? does that even make sense?
But generally I'm happy to push anything that makes sense and is well written :)

lucc:
I will try to elaborate a bit on my plan:

I had a look at the code and as far as I understand it the current implementation iterates over the recipients and tries to find valid (not expired, usable for encryption, the function used for this is `alot.crypto.validate_key`) keys for them. It will encrypt to all the keys that match these criteria. It is important to note that the message will be send to all recipients regardless if it is encrypted to all of them. So some recipients might end up with mail that they can not read. (Even if that is not optimal -- it would be better if alot warns the user before sending, but that is a topic for a new issue -- it is better then sending the message unencrypted along the encrypted message because that would compromise the whole idea.)

[...]

PR #854, opened 2016-03-11
Turn `encrypt_by_default` into a three choice option.

lucc:
See #849 for the initial discussion.

Currently the biggest question seems to be which values we should use for the three choices.  I might add more points for discussion.

This is still WIP.

### DIFF
--- a/alot/account.py
+++ b/alot/account.py
@@ -65,16 +65,23 @@ class Account(object):
         self.signature_filename = signature_filename
         self.signature_as_attachment = signature_as_attachment
         self.sign_by_default = sign_by_default
-        self.encrypt_by_default = encrypt_by_default
         self.sent_box = sent_box
         self.sent_tags = sent_tags
         self.draft_box = draft_box
         self.draft_tags = draft_tags
         self.abook = abook
-        if self.encrypt_by_default == u"True" or \
-                self.encrypt_by_default == u"False":
-            logging.debug("Deprecation warning: The format for the "
-                          "encrypt_by_default option changed.")
+        # Handle encrypt_by_default in an backwards compatible way.  The
+        # logging info call can later be upgraded to warning or error.
+        encrypt_by_default = encrypt_by_default.lower()
+        msg = "Deprecation warning: The format for the encrypt_by_default " \
+              "option changed.  Please use 'none', 'all' or 'trusted'."
+        if encrypt_by_default in (u"true", u"yes", u"1"):
+            encrypt_by_default = u"all"
+            logging.info(msg)
+        elif encrypt_by_default in (u"false", u"no", u"0"):
+            encrypt_by_default = u"none"
+            logging.info(msg)
+        self.encrypt_by_default = encrypt_by_default
 
     def get_addresses(self):
         """return all email addresses connected to this account, in order of
--- a/alot/account.py
+++ b/alot/account.py
@@ -54,7 +54,7 @@ class Account(object):
                  signature_filename=None, signature_as_attachment=False,
                  sent_box=None, sent_tags=['sent'], draft_box=None,
                  draft_tags=['draft'], abook=None, sign_by_default=False,
-                 encrypt_by_default=False,
+                 encrypt_by_default=u"none",
                  **rest):
         self.address = address
         self.aliases = aliases
@@ -71,6 +71,10 @@ class Account(object):
         self.draft_box = draft_box
         self.draft_tags = draft_tags
         self.abook = abook
+        if self.encrypt_by_default == u"True" or \
+                self.encrypt_by_default == u"False":
+            logging.debug("Deprecation warning: The format for the "
+                          "encrypt_by_default option changed.")
 
     def get_addresses(self):
         """return all email addresses connected to this account, in order of
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -573,5 +573,9 @@ class EncryptCommand(Command):
                 envelope.encrypt_keys.update(keys)
             else:
                 envelope.encrypt = False
+        if not envelope.encrypt:
+            # This is an extra conditional as it can even happen if encrypt is
+            # True.
+            envelope.encrypt_keys = {}
         # reload buffer
         ui.current_buffer.rebuild()
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -498,6 +498,7 @@ class SignCommand(Command):
 
 
 @registerCommand(MODE, 'encrypt', forced={'action': 'encrypt'}, arguments=[
+    (['--trusted'], {'action': 'store_true', 'help': 'only add trusted keys'}),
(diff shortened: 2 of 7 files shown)


pazz:
i directly pushed the docstrings; the second patch is in `testing` and `0.3.8-settings-encrypt-by-default-854` for now. thanks!

lucc:
You should not have pushed that yet! The PR is unfinished, thats why I said it is "WIP = work in progress". I just put it here to be able to discuss it publicly.

Discussion: What do you think about the doc/Makefile commit? Why again are you keeping generated files in git?

lucc:
PS: Ok the docstrings can go to master directly that's no problem but the rest is not ready at all!

pazz:
oh sorry! i thought the 2nd patch actually contained more than the spec change. my bad.
but its only in testing and should not even break any existing configs.

I am not sure i usnerstand the dependency in Makefile patch..

lamarpavel:
Regarding the naming of the option, I think the most transparent would be `encrypt_by_default =` __never__, __always__ or __trusted__.

However, this would break compatibility with the current `True` or `False`. So maybe we can have a transition period where `always` is evaluated the same as `True` and `never` as `False`, so you could say True/False or always/never for the same effect. Then maybe give a warning message upon starting alot that True/False will become deprecated in the future.

lucc:
About the Makefile commit:

It is not necessary for this PR it is just something I noticed when changing the specs. When I updated the spec file and rebuild the html documentation the new text was not in the html file. So I thought that it would be nice to tell the makefile when to rebuild these files automatically. 

About the option change:

1. Are there more (and maybe authoritative from @pazz ?) opinions about the names for the three option values? What about **all**, **none**, **trusted** (or **signed** instead of trusted)?
2. Currently I can not test the code as it is here because of this error:
```
key "encrypt_by_default" in section "accounts, web" failed validation: the value "Valid" is of the wrong type.
```
Where, aside from the spec file, do I need to tell alot that the option can be "Valid"?

pazz:
I have no strong feelings about the naming scheme, but  all/none/trusted sounds about right.
I think it's a good idea to also allow True/False for backwards compatability and interpret them as all/none.

@lucc in the spec file, the line should read


```
 encrypt_by_default = option('all', 'none', 'trusted', 'True', 'False', default='none') 
```

Note the quoted strings for True/False! The values must all be of the same type IIRC.

pazz:
Ah: If you have more unrelated improvements, it'd be best if you put them into a different branch.
I'll cherry-pick the Makefile stuff for instance anyway. Don't worry about this patch, just saying in case you have more :)

lucc:
The problem why I could not test it was just a dangling pip installation of alot that did interfere. This can now be tested by others. 

The "clean up recpient" commit is interesting. I think the last author intended to take only the address (match group 1) because there are parentheses in the regex. It was not a problem with encrypt_by_default=True but with trusted we need it.

pazz:
yes, that looks about right. By the way: the alot-command `encrypt` should now also be able to set one of the three possible values..

lucc:
There is one quirk with the current implementation of `encrypt --trusted` and `toggleencrypt --trusted`. Keys which are already set in the envelope are not checked. So the sequence `encrypt; unencrypt; encrypt --trusted` will sign the message with the same keys as a simple `encrypt`. The reason is that `unencrypt` does not remove the keys from the envelope but simply unsets the encryption bit, or alternatively that `encrypt --trusted` only checks the keys it tries to add and not the ones that are already set in the envelope.
Does this need to be fixed?

Apart from that what do you think about the last commits?

pazz:
Quoting Lucas Hoffmann (2016-03-24 12:38:41)
>    There is one quirk with the current implementation of encrypt --trusted
>    and toggleencrypt --trusted. Keys which are already set in the envelope
>    are not checked. So the sequence encrypt; unencrypt; encrypt --trusted
>    will sign the message with the same keys as a simple encrypt. The reason
>    is that unencrypt does not remove the keys from the envelope but simply
>    unsets the encryption bit, or alternatively that encrypt --trusted only
>    checks the keys it tries to add and not the ones that are already set in
>    the envelope.
>    Does this need to be fixed?


I think you have a good case here.. un[encrypt|sign] should indeed remove the
keys from the envelop.

lucc:
Ok, I will add that soon.

lucc:
I will put the fix for `unsign` in a new PR as this here did not touch the sign topic at all till now.

pazz:
thanks. This patchset looks good to me. I pushed this to branch `0.3.8-settings-encrypt-by-default-854`
and again merged to testing for now.

lamarpavel:
What's the behaviour for recipients with mixed key status? Say there are 3 recipients: A, B and C where A has a trusted key, B has an untrusted key and C has no key.

With `encrypt_by_default = trusted`, this should not encrypt because neither B nor C match the criteria, same for `encrypt_by_default = True`. Currently (in the master branch), alot encrypts the mail with `encrypt_by_default = True` even if one of the recipients (in either `to:` or `cc:`) has no key at all. This results in some people not being able to read the mail at all and I have to manually toggle encryption off in the envelope mode.

I would think that this issue should be addressed by this PR as well, in case it isn't.

lucc:
@lamarpavel check the docs changes in this PR. Currently we behave similar to the old behavioue on master. You are also right about the implications you name (manual toggling etc).

Do you have a suggestion how to handle this scenario differently? Currently I opt for "put the user in control and don't add a gazilion new options to handle every corner case" (as noted elsewhere: https://github.com/pazz/alot/issues/849#issuecomment-191166913)

lamarpavel:
@lucc My proposal would be to introduce no new options. Instead, form a list of all recipients, find the "weakest" one and set it as the bar to be matched against your config.

- If the "weakest" recipient has no public key at all (or none that can be found), then no encryption is applied.
- If the "weakest" recipient has an untrusted key __and__ you set `encrypt_by_default = Trusted`, then no encryption is applied.

Simply define that if any of the recipients does _not_ meet your reguirements, treat all recipients as if they haven't.

Together with issue #849 this would mean that the user is notified, that the outgoing mails are _not_ encrypted (maybe phrase it as "applied encryption does not meet preferred standards") and the user can manually toggle it on.

lucc:
@lamarpavel I like the idea to treat the value of `encrypt_by_default` as a user preferred standard against which the available keys should be checked in a threshold, all-or-nothing like decision.

Not sure how a warning should be displayed. I would not like it to intrusive (like an error message for example) b/c it might often happen that one writes emails to a group of people of whom one knows that one does not have a key for everybody.

lucc:
What do you think @pazz ?

lamarpavel:
@lucc I imagined the warning like a highlight of the "Encrypt" field in envelope mode. Different font color and a text next to it (eg. a unicode symbol of a broken lock, or text such as "encryption preference not met").

lucc:
@lamarpavel sounds good.

@pazz Do you think we/I should add that? To this PR?

PR #866, opened 2016-05-01
Tagging hook

np:
Here is an example of a simple hook to keep a log of the tagging commands:

``` python
import json
def pre_tagging(command, query, tags):
    with open('tagging.log', 'a') as h:
        doc = {'command': command, 'query': query, 'tags': tags}
        h.write(json.dumps(doc) + '\n')
```

### DIFF
--- a/alot/db/manager.py
+++ b/alot/db/manager.py
@@ -264,6 +264,11 @@ class DBManager(object):
 
     def _append_queue(self, command, afterwards, querystring, tags):
         self.writequeue.append((command, afterwards, querystring, tags))
+        tagging_hook = settings.get_hook('pre_tagging')
+        if tagging_hook:
+            res = tagging_hook(command, querystring, tags)
+            if res:
+                tags = res
 
     def count_messages(self, querystring):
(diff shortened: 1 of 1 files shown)


pazz:
I'm not sure: first of all, any such hook needs documentation in `docs/source/configuration/hooks.rst` of course.

More importantly, the way you implemented this, the hook will get called as soon as you queue some
tagging operation but not necessarily directly before that op is carried out.
The problem can be that the notmuch index is locked, in which case a new attempt is scheduled after some timeout. But you coud exit alot by then, which messes with your logged operations.
Actually, the proposed behaviour is already possible using `pre_tag` / `pre_untag` etc command hooks. 

If you want a hook that gets only called before the index-writing happens,
one could add a "before" callable to writequeue entries and call them after the write-lock
has been secured, in `DBManager.flush` (https://github.com/pazz/alot/blob/master/alot/db/manager.py#L133)...

pazz:
closing this here. Feel free to reopen or send another PR if you want to put forward another variant..

PR #878, opened 2016-07-04
Encrypt for the CC and Bcc addresses too

meskio:
If you activate encrypt on the envelop mode or you reply to an encrypted email by default only addresses in the To header are used to encrypt to, CC and BCC are not taking into account.

I'm solving this issue and also removing duplication of code.

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -14,7 +14,7 @@ from twisted.internet.defer import inlineCallbacks
 
 from . import Command, registerCommand
 from . import globals
-from .utils import get_keys
+from .utils import set_encrypt
 from .. import buffers
 from .. import commands
 from .. import crypto
@@ -563,27 +563,9 @@ class EncryptCommand(Command):
             encrypt = False
         elif self.action == 'toggleencrypt':
             encrypt = not envelope.encrypt
-        envelope.encrypt = encrypt
         if encrypt:
-            if not self.encrypt_keys:
-                for recipient in envelope.headers['To'][0].split(','):
-                    if not recipient:
-                        continue
-                    match = re.search("<(.*@.*)>", recipient)
-                    if match:
-                        recipient = match.group(1)
-                    self.encrypt_keys.append(recipient)
-
-            logging.debug("encryption keys: %s", self.encrypt_keys)
-            keys = yield get_keys(ui, self.encrypt_keys,
-                                  signed_only=self.trusted)
-            if self.trusted:
-                logging.debug("filtered encrytion keys: " +
-                              " ".join(x.uids[0].uid for x in keys.itervalues()))
-            if keys:
-                envelope.encrypt_keys.update(keys)
-            else:
-                envelope.encrypt = False
+            yield set_encrypt(ui, envelope, signed_only=self.trusted)
+        envelope.encrypt = encrypt
         if not envelope.encrypt:
             # This is an extra conditional as it can even happen if encrypt is
             # True.
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -7,7 +7,6 @@ import email
 import glob
 import logging
 import os
-import re
 import subprocess
 from StringIO import StringIO
 
@@ -17,7 +16,7 @@ from twisted.internet import threads
 
 from . import Command, registerCommand
 from . import CommandCanceled
-from .utils import get_keys
+from .utils import set_encrypt
 from .. import commands
 
 from .. import buffers
@@ -864,12 +863,12 @@ class ComposeCommand(Command):
             logging.debug("Trying to encrypt message because encrypt=%s and "
                           "encrypt_by_default=%s", self.encrypt,
                           account.encrypt_by_default)
-            yield self._set_encrypt(ui, self.envelope)
+            yield set_encrypt(ui, self.envelope, block_error=self.encrypt)
         elif account.encrypt_by_default == u"trusted":
             logging.debug("Trying to encrypt message because "
                           "account.encrypt_by_default=%s",
                           account.encrypt_by_default)
-            yield self._set_encrypt(ui, self.envelope, trusted_only=True)
+            yield set_encrypt(ui, self.envelope, block_error=self.encrypt, signed_only=True)
         else:
             logging.debug("No encryption by default, encrypt_by_default=%s",
                           account.encrypt_by_default)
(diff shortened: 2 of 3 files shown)


lamarpavel:
That is a very welcome feature, thank you.

dcbaker:
@meskio I'm interested in getting this landed (and have reviewer/merge powers), if you would be willing to rebase this I'd be happy to review it.

zenhack:
This introduces a pitfall that we might want to think about -- when a pgp message is encrypted to multiple recipients, each of them can view the other recipients from the pgp message itself, with no aid from the email headers.

This is a problem because the point of bcc is to hide some recipients from others.

I don't know the best way to address this, but it bares consideration. I suspect many users don't know this detail of pgp, so at least a warning would be prudent, if nothing else.

lucc:
Apparently there is a gpg option `--throw-keyids` which doesn't save the keyids in the gpg blob. That means that bcc keys can be added to the list of encryption keys. But it also means that the receiving gpg has to try all available secret keys because it can not make an id match first.

Searching the web for "gpg bcc" lead me to learn that it might be possible to send the message to each bcc recipient individually and hence hide the key from all other copies being send: http://www.gossamer-threads.com/lists/gnupg/users/65738#65738

pazz:
but this aproach would create new messages (with individual IDs) for each BCC and therefore somewhat clutter the notmuch index (notmuch identifies duplicates by ID).

zenhack:
Quoting Lucas Hoffmann (2016-12-18 04:01:38)

>    Apparently there is a gpg option --throw-keyids which doesn't save the
>    keyids in the gpg blob. That means that bcc keys can be added to the
>    list of encryption keys. But it also means that the receiving gpg has
>    to try all available secret keys because it can not make an id match
>    first.

Am I correct in understanding that this still exposes the *number* of
recipients? If so, this still leaks information, in that a recipient may
infer that there were bcc recipients, and how many.

zenhack:
Quoting Patrick Totzke (2016-12-18 05:42:24)

>    but this aproach would create new messages (with individual IDs) for
>    each BCC and therefore somewhat clutter the notmuch index (notmuch
>    identifies duplicates by ID).

Could we just not store the Bcc variants? i.e. we store the copy that is
only encrypted to the `to` and `cc` headers, and send -- but do not
store -- A copy encrypted additionally to each bcc recipient.

On the topic of different message IDs, why do they have to be different?
One thing that occurs to me is that that could muck with things like the
In-Reply-To header.

pazz:
Quoting Ian Denhardt (2016-12-18 18:12:11)
>    Quoting Patrick Totzke (2016-12-18 05:42:24)
> 
>    > but this aproach would create new messages (with individual IDs) for
>    > each BCC and therefore somewhat clutter the notmuch index (notmuch
>    > identifies duplicates by ID).
> 
>    Could we just not store the Bcc variants? i.e. we store the copy that is
>    only encrypted to the `to` and `cc` headers, and send -- but do not
>    store -- A copy encrypted additionally to each bcc recipient.

yes, that would be an option.
But in my own local copy I still want to see the information who I
BCC-d. Otherwise I will forget and send stuff multiple times..
 
>    On the topic of different message IDs, why do they have to be different?
>    One thing that occurs to me is that that could muck with things like the
>    In-Reply-To header.

Well, you don't reply-to-all to messages from someone who received a BCC right? 
But yes, these things become tricky

zenhack:
Quoting Patrick Totzke (2016-12-18 13:25:37)

> yes, that would be an option. But in my own local copy I still want
> to see the information who I BCC-d. Otherwise I will forget and send
> stuff multiple times..

How do you manage this now? Do we currently store the bcc info?

>    Well, you don't reply-to-all to messages from someone who received a
>    BCC right?

If the both the to and bcc recipients reply to you, then it gets weird
because you end up with messages that are actually replies to the same
message, but specify different message IDs.

dcbaker:
> How do you manage this now? Do we currently store the bcc info?
This interests me as well. I would assume that we store a single message for the BCC, and reply-all does something different based on whether you're the original sender, a TO or CC recipient or the BCC recipient. I would assume for the sender a reply-all would keep all TO, CC, and BCC addresses the same, a TO or CC would behave like normal without a BCC repient, and a reply-all from a BCC recipeient would only reply to the sender.

Is that a correct understanding?

lucc:
I just wrote a mail to myself and put myself in the bcc field. Here is what I found. In the end I have three messages in the notmuch mail dir. The original file has the bcc header so I (the sender) can still see whom I send the mail to. The other two mails don't have the bcc field so no recipient can see it.

All files have the same message ID.

I my msmtp (my sendmail replacement) log I can see that the mail is send to two addresses (so the to header and the bcc header are converted in the envelope to field). And the mail size that msmtp reports is exactly the size of the file stored after sending minus the bcc header and some carriage returns. So that is the state the mail is going out of alot.

(Are you interested in any other fields for this discussion?)

meskio:
I just rebased my code into the master.

I would propose to merge this (if needed I can remove BCC and just encrypt for CC) and open another issue to deal with the privacy issues of encryption for BCC. it will require some discussion on what to do and some extra complexity in the code. But if people prefer to hold this until we decide what to do with bcc I'll be ok with it.

zenhack:
I think it makes sense to take the Bcc out of this PR and merge it;
adding Cc is fairly simple and a strict improvement over the status quo.
We can figure out the trickier issue later.

Quoting Ruben Pollan (2016-12-31 16:20:28)

meskio:
I removed the BCC from the code and opened #949.

pazz:
@dkg what to you think? Push this first and amend your #915 or the other way around?

dcbaker:
@meskio I'm seeing the following error when trying to encrypt with your updated patch `(7, 55, u'Invalid value')`

meskio:
@dcbaker I don't see this error. Can you provide more details on how to reproduce it? Or the log when this happen to see why it happens?

dcbaker:
@meskio I can no longer reproduce that error, so consider my objection withdrawn.

meskio:
@pazz, @dkg: as #915 looks that needs a bit more of work to be merged I would propose to merge this one now and I'll try to give a hand in rebasing #915 on top of it.

dcbaker:
I'll second meskio's opnion that this is ready to land. I'm planning to merge this tomorrow morning (PST) [in about 24 hours] unless someone objects to it before then.

dkg:
I agree with this approach: merge this, and then re-base #915

PR #880, opened 2016-07-10
Add option to add a prefix line to signature

olorin:
Adds `signature_prefix` option (default `None`).

For example, if `signature_prefix` is set to "--", a signature file
containing the line "J. Hacker, head widget-frobber" would be rendered
(if included in the message body):

```
--
J. Hacker, head widget-frobber
```

This allows `alot` to emulate the default behavior of other MUAs (e.g.,
`mutt` or `sup`) - many other programs automatically assume a "--"
prefix for the signature file and so adding "--" to the file
itself causes the result to have two "--" lines in `mutt` or `xlock`.

### DIFF
--- a/alot/account.py
+++ b/alot/account.py
@@ -45,6 +45,9 @@ class Account(object):
     """filename of signature file in attachment"""
     signature_as_attachment = None
     """attach signature file instead of appending its content to body text"""
+    signature_prefix = None
+    """line to prefix signature with in body of message"""
+
     abook = None
     """addressbook (:class:`addressbook.AddressBook`)
        managing this accounts contacts"""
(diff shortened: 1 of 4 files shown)


olorin:
@pazz any interest in this PR, or should I drop it?

olorin:
@pazz this is cluttering up my open-PRs list so I'm going to close it. Feel free to reopen if you have time to look at it at some point.

pazz:
apologies for the wait. Question: is this about displaying mails that have text/plain parts that should be interpreted as signatures?

Because when creating a new mail, you have the option to either just append the sig-files content to the main "body" text of the mail or you could create a new text/plain part of the mail for this content.
As far as I understand your code, it simply prepends this string to your signature text. So this means the result should be no different from just putting that string into your signature file.

olorin:
Thanks for the response.

> apologies for the wait. Question: is this about displaying mails that
> have text/plain parts that should be interpreted as signatures?

No, it's about composition rather than display.

> As far as I understand your code, it simply prepends this string to
> your signature text. So this means the result should be no different
> from just putting that string into your signature file.

Yes, this is correct, the result is no different. To use my case as an
example, I keep my signature in `~/.signature`, which is read by several
programs (e.g., mail, news, screenlocker). The dominant behaviour in
other programs is to insert a separator (usually `--`) before appending
the signature to an outgoing message, so with the current behaviour I
end up with either no separator in `alot` or a doubled separator in the
other programs which read this file. I could just keep a separate copy
of my signature for `alot`, but I found the default behaviour quite
surprising so I thought it was worth submitting a patch.

--
OpenPGP: 6FB7 ED25 BFCF 3E22 72AE 6E8C 47D4 CE7F 6B9F DF57

pazz:
Ah I see, thanks for the clarification.
So in thin cas I'd rather avoid this PR, simply in order to prevent the
code from further cluttering up too much and because there is the
obvious easy solution of keeping separate signature files.
Thanks anyway,
p

Quoting Sharif Olorin (2016-11-22 00:00:10)

PR #884, opened 2016-07-15
Improve and fix building of docs

lucc:
I noticed some problems with the building of the docs when working on a new branch (saving cmd history to disk, PR coming soon).  I also notices this before: https://github.com/pazz/alot/pull/854#issuecomment-195501799  One of the commit here is also cherry picked from there.

About b5dd5b8:  Why were the generated files stored in git?  Was it necessary for readthedocs (maybe this is now fixed by 0f078ae)?  Otherwise I can replace b5dd5b8 with a commit that updates all generated files.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -4,6 +4,7 @@
 # You can set these variables from the command line.
 SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
+PYTHON        = python
 PAPER         =
 BUILDDIR      = build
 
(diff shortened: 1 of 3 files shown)


pazz:
the first two patches look good to me. And yes, as far as I remember, RTFD.org generates docs by checking out the git source and using sphinx on these files. It does not use the provided Make file and therefore will not generate the table files. So a built should fail unless these files are not checked in.

lucc:
How do you push the files to RTD? If we use travis we might be able to use the after_success hook. But you can also just cherry-pick these two commits that you like.

pazz:
RTD monitors the git repository via github webhooks, i.e., it pulls
updated versions when i push to github and then it starts rebuilding.

Quoting Lucas Hoffmann (2016-07-20 00:13:07)

lucc:
#912 supersedes this.

Issue #562, opened 2013-02-16
Persistent history


josch:
Store search and command history in a file so that they are available during the next alot invocation. This was also mentioned in #8.

lucc:
A simple implementation for this is in master now: #885 

Should this be kept open for the libreadline discussion or be closed?

pazz:
no need, we can always re-open.
If you find the time to check if things have changed around libreadline
let us know. That would certainly be superior than anything I cooked up
myself, but iirw, it was problematic that there was no "concurrent"
interface to libreadline..

Quoting Lucas Hoffmann (2016-12-19 06:33:41)

lucc:
Saved buffers can actually be implemented with the `initial_command` option in the config file:
```
initial_command = bufferlist; taglist; search foo; search bar; buffer 0
```
@pazz what do you think about a generic "Tipps, Tricks and Hacks" wiki page? I would put the above initial_command there. If it grows to long we can split it later (but we don't have to think about good structure now). I think the wiki is used very little and we can make more use of it.

pazz:
Quoting Lucas Hoffmann (2017-01-27 09:48:25)
>    Saved buffers can actually be implemented with the initial_command option
>    in the config file:
> 
>  initial_command = bufferlist; taglist; search foo; search bar; buffer 0


cool idea, i never thought of that.

>    [1]@pazz what do you think about a generic "Tipps, Tricks and Hacks" wiki
>    page? I would put the above initial_command there. If it grows to long we
>    can split it later (but we don't have to think about good structure now).
>    I think the wiki is used very little and we can make more use of it.

sure, why not. I like the idea of having the wiki for user contributed
stuff separate from the "official" docs.

josch:
@lucc wow I also never thought of that! I guess I never realized that I could put multiple commands into `initial_command`. This is not restoring the buffers I had open the last time I closed alot but for my use case this solution is even *better* because I might've inadvertedly closed a buffer that I actually want available by default. This solution gives me that ability.

A tips&tricks wiki page would be great and a good companion to this useful page: https://github.com/pazz/alot/wiki/Contrib-Hooks

lucc:
@josch it's already [here](https://github.com/pazz/alot/wiki/Tips,-Tricks-and-other-cool-Hacks).

josch:
@lucc thanks for taking care of that! I added two more entries. :)

PR #885, opened 2016-07-16
Save command line history across sessions

lucc:
Implements the first half of #562.

Is there some more documentation needed?

### DIFF
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -2,7 +2,9 @@
 # This file is released under the GNU GPL, version 3 or a later revision.
 # For further details see the COPYING file
 import logging
+import os
 import signal
+
 from twisted.internet import reactor, defer
 import urwid
 
@@ -72,6 +74,12 @@ class UI(object):
 
         signal.signal(signal.SIGINT, self.handle_signal)
         signal.signal(signal.SIGUSR1, self.handle_signal)
+
+        # load command history
+        self._cache = os.path.join(
+            os.environ.get('XDG_CACHE_HOME', os.path.expanduser('~/.cache')),
+            'alot', 'commandhistory')
+        self.commandprompthistory = self._load_history_from_file( self._cache)
 
         # set up main loop
         self.mainloop = urwid.MainLoop(self.root_widget,
@@ -664,3 +672,37 @@ class UI(object):
             if isinstance(self.current_buffer, SearchBuffer):
                 self.current_buffer.rebuild()
                 self.update()
+
+    def cleanup(self):
+        """Do the final clean up before shutting down."""
+        self._save_history_to_file(self.commandprompthistory, self._cache)
+
+    @staticmethod
+    def _load_history_from_file(path):
+        """Load a history list from a file and split it into lines.
+
+        :param path: the path to the file that should be loaded
+        :type path: str
+        :returns: a list of history items (the lines of the file)
+        :rtype: list(str)
+
+        """
+        if os.path.exists(path):
+            return [line.rstrip('\n') for line in open(path).readlines()]
+
+    @staticmethod
+    def _save_history_to_file(history, path):
+        """Save a history list to a file for later loading (possibly in another
+        session).
+
+        :param history: the history list to save
+        :type history: list(str)
+        :param path: the path to the file where to save the history
+        :type path: str
+        :returns: None
+
+        """
+        directory = os.path.dirname(path)
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+        open(path, 'w').write('\n'.join(history))
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -79,7 +79,8 @@ class UI(object):
         self._cache = os.path.join(
             os.environ.get('XDG_CACHE_HOME', os.path.expanduser('~/.cache')),
             'alot', 'commandhistory')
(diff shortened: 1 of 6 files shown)


alot/ui.py
```diff
@@ -1,9 +1,10 @@
 # Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
 # This file is released under the GNU GPL, version 3 or a later revision.
 # For further details see the COPYING file
-import urwid
 import logging
+import os
```
pazz: I'd much rather have this os and XDG related stuff in `alot.init` than here. The UI class is too big already anyway..

alot/ui.py
```diff
@@ -73,6 +75,12 @@ def __init__(self, dbman, initialcmdline):
         signal.signal(signal.SIGINT, self.handle_signal)
         signal.signal(signal.SIGUSR1, self.handle_signal)
 
+        # load command history
+        self._cache = os.path.join(
+            os.environ.get('XDG_CACHE_HOME', os.path.expanduser('~/.cache')),
+            'alot', 'commandhistory')
+        self.commandprompthistory = self._load_history_from_file( self._cache)
```
dcbaker: There's a stray space between load_history_from_file and self._cache

alot/ui.py
```diff
@@ -664,3 +672,37 @@ def handle_signal(self, signum, frame):
...
+
+    @staticmethod
+    def _load_history_from_file(path):
+        """Load a history list from a file and split it into lines.
+
+        :param path: the path to the file that should be loaded
+        :type path: str
+        :returns: a list of history items (the lines of the file)
+        :rtype: list(str)
+
```
dcbaker: Extra newline

alot/ui.py
```diff
@@ -664,3 +672,37 @@ def handle_signal(self, signum, frame):
...
+    def _save_history_to_file(history, path):
+        """Save a history list to a file for later loading (possibly in another
+        session).
+
+        :param history: the history list to save
+        :type history: list(str)
+        :param path: the path to the file where to save the history
+        :type path: str
+        :returns: None
+
```
dcbaker: and here as well.

alot/ui.py
```diff
@@ -664,3 +672,37 @@ def handle_signal(self, signum, frame):
...
+        """Load a history list from a file and split it into lines.
+
+        :param path: the path to the file that should be loaded
+        :type path: str
+        :returns: a list of history items (the lines of the file)
+        :rtype: list(str)
+
+        """
+        if os.path.exists(path):
+            return [line.rstrip('\n') for line in open(path).readlines()]
```
dcbaker: This will return None if os.path.exists is False, but the documentation says it will always return a list of `str`s. Also, this use of open is not correct, it'll leak a file descriptor until gc runs.

I think that this function should be:

```python
if os.path.exists(path):
    with open(path, 'r') as f:
        return [l.rstrip('\n') for l in f]
return []
```

Note that iterating a file object creates iterator (non-materialized) instance rather than creating a concrete list.

alot/ui.py
```diff
@@ -664,3 +672,37 @@ def handle_signal(self, signum, frame):
...
+        :type history: list(str)
+        :param path: the path to the file where to save the history
+        :type path: str
+        :returns: None
+
+        """
+        directory = os.path.dirname(path)
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+        open(path, 'w').write('\n'.join(history))
```
dcbaker: This will leak a file descriptor until the gc runs. Use a context manager instead:
```python
with open(path, 'w') as f:
    for l in history:
        f.write(l)
        f.write('\n')
```

This other advantage of this approach is that it avoids building a (potentially) large string in memory before writing it.

lucc:
@dcbaker I tried to address your comments.

lucc:
As this is a new feature I would wait for @pazz  to give his OK.

pazz:
I like the feature and the code. The only thing I'm not entirely convinced of is that all of this further blows up the UI class. It's not a bit issue really but can we stop and think if it makes sense to put these things elsewhere? does it make sense to put the history file loading etc in the widgets? I suppose not does it?

pazz:
one more thing: I think it'd make sense to log new features and config options (anything that affects the casual and updating user) in the NEWS  file.

alot/ui.py
```diff
@@ -741,4 +740,8 @@ def _save_history_to_file(history, path, size=-1):
         directory = os.path.dirname(path)
         if not os.path.exists(directory):
             os.makedirs(directory)
-        open(path, 'w').write('\n'.join(history))
+        # Write linewise to avoid building a last string in menory.
```
zenhack: I assume s/last/large/ was intended.

lucc: sure

lucc:
about the big ui class:
I think the first thing is that the ui class should contain the **U**ser **I**interface.  And I would argue that the history belongs to the prompt and that belongs to the ui. Therefore what I think is worth having a look at is splitting things out of the ui class: maybe a prompt class (which could handle the history itself) or a statusbar class. Maybe you have more ideas of that kind. That might result in more than one class in alot/ui.py

pazz:
I was thinking of sticking the history inside the `CompleteEdit` widget, or subblasses thereof..
But never mind for now, I'm OK with the current series as is. Thanks.

PR #886, opened 2016-07-17
Add a travis file, prepare to run tests on travis-ci.org

lucc:
From the existence of https://github.com/pazz/alot/tree/0.3.8-unittests I take it that it is also planned to run these test somewhere.  So I suggest to use travis and start with this travis file.  I already set up travis for my fork of alot so you can see the build for this PR here: https://travis-ci.org/lucc/alot/branches

I initially did some more tests with travis wich can be found in my fork (branches travis-orig and travis-multi-os).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ addons:
     packages:
       # The gpgme build files are needed by the gpgme python module.
       - libgpgme11-dev
+      # The notmuch libs are needed to actually run alot.  The notmuch package
+      # should pull the correct libs.
+      - notmuch
 
 install:
   # urwid needs to be installed first.  The installation of urwidtrees will
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,3 +73,5 @@ script:
   - make -C docs html
   # Minimal test: Start alot in order to quit immediatly.
   - alot --config conf
+  # Run the testsuite.
+  - python setup.py test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,17 @@
 language: python
-sudo: true
 
 python:
   # We can add more version strings here when we support other python
   # versions.
   - "2.7"
 
+addons:
+  apt:
+    packages:
+      # The gpgme build files are needed by the gpgme python module.
+      - libgpgme11-dev
+
 install:
-  - sudo apt-get update
-  # The gpgme build files are needed by the gpgme python module.
-  - sudo apt-get -yqq install libgpgme11-dev
   # urwid needs to be installed first.  The installation of urwidtrees will
   # fail otherwise.
   # TODO This should be fixed upstream.
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,11 @@ install:
   - pip install .
   # Install sphinx for building the documentation of alot.
   - pip install sphinx
+  # Prepare a minimal config file for the final test.
+  - echo 'initial_command=call os._exit(0)' > conf
 
 script:
+  # Build the documentation for alot.
   - make -C docs html
+  # Minimal test: Start alot in order to quit immediatly.
+  - alot --config conf
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,23 @@
+language: python
+sudo: true
+
+python:
+  # We can add more version strings here when we support other python
+  # versions.
+  - "2.7"
+
+install:
+  - sudo apt-get update
+  # The gpgme build files are needed by the gpgme python module.
+  - sudo apt-get -yqq install libgpgme11-dev
+  # urwid needs to be installed first.  The installation of urwidtrees will
+  # fail otherwise.
+  # TODO This should be fixed upstream.
+  - pip install urwid
+  # Install alot and the dependencies it declares.
+  - pip install .
+  # Install sphinx for building the documentation of alot.
+  - pip install sphinx
+
+script:
+  - make -C docs html
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,7 @@ install:
 
 before_script:
   # Prepare a minimal config file for the final test.
+  - touch ~/.notmuch-config
   - echo 'initial_command=call os._exit(0)' > conf
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,38 @@ python:
 addons:
   apt:
     packages:
-      # The gpgme build files are needed by the gpgme python module.
+      # The gpgme build files are needed by the gpgme python module
+      # (a dependency of alot).
       - libgpgme11-dev
       # The notmuch libs are needed to actually run alot.  The notmuch package
-      # should pull the correct libs.
-      - notmuch
+      # should pull the correct libs but currently the available version is
+      # not compatible with alot, so we have to build from source.
+      #- notmuch
+      # Dependencies to build notmuch from source.
+      - libxapian-dev
+      - libtalloc-dev
+      - zlib1g-dev
+
+# Build notmuch and the python notmuch libs manually.  The versions of the
+# notmuch library and the python module which are available in the 12.04 and
+# 14.04 Ubuntu repos do not match and do not fullfill the version requirement
+# for alot.
+before_install:
+  # Clone the notmuch repository and move into it.
+  - git clone git://notmuchmail.org/git/notmuch
+  - cd notmuch
+  # Make and install the library.  We install the library without sudo as we
+  # might want to switch to the travis container later.
+  - ./configure --prefix=$HOME/.local
+  - make
+  - make install
+  # Export the library search path.
+  - export LD_LIBRARY_PATH=$HOME/.local/lib
+  # Install the python bindings.
+  - cd bindings/python
+  - pip install .
+  # Move out of the notmuch dir again.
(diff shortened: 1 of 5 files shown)


lucc:
I was hoping to do something more thorough in the last commit but I could not think of any way to start alot and quit it immediatly afterwards (like `vim -c quit` for example). Setting `initial_command=exit` results in a blank screen. Only after I press `q` alot quits. This is the corresponding log:
```
DEBUG:manager:/home/luc/.config/alot/themes
INFO:ui:setup gui in 256 colours
DEBUG:ui:fire first command
DEBUG:ui:global command string: "exit"
DEBUG:__init__:mode:global got commandline "exit"
DEBUG:__init__:ARGS: [u'exit']
DEBUG:__init__:cmd parms {}
ERROR:ui:Could not stop reactor: Can't stop reactor that isn't running..
Shutting down anyway..
DEBUG:ui:Got key (['q'], [113])
DEBUG:ui:cmdline: 'exit'
DEBUG:ui:global command string: "exit"
DEBUG:__init__:mode:global got commandline "exit"
DEBUG:__init__:ARGS: [u'exit']
DEBUG:__init__:cmd parms {}
```

lucc:
I found a way to exit alot immediately, will clean up the git history next.

pazz:
cool! I have no idea how to use this yet but feel free to integrate the unitttest branch. I'm not working on this atm.

lucc:
In order to use this you have to got to http://travis-ci.org, sign in with you Github account and give travis access to this repository and configure it to run on pull requests. Then you should see the result here next to the merge-conflict-indicator.

If you don't say otherwise I will just cherry-pick your unittests.

pazz:
Quoting Lucas Hoffmann (2016-07-20 00:26:17)
>    In order to use this you have to got to [1]http://travis-ci.org, sign in
>    with you Github account and give travis access to this repository and
>    configure it to run on pull requests. Then you should see the result here
>    next to the merge-conflict-indicator.
> 
>    If you don't say otherwise I will just cherry-pick your unittests.

go ahead! thanks. I think i spend some time trying to set up a sensible
notmuch test database to deal with malformed mails. But it wasnt
completely finished..

lucc:
Do you want (me) to add more tests to this PR or are you considering it as is?

test/corpus.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self):
+        self.TESTDIR = tempfile.mkdtemp()
+        self.TESTCORPUS = path.join(self.TESTDIR, 'corpus')
+        HERE = path.dirname(path.realpath(__file__))
+
+        # copy over corpus
+        shutil.copytree(path.join(HERE,'corpus'), self.TESTCORPUS)
+
+        # create a temporary notmuch database
+        # create notmuch config file
```
toogley: That is a dependency on the filesystem and on notmuch. That means, this test is not an unit test, rather an functional test (i think). An Unit Test should, according to the book [The Art of Unit Testing](http://artofunittesting.com/), have no real dependencies on anything. The reason is that it dependends on the correct working of the filesystem and notmuch - so if the unit test fails, we don't know whether this bug is really in OUR codebase.

pazz:
The thing is that in order to test many features it is necessary to have some dummy data (e.g. email files) to test on.

lucc:
I am not sure if we can get rid of this filesystem dependency as even the test files will be *files*. Also the alot code base.

Also not sure if it is possible to get rid of the dependency on notmuch. If we want to argue that our code works we have to argue that it uses the notmuch bindings correctly (because that is what happens in production). And I can also try to continue this mocking idea ad absurdum: Why not mock the other python dependencies (urwid, urwidtrees, ...) or even python itself. 

Sure this is very artificial and I hope you don't take any offense from it, none was intended. I just have to learn some more about the principles behind unit testing (and have a look at that link).

We could for now also remove the unit testing commits from this PR and focus on getting the travis integration to work. It will then hopefully be easier to add linting/testing/whatnot jobs in the future.

pazz:
Sorry for my ignorance: Why would i want "travis integration"
if we already have some automated flake / pep8 comments
if not for running unit tests?

lucc:
@pazz you are right, I overlooked that quantifycode gives us linting. The only thing I can think of right now is that travis could run `make -C docs html && git diff --exit-code` in order to check that the last commiter did update *and commit* autogenerated documentation.

.travis.yml
```diff
@@ -0,0 +1,77 @@
...
+# for alot.
+before_install:
+  - sudo apt-get update
+  # Until https://github.com/travis-ci/apt-package-whitelist/issues/3215 is
+  # resolved we have to install gmime with sudo.  Afterwards we can swtich to
+  # the container based infrastructure.
+  # gmime is needed to build notmuch from source.
+  - sudo apt-get -yqq install libgmime-2.6-dev
+  # Clone the notmuch repository and move into it.
+  - git clone git://notmuchmail.org/git/notmuch
```
zenhack: Observation: this is always going to test against notmuch's master branch. We ought to choose explicitly what to test against. One approach: test against both master and the release branch, possibly not blocking on failures from master (since it's a moving target anyway, but at least we hear about breaking before it actually lands).

I ended up doing something like this with my golang binding to notmuch:

https://github.com/zenhack/go.notmuch/blob/master/.travis.yml

test/test_database.py
```diff
@@ -0,0 +1,17 @@
+import unittest
+from corpus import Corpus
+import os
+
+
+class TestDatabase(unittest.TestCase):
+    def setUp(self):
+        self.corpus = Corpus()
+
+    def test_messagefile(self):
```
zenhack: What is this test actually trying to verify? It doesn't actually invoke anything from alot.

pazz:
That's nice as well! Maybe that could be a start for a separate travis
integration, and a further patch can introduce the unit tests, and
another PR can then integrate those into travis?

lucc:
@pazz OK I tried to do that but it did not yet work. I was hoping that [job 74](https://travis-ci.org/lucc/alot/builds/182826784) would pass and [job 75](https://travis-ci.org/lucc/alot/builds/182827534) would fail but both passed.

I have trouble testing all this locally because the makefile in docs/ uses `python` but that is python3 on Arch Linux. I will try to fix that in a separate PR that should be merged first. It is here: #912

.travis.yml
lucc: that's cool. I will copy some of that to one of my travis related branches and will see when to integrate it (this PR or later). But I want to use this approach definitely.

lucc:
If we want to reduce the manual building of deps we can investigate https://github.com/travis-ci/travis-ci/issues/5821

zenhack:
Re: unexpected passes, `sphinx-build` has a `-W` that treats warnings as errors; we could make use of it.

lucc:
@zenhack I have see that too and was planing to use it but my comment is more about `git diff --exit-code` not returning failure though I assumed there would be changes to commit after the makefile updated something. I hope I fixed that in #914. I will use `-W` in the travis file afterwards.

zenhack:
Ah, I see -- I misinterpreted what you were referring to.

lucc:
I managed to do it on top of #914. See [build #89](https://travis-ci.org/lucc/alot/builds/182983622) and [build #90](https://travis-ci.org/lucc/alot/builds/182983882).

What I did to simplify and speed things up is this: Install the simple dependencies with pip and mock the complicated ones. Then delete all generated files and have the makefile regenerate them. Then use `git diff --exit-code` to see if the commited versions were up to date.

I can put the commits from https://github.com/lucc/alot/tree/travis-check-docs-3 into this PR as soon as #914 is merged.

The question that remains is this: Do we only want to check if the generated doc sources are up to date (like in the other branch right now) or do we also want to build the complete documentation? I would then remove `SPHINXBUILD=true` from the make call (don't be fooled by this argument to make, it doesn't tell make to build the docs with sphinx but to use the `true` command in order to build it, so actually it doesn't build the html docs :)

pazz:
@lucc (you didn't stick around long enough on irc for me to reply):
Generally, I prefer small PRs, so I would like to see one for introducing this travis file, maybe one or two "dummy" tests as individual patches, like the ones you have for testing if the docs build.
Maybe another patch that mentions travis in the docs, i.e., in the development section explains that we expect PRs to pass all tests etc.

I think adding a unit test suite should be another series, because what I wrote there was really preliminary and your travis stuff seems stable enough.
Also, conceptually these seem like different things: one contains (mostly) semantic tests around the code, the other is about running these (and others) on some server.

The only grey area is how/where to set up a test corpus. I tried to do something in this testsuite branch but its..well preliminary. But I think this should be done in some pre-hook of the test suite in order to run tests locally as well. If we put all this into the travis file then I see no easy way to test a PR before actually submitting it and waiting for travis.

lucc:
This was superseded/resolved/fixed/done by #916 and #919. So I will close it now.

Three interesting things can be done later:
- run tests against multible notmuch branches, see https://github.com/pazz/alot/pull/886#discussion_r91829569 (I have a branch ready for this but would wait for actual unit tests to be commited first)
- only use containers and no VMs, this is blocked by https://github.com/travis-ci/apt-package-whitelist/issues/3895, I have a branch ready to open a PR
- add unit tests, discussed in #917

PR #887, opened 2016-07-22
Add option to handle servers with non-standard case-insensitive matching

### LINKED ISSUE #847: To: field with upper case user name causes exception on reply.
I recieved an email with the user name part of my email address in upper case letters. When I view the email and hit `r` I get this exception:

```
Unhandled error in Deferred:


  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 588, in _runCallbacks
  File "/usr/lib/python2.7/site-packages/alot/ui.py", line 633, in call_apply
  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 1274, in unwindGenerator
  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 1128, in _inlineCallbacks
    result = g.send(result)
  File "/usr/lib/python2.7/site-packages/alot/commands/globals.py", line 846, in apply
    if self.encrypt or account.encrypt_by_default:
exceptions.AttributeError: 'NoneType' object has no attribute 'encrypt_by_default'
```

If I edit the mail file on disk and change the mail header line from `To: USER@domain.tl` to `To: user@domain.tl` I can use the `reply` command.

In the config file I identify this account with lower case letters: `address = user@domain.tl` and I have set `sign_by_default` and `encrypt_by_default`.

dcbaker:
Some email Servers (some versions of Exchange, possibly others) allow
matching email addresses without regard to case, this is problematic
because alot assumes the standard case sensitive behavior.

This patch addresses this issue by adding a new flag to the accounts
configuration "non_standard_case_insensitive", which defaults to False,
but when set to True will cause alot to match accounts without regard to
case. It does this is a non-intrusive manner using a subclass of unicode
with it's comparison methods overwritten to be case insensitive.

Fixes #847

### DIFF
--- a/alot/account.py
+++ b/alot/account.py
@@ -18,6 +18,132 @@ class StoreMailError(Exception):
     pass
 
 
+class _CaseInsensitiveUnicode(unicode):
+    """A subclass of unicode that matches case insensitive.
+
+    This behaves exactly like a standard Unicode object, except that it
+    overrides the rich comparison methods and __contains__ to do case
+    insensitve comparison.
+    """
+    # TODO: There are some cases that are not going to be handled correctly,
+    # the strip methods, for example, will return a standard Unicode object
+    # rather than a _CaseInsensitveUnicode.
+
+    def __lt__(self, other):
+        if isinstance(other, basestring):
+            return super(_CaseInsensitiveUnicode, self).lower() < other.lower()
+        return super(_CaseInsensitveStr, self).__lt__(other)
+
+    def __le__(self, other):
+        if isinstance(other, basestring):
+            return super(_CaseInsensitiveUnicode, self).lower() <= other.lower()
+        return super(_CaseInsensitveStr, self).__le__(other)
+
+    def __gt__(self, other):
+        if isinstance(other, basestring):
+            return super(_CaseInsensitiveUnicode, self).lower() > other.lower()
+        return super(_CaseInsensitveStr, self).__gt__(other)
+
+    def __ge__(self, other):
+        if isinstance(other, basestring):
+            return super(_CaseInsensitiveUnicode, self).lower() >= other.lower()
+        return super(_CaseInsensitveStr, self).__ge__(other)
+
+    def __eq__(self, other):
+        if isinstance(other, basestring):
+            return super(_CaseInsensitiveUnicode, self).lower() == other.lower()
+        return super(_CaseInsensitveStr, self).__eq__(other)
+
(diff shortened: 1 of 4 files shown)


dcbaker:
There are a couple of things about this patch that aren't complete, the _CaseInsenstiveUnicode probably should have the rest of the methods fixed, it might also be better to not make it a sublcass of unicode, but a wrapper around it.

There's also one change I'm not sure of, which is the replacement of 'myad in address' with 'myad == address'. I'm not sure why in is being used right now, except maybe to handle whitespace? It seems incorrect to use in for that since it could cause false positive matches.

I'm happy to clean this up and flesh it out more, but I wanted to make sure you were going to be okay with this approach before I completely finished it.

pazz:
sounds reasonable. sure go ahead. Does this also address the send-box issue mentioned in the referred thread?

dcbaker:
It only fixes the underlying issue if it's caused by the server passing a different case than alot is configured to know about. You could still reach that by trying to respond to an address you've never configured AFAICT.

dcbaker:
This should be fully fleshed out, and shouldn't have too many suprises if someone calls one of the methods that returns a copy. I've also included a patch with tests to prove it works. It uses pytest which I've found to be especially powerful and easy to work with, though I can port them to another framework if you have something else in mind for tests.

lucc:
There exists a branch on this repo where @pazz started to add a testing framework. I copied the commits to #886 as well. They use `unittest` so it might be worth it to convert your tests to that. But wait what @pazz says.

dcbaker:
Bump.

dcbaker:
Is there anything missing to get this merged? I've been running this branch for over a month now with no issues.

dcbaker:
So, I went and did some looking, it appears that while user names *may* be case insensetive (most are case insensitive, though some apparently are not) domain names must be treated as case insensitive. (https://stackoverflow.com/questions/9807909/are-email-addresses-case-sensitive). So I don't think this patch is really correct.

If I fixed this so that the domain name and the user name were stored separately (say as a tuple) and were compared separately would you be interested in merging such a change?

dcbaker:
I'm going to close this because it's not the right solution for either the dict lookup failing or the correct way to handle case sensitivity. I'll writie a proper patch and send a new merge request.

PR #893, opened 2016-09-01
add ctrl-g keybinding to cancel prompt 

bx:
I have found it hard to get to using Esc to cancel commands, and as far as I can tell that keybinding is baked-in and cannot be changed in the configuration file.  Thus I have created a patch that allows "ctrl-g" to be used  in addition to Esc (for cancelling commands) in order to allow it to be more friendly to those who have emacs keybindings built into their muscle memory.

### DIFF
--- a/alot/widgets/globals.py
+++ b/alot/widgets/globals.py
@@ -84,7 +84,7 @@ class CompleteEdit(urwid.Edit):
 
     The interpretation of some keypresses is hard-wired:
         :enter: calls 'on_exit' callback with current value
-        :esc: calls 'on_exit' with value `None`, which can be interpreted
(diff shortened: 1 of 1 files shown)


pazz:
Your're right: this should not be hardcoded.. anyway, thanks for the
patch and welcome to alot. Could you also update that class's docstring
accordingly?
cheers,
p

Quoting Rebecca Shapiro (2016-09-01 19:49:27)

bx:
No problem, I hope the new patch I pushed works for you. Thanks! I only started using alot a few weeks ago (having previously used sup) so there may be more pull requests in the future, I'm still getting used to alot.

Issue #873, opened 2016-06-09
ALOT and very large threads


bserrao:
I'm having trouble using alot because of very large threads (over 1000 messages).
This happens because i receive a lot of mail from people that compose similar mails (usually to the recipients) by replying to a previous mail and changing the subject. It's impossible for me to change this behaviour (belive me, i've tried) so i would like to find a solution for this...

1) Give the possibility to define a limit of messages to be show by thread. (most recent ones maybe?)
2) Have some kind of strict mode search that shows only the messages of the thread that really matches the query. For instance if the search is tag:inbox, show only the messages of the thread having this tag.
3) I belive this would be more a notmuch change than alot but... Give the possibility to group or break threads by thread_id, thread_id + subject, thread_id + folder/path, ...

Thanks a lot for your great work.

lucc:
I am thinking about a mutt feature that I would like to have in alot for a while now, which might help in this case. For me it is the "link threads" command but there is a corresponding "break thread" command.

Maybe we can propose two new commands to help in these kind of situations. As far as I can remember mutt did modify the actual message on disk and include or remove the needed `References` headers. That is contrary to the general approach of notmuch but maybe it can still be included in alot or we can up with some plugin code for the hooks file to do the job.

pazz:
I know this problem only too well. It is not a great solution but one
that pops to mind is writing a python hook that, for a selected msg,
removes the "In-Reply-To" header from the msg file and reindexes it.
This would effectively create a new thread with this message as root.

Quoting bserrao (2016-06-10 00:05:18)
>    I'm having trouble using alot because of very large threads (over 1000
>    messages).
>    This happens because i receive a lot of mail from people that compose
>    similar mails (usually to the recipients) by replying to a previous mail
>    and changing the subject. It's impossible for me to change this behaviour
>    (belive me, i've tried) so i would like to find a solution for this...
> 
>    1) Give the possibility to define a limit of messages to be show by
>    thread. (most recent ones maybe?)

Thats a great idea!

>    2) Have some kind of strict mode search that shows only the messages of
>    the thread that really matches the query. For instance if the search is
>    tag:inbox, show only the messages of the thread having this tag.

Yes, I was planning a "message search buffer" as opposed to the current
thread search for a while now, but haven't got around implementing it
apart from a PoC (i think).

>    3) I belive this would be more a notmuch change than alot but... Give the
>    possibility to group or break threads by thread_id, thread_id + subject,
>    thread_id + folder/path, ...

[...]

PR #896, opened 2016-09-30
new buffer to search and display messages

andresmrm:
This is a search buffer that list messages chronologically and displays a selected message. It can be used to display all messages from a thread in a way similar to Mutt's thread viewer:

search-messages-buffer

It can't display the thread tree like current thread buffer or the proposed retro thread buffer (#734, #740), but it seems to be much faster for big threads (#694, #873) and it doesn't have the indentation problem (>20th messages having a half screen left margin).

It still needs some work for better theme settings and more commands, but it seems usable. The code should also be much better than the one from retro thread PR.

I'm using these hotkeys for this buffer:

    [[searchmessages]]
      enter = select
      l = move next
      h = move previous
      r = reply --all
      R = reply
      n = editnew
      f = forward
      w = pipeto urlscan
      s = toggletags unread
      d = toggletags inbox
      a = save
      A = save --all
      c = togglesource

And to call it from search buffer:

	[[search]]
		enter = selectthreadmessages

I'll start to use this mode as my main thread viewer (instead of retro thread), let's see what bugs it has... =P
If anyone is in the mood to also test it: hope you enjoy it! =)

### DIFF
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -96,8 +96,13 @@ class Options(usage.Options):
     ]
     search_help = "start in a search buffer using the querystring provided "\
                   "as parameter. See the SEARCH SYNTAX section of notmuch(1)."
+    searchmessages_help = "start in a search messages buffer using the "\
+                          "querystring provided as parameter. See the SEARCH"\
+                          " SYNTAX section of notmuch(1)."
 
     subCommands = [['search', None, SearchOptions, search_help],
+                   ['searchmessages', None, SearchOptions,
+                    searchmessages_help],
                    ['compose', None, ComposeOptions, "compose a new message"]]
 
     def opt_version(self):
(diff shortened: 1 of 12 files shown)


geier:
This seems to work as intended for me. Some questions/suggestions:
* is the focus line gone on purpose? I believe it would be better to be consistent with the standard buffer here
* and to be consistent with mutt you could jump to the next/previous message if the end/beginning of the current mail has been reached

andresmrm:
Thanks for testing it, @geier!

> is the focus line gone on purpose? I believe it would be better to be consistent with the standard buffer here

You mean to have a focus line moving up/down while you scroll the email text? I would argue against that...

In the Retro Thread Buffer PR that I made before this rewrite, there was an unintended invisible focus line moving while scrolling (it happened because of the reuse of UI elements). It was bad because when switching from going up/down you had to wait the focus line to cross all the screen until the other lower/upper limit for the scroll to start again. Of course you could make it jump from the lower/upper limit to the other, but that seems to me a strange behavior for a focus line.

Besides that, the focus line seems useful for me when there is something to be selected, and that's not the case in the email text. But I did keep the focus line for the message list (top of the screen) and for the attachment list, that stays between the email metadata and text. The attachment list share the same move commands as the text scroll. Maybe I was a bit confusing, but try with an email with more than 1 attachment and enough text to need scroll... 

> and to be consistent with mutt you could jump to the next/previous message if the end/beginning of the current mail has been reached

Does Mutt do that? I didn't remember... It seems more confusing to me (I would switch emails without noticing it). But I could add this behavior as an option, if you like it. =)

PR #901, opened 2016-10-20
Removed incorrect backslash-t substitution in envelope.

xunam:
The substitution of the sequence `\t` (backslash followed by lowercase t, not tab) by four spaces when constructing a message alters the message contents (and makes no sense).

### DIFF
--- a/alot/db/envelope.py
+++ b/alot/db/envelope.py
@@ -171,7 +171,6 @@ class Envelope(object):
         # Build body text part. To properly sign/encrypt messages later on, we
         # convert the text to its canonical format (as per RFC 2015).
         canonical_format = self.body.encode('utf-8')
-        canonical_format = canonical_format.replace('\\t', ' ' * 4)
         textpart = MIMEText(canonical_format, 'plain', 'utf-8')
 
         # wrap it in a multipart container if necessary


pazz:
well, but this breaks the construction of encrypted/ signed mails.

This line should rather be moved inside some conditional:

```python
if if self.sign or self.encrypt:
    here.
````

**Comment by xunam** [comment]
There is a misunderstanding somewhere.

I am sorry but making this substitution optional would not fix anything. Replacing backslash-t by four spaces in a composed message is wrong, period. The message being signed or encrypted or anything else is irrelevant.

When I compose a message like
```
From: me@example.com
To: you@example.com
Subject: TeX code

To get italics, just type \textit{foo}.
```
I certainly do not want to send
```
From: me@example.com
To: you@example.com
Subject: TeX code

To get italics, just type     extit{foo}.
```

Encryption issues are totally unrelated with this issue.

I think this bug is a combination of two issues:

1. There is a substitution in the message *contents*, which is not acceptable in any way. Information-preserving substitutions can be acceptable at the encoding level (depending on `Content-Transfer-Encoding` and similar properties) but the actual contents must not be affected.

2. There is probably a confusion between backslash-followed-by-t and the tabulation character in the current code, because replacing something with spaces is only ever imaginable in an encoded form of a message and if this something is already whitespace. Nevertheless, this is only applicable if some RFC forbids the presence of tab at some point (I have not seen any such specification anywhere but I did not look too thoroughly).

Besides, I just tried with Mutt and it happily sends signed and encrypted messages containing the backslash-t sequence, as well as tabs, while preserving everything.

pazz:
@dkg can you comment on this? Will this affect the gpg stuff at all?

dkg:
I don't know of any encryption or signature that would require you to replace a literal `\t` sequence with a series of four spaces.

The comments around there refer to RFC 2015, which is a little worrisome, since i'm sure they were written after RFC 3156 came out, and 3156 is more current.  But neither of them suggest replacing `\t` with spaces, afaict.  is there some document that you think is relevant here?

this replacement was introduced in 40756e734ae4ec14d0a9705910b0d6385517d27f, but without any clear justification that i can see.  I think xunam's patch looks good.

pazz:
great, thanks. I just wanted to make sure.

PR #903, opened 2016-11-13
ui: Only update via SIGINT1 when in search buffer

### LINKED ISSUE #871: SIGUSR1 should only refresh search views?
My MDA (actually, notmuch new's post-new hook) is using SIGUSR1 to tell alot that new messages are available; when sitting at the search list this works well.

However, when reading a thread the external interrupt refreshes the buffer, meaning that I lose my cursor position and return to the top. It makes sense to refresh the thread _if_ there is a change in the thread (i.e. the new message that arrived was in this thread; although many other MUAs don't bother with a live refresh of a thread view at all)

Is there some solution like the post_buffer_focus hook mentioned in #828 ?

dset0x:
Refreshing non-search buffers (eg the thread view) can have undesired
effects. This defensively only refreshes when the search
buffer is visible. (fixes #871)

Signed-off-by: Dimitrios Semitsoglou-Tsiapos <kmhzsem@gmx.com>

### DIFF
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -7,7 +7,7 @@ import signal
 from twisted.internet import reactor, defer
 
 from settings import settings
-from buffers import BufferlistBuffer
+from buffers import BufferlistBuffer, SearchBuffer
 from commands import globals
 from commands import commandfactory
 from commands import CommandCanceled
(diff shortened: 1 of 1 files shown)


dset0x:
Note that this still brings the cursor to the first item in the buffer. A solution for that can be worked on at a later time.

pazz:
thanks. pushed to testing for now.

PR #906, opened 2016-11-30
Use python gpg

dkg:
GnuPG upstream is now producing a python module named "gpg", which is now shipping as part of the gpgme library.  It should get all improvements to gpgme automatically as gpgme is updated.

The pygpgme package is maintained by a third party and is not keeping up with newer versions of either GPGME or GnuPG itself.

Converting to python-gpg should be an improvement.

### DIFF
--- a/alot/crypto.py
+++ b/alot/crypto.py
@@ -171,23 +171,6 @@ def decrypt_verify(encrypted):
     return verify_result.signatures, plaintext
 
 
-def hash_key(key):
-    """
-    Returns a hash of the given key. This is a workaround for
-    https://bugs.launchpad.net/pygpgme/+bug/1089865
-    and can be removed if the missing feature is added to pygpgme.
-
-    :param key: the key we want a hash of
-    :type key: gpgme.Key
-    :returns: a hash of the key
-    :rtype: str
-    """
-    hash_str = ""
-    for tmp_key in key.subkeys:
-        hash_str += tmp_key.keyid
-    return hash_str
-
-
 def validate_key(key, sign=False, encrypt=False):
     """Assert that a key is valide and optionally that it can be used for
     signing or encrypting.  Raise GPGProblem otherwise.
--- a/alot/crypto.py
+++ b/alot/crypto.py
@@ -5,38 +5,7 @@ import os
 
 from cStringIO import StringIO
 from alot.errors import GPGProblem, GPGCode
-import gpgme
-
-
-def _hash_algo_name(hash_algo):
-    """
-    Re-implements GPGME's hash_algo_name as long as pygpgme doesn't wrap that
-    function.
-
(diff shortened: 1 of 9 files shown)


pazz:
thanks. Do you know if this ships with / is supported by majour distributions?

Quoting dkg (2016-11-30 07:44:23)

dkg:
sure, python-gpg 1.8.0 is in debian unstable now, and i intend it to ship with stretch (i'm part of debian's pkg-gnupg-maint team).

pazz:
Thanks for the clarification. I'll try this soon, and will await some
more comments re packaging in other distributions.

Quoting dkg (2016-11-30 12:33:11)

setup.py
```diff
@@ -32,6 +32,6 @@
         'twisted>=10.2.0',
         'python-magic',
         'configobj>=4.7.0',
-        'pygpgme>=0.2'],
+        'gpg'],
```
pazz: is there a minimal required version#?

pazz:
I like the fact that this cleans up a lot of the cluttered code. Its a nice patch :)

How would i test this on debian testing? `python-gpg` is not available there, only `python-gpgme`..

dkg:
You can install python-gpg from unstable.  It is built from the gpgme1.0 source package, which should have migrated by now, but is apparently blocked due to https://bugs.debian.org/846232

I'll upload a new version to unstable shortly to try to unblock that bug in hopes that it will migrate soon.

0x64746b:
Ftr: afaik @teythoon built the the pypi/pip package, so he should know details ;)

teythoon:
I did, and I am also maintaining the bindings upstream.  I support this change :)

pazz:
OK, I've decided not to push this before the upcoming release so that v0.3.8 can more easily make it into debian stable soon.
After the release, you're welcome to merge this yourself :)
However, I'd appreciate a few lines of install notes (for debian testing), even if it's just pointers..

pazz:
rebased: #915. closing this here

alot/crypto.py
```diff
@@ -181,15 +141,14 @@ def verify_detached(message, signature):
...
     :raises: :class:`~alot.errors.GPGProblem` if the verification fails
     '''
-    message_data = StringIO(message)
-    signature_data = StringIO(signature)
-    ctx = gpgme.Context()
+    ctx = gpg.Context()
     try:
-        return ctx.verify(signature_data, message_data, None)
-    except gpgme.GpgmeError as e:
+        (data, verify_results) = ctx.verify(gpg.Data(string=message), gpg.Data(string=signature), message_data)
```
dedekind: Where is 'message_data' defined now?

lucc: The code has since changed in master.

PR #911, opened 2016-12-10
docs: fix incorrect module/class names

zenhack:
Ran into this trying to update the AUR package for alot to 0.4:

https://aur.archlinux.org/packages/alot

Building the docs was failing.

Hopefully this can be squeezed into a fixup release? would prefer not to have to keep a backported version of this patch in the aur package.

### DIFF
--- a/docs/source/api/settings.rst
+++ b/docs/source/api/settings.rst
@@ -61,11 +61,16 @@ Accounts
 Addressbooks
 ------------
 
-.. module:: alot.addressbooks
+.. module:: alot.addressbook
 
 .. autoclass:: AddressBook
     :members:
-.. autoclass:: MatchSdtoutAddressbook
-    :members:
+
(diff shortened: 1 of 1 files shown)


lucc:
Looks good to me. Do you know sphinx and rst and can tell me the reason why some of the documentation is done with 
```rst
.. automodule:: alot.foo
    :members:
```
and some with 
```rst
.. module:: alot.bar

.. autoclass:: alot.baz
    :members:
```
?

zenhack:
I'd have to dig at it to figure it out. Do any of the ones that use autoclass actually have things in the modules that aren't specified in the rst?

pazz:
@lucc I believe that sometimes `automodule` was avoided in order to not clutter the docs with so far undocumented stuff..
In any case, this looks OK to me as well, I'll push this.

PR #912, opened 2016-12-10
Fix building of docs to ease development

lucc:
I am not sure why the NoneType error mentioned in the first commit exists.  Is it ok to fix it like this?

I hope this helps with #886.


lucc:
I just noticed that I did something similar in #884. I will have a look at the differences and put all the relevant commits here.

PR #914, opened 2016-12-11
Some more fixes for building docs

lucc:
Sorry there where quite some things I overlooked in #912.

### DIFF
--- a/docs/source/configuration/accounts_table
+++ b/docs/source/configuration/accounts_table
@@ -31,6 +31,16 @@
     :default: ,
 
 
+.. _alias-regexp:
+
+.. describe:: alias_regexp
+
+     a regex for catching further aliases (like + extensions).
+
+    :type: string
+    :default: None
+
+
 .. _sendmail-command:
 
 .. describe:: sendmail_command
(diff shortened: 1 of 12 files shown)


docs/source/configuration/accounts_table
```diff
@@ -47,7 +57,7 @@
 
      where to store outgoing mails, e.g. `maildir:///home/you/mail/Sent`.
      You can use mbox, maildir, mh, babyl and mmdf in the protocol part of the URL.
-
+
```
pazz: Nitpick: should we try to avoid these trailing whitespaces to be introduced?

lucc: Done.

lucc:
I used two different approaches to remove the trailing ws from the autogen. docs. Do you think this is ok or would you prefer to remove trailing ws in the second script as well (I think that might be a bit more difficult as with the first script).

lucc:
Also NB: The "Changes approved" mark from your review was not removed by github even though I pushed new commits.

pazz:
re reviews: yes, that's strange.. next time i'll simply hold back longer :)
I think the current way of dealing with trailing spaces is ok. Its not the end of the world anyway if one sneaks in after all..

PR #918, opened 2016-12-13
Pr/misc cleanups

dcbaker:
This is a bunch of misc cleanups of the code, which shouldn't have any behavioral changes (though I've only lightly tested, I'm dog fooding it right now). Most of these were caught by quantified code, though the first patch was generated through manual inspection.

There is one change I'm not completely sure about, and thats the change to Account.__init__, should 'sent' always be added to the sent_tags, or only if there are no tags specified (and the same with draft_tags)?

### DIFF
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -415,8 +415,9 @@ class SettingsManager(object):
         """returns addresses of known accounts including all their aliases"""
         return self._accountmap.keys()
 
-    def get_addressbooks(self, order=[], append_remaining=True):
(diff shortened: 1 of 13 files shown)


alot/settings/manager.py
```diff
@@ -403,7 +403,7 @@ def get_account_by_address(self, address):
         """
 
         for myad in self.get_addresses():
-            if myad in address:
```
dcbaker: since myad is already in user@domain form that wouldn't have matched, correct?

pazz: Oh yes, you are right. So let's go with your fix :)

Issue #917, opened 2016-12-12
Unit Tests


dcbaker:
Okay, lets discuss writing unit tests, since it sounds like we agree that is probably the next major milestone for alot.

Here's some thoughts I have from experience writing a lot of unit tests, and writing unit tests for a fairly mature project. Writing a ton of new unit tests is a daunting amount of work, but writing them incrementally here and there is much easier. I'd suggest that new features should be required to come with unit tests immediately, and that the tests are reviewed and part of the review process. I also suggest that we turn on unit tests in travis immediately, and enforce a policy of no unit test breaks across a series (since that makes bisecting hard). It may also be worth breaking out the unit test milestone into smaller chunks if we want to track it (say per module or per feature) with a milestone target, though I don't know how much of that kind of tracking we want to do.

The second thing I'll suggest is not using the built in unittest module for writing the tests. I've written a *lot* of tests, with unittest, nose, nose2 and pytest. I will say of those pytest is by far the best, nose2 is next (though last time I used it it was immature), nose, and then the builtin unittest module. The thing that makes pytest the best is that it is the most "batteries included" of the ones I've used, and removes the need to write a lot of boilerplate. I don't think its a problem (within reason) to pull in dependencies for testing. Mock is another module that is *incredibly* useful for writing good unit tests, though it's a bit tricky to know when to use mock.

Also, I think it would be worth while for some of us newer contributers to work on the unit tests since it will force us to learn more of the code base.

zenhack:
Quoting Dylan Baker (2016-12-12 14:30:42)

>    The second thing I'll suggest is not using the built in unittest module
>    for writing the tests. I've written a lot of tests, with unittest,
>    nose, nose2 and pytest. I will say of those pytest is by far the best,
>    nose2 is next (though last time I used it it was immature), nose, and
>    then the builtin unittest module. The thing that makes pytest the best
>    is that it is the most "batteries included" of the ones I've used, and
>    removes the need to write a lot of boilerplate. I don't think its a
>    problem (within reason) to pull in dependencies for testing. Mock is
>    another module that is incredibly useful for writing good unit tests,
>    though it's a bit tricky to know when to use mock.

I'm actually *really* not a fan of pytest. I've used it for a number of
projects, including things that I've maintained for years, and my
experience is that while it's really nice when you're getting started,
it gets to be a pain as you try to do more with it. There is generally
too much magic. It breaks namespacing in a number of places, which makes
maintaining fixtures in a module a pain. If the library itself doesn't
[...]

PR #919, opened 2016-12-13
Set up several jobs to build docs and run tests in parallel

lucc:
Two jobs are set up to run on travis.  Both run on Ubuntu trusty (14.04) and
python2.7.  One checks and builds the docs and the other is set up to run the
test suite (currently only a small config file is used to start and quit alot
again, much like `vim -c quit`).

The longer and more complicated shell scripts are still kept in the travis
file (with yaml multi line strings).  They get the shell option `set -e` to
get the "short circuit" behaviour otherwise present with single statements
(also known from make files) to fail the whole job as soon as any single line
fails.

Currently the tests are run in a full VM as we need sudo to manually build a
new enough version of notmuch.  This can be changed when this PR is resolved:
https://github.com/travis-ci/apt-package-whitelist/issues/3895

Follow up of #886, needed by #917.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,112 @@
 language: python
 
-python:
-  # We can add more version strings here when we support other python
-  # versions.
-  - "2.7"
+# We want the docs checks to run as fast as possible and with minimal setup,
+# hence they will be run in a container.  For the full build (needed for the
+# test runs) this is not possible as neither Ubuntu 12.4 nor 14.4 offer recent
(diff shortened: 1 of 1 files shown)


pazz:
the same strange issue with merges on github: it says "This branch is out-of-date with the base branch" but lists no conflicts.
Can I just hit "Rebase and merge"?

pazz:
Ah yes, the change of settings worked and this branch can now be merged directly..

PR #920, opened 2016-12-13
Small fixes found with pylint

lucc:


alot/errors.py
```diff
@@ -3,7 +3,7 @@
 # For further details see the COPYING file
 
 
-class GPGCode:
+class GPGCode():
```
dcbaker: That isn't a newstyle class, it needs to be ```python class GPGCode(object)```, right?

lucc: You are right. Only in python3 object is the default parent if none is given.

lucc:
@dcbaker I have addressed you comment.

lucc:
here it is the same question: why does github not allow us to merge this? it says  "This branch is out-of-date with the base branch" but there are no conflicts (it is just not a fast-forward merge. But when I use the merge button (as soon as it is available) it doesn't even create a fast-forwart merge even if that would have been possible.

@pazz is it possible to change that? (I am happy about forcing a merge commit when mergeing PRs, so doing a `merge --no-ff` but I am not happy that I have to rebase even in trivial situations)

pazz:
@lucc: I unticked the box for 

```Require branches to be up to date before merging This ensures the branch has been tested with the latest code on master.
```
in the settings for the master branch. This is the only setting that could be related. Let's see if it fixes the merging issue

lucc:
about the merge setting: What they mention in the setting that you unchecked makes actually sense (because the test run for the merge could fail even though it passes for both branches). But I am used to it like it is now and currently we don't have a real test suite so we can discuss it later if we want to.

About logging: It was actually reported by pylint I only got the hint about lazy string interpolation from quantifycode I think. But that is actually cool: If the logging interface decides it doesn't log some stuff (because the level doesn't fit, it will skip the string interpolation entirely. If we do the string interpolation before the call to logging it is always done.

That's the argument any way, but I will follow your decision whatever it will be.

pazz:
Let's wait for @dcbaker to remove his veto and then merge :)

dcbaker:
Everything looks good to me now. I'll go ahead and merge this.

PR #922, opened 2016-12-14
Use open() as a context manager

dcbaker:
This little series replaces most uses of open() that are not as a context manager (using the `with` statement) with context manager uses. Using open as a context manager has a number of advantages, it makes visual inspection of correctness easier, it ensures that the file is closed and the fd is released (even if an exception is closed which would otherwise require a try/finally construct), and it uses one line less code than `f = open(...); f.write(...); f.close()`.

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -324,10 +324,9 @@ class EditCommand(Command):
 
             # get input
             # tempfile will be removed on buffer cleanup
-            f = open(self.envelope.tmpfile.name)
             enc = settings.get('editor_writes_encoding')
-            template = string_decode(f.read(), enc)
-            f.close()
+            with open(self.envelope.tmpfile.name) as f:
+                template = string_decode(f.read(), enc)
 
             # call post-edit translate hook
             translate = settings.get_hook('post_edit_translate')
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -733,7 +733,8 @@ class ComposeCommand(Command):
                           priority='error')
                 return
             try:
-                self.envelope.parse_template(open(path).read())
+                with open(path) as f:
+                    self.envelope.parse_template(f.read())
             except Exception as e:
                 ui.notify(str(e), priority='error')
                 return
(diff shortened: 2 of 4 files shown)


alot/commands/envelope.py
```diff
@@ -324,10 +324,9 @@ def openEnvelopeFromTmpfile():
 
             # get input
             # tempfile will be removed on buffer cleanup
-            f = open(self.envelope.tmpfile.name)
-            enc = settings.get('editor_writes_encoding')
-            template = string_decode(f.read(), enc)
-            f.close()
+            with open(self.envelope.tmpfile.name) as f:
+                enc = settings.get('editor_writes_encoding')
```
lucc: I know it was in the middle before but does this line need to be in the middle or could it also be above the with block?

alot/commands/globals.py
```diff
@@ -733,7 +733,8 @@ def apply(self, ui):
                           priority='error')
                 return
             try:
-                self.envelope.parse_template(open(path).read())
+                with open(path) as f:
+                    self.envelope.parse_template(f.read())
```
lucc: Is it the pythonic way to use a with block even if the opened object is only used once? If I remember correctly all open objects are automatically closed on destruction and if there is no reference to them any longer (like in all these one line usages) there are no further references/names. Or are these objects only cleaned up when the next block or even the only if the function is finished?

dcbaker: They're not cleanuped up, they're marked for deletion (since there are 0 references, and CPython is reference counted). There are a couple of caveats to that though. The first is that until the gc runs it won't be closed. The second is that this only works in CPython because of the reference counting, other implementations (IronPython, Jython, Pyston, Pypy) don't reference count and may or may not collect the object properly (maybe too early, maybe never).

Here's a stack overflow about the question: https://stackoverflow.com/questions/1834556/does-a-file-object-automatically-close-when-its-reference-count-hits-zero

tl;dr: The existing code works because of a CPython implementation detail, and relying on implementation details is always bad practice in any language.

alot/commands/envelope.py
dcbaker: It could and should go outside the block, I'll update that.

alot/commands/globals.py
lucc: Thanks, in the meanwhile I also read both commit messages.

Now I am all for it!

pazz:
I'm with you.

PR #924, opened 2016-12-15
Use footnote-links to clean up badges

zenhack:
Per @lucc's comments in #923:

> The first few lines start to look patched together more and more. Maybe we can find a way to say
> things in a nicer looking way or we don't say everything at the top.
>
> But that can all be done in another PR. The badges are good and seem to point to the right place.

IMO footnote links make the source nicer to look at.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,15 @@
 Alot is an experimental terminal MUA based on [notmuch mail][notmuch].
 It is written in python using the [urwid][urwid] toolkit.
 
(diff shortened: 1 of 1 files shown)


pazz:
while you're at it: can you make sure that the headers are all at the same level?
And: Do we need the header "Badges" ?

zenhack:
Done. I also moved the badges up a bit; per the commit message this seems to be the conventional thing to do.

PR #928, opened 2016-12-16
unittests for helper module.

dcbaker:
This series is __not__ ready to land, or even close (though I did find some bugs already), but is really meant to help the discussion in issue #917. This adds a few tests for the helper library (which seemed like a good place to start adding tests), and implements them as both `unittest.TestCase` instances and as `pytest` cases.

The hope is that this will give a better view the differences between the two approaches to writing unit tests, the verbosity of `unittest` (which isn't as bad as I remember since setup.py has gained auto-discovery of tests) and the magic of `pytest` (magic in the "it magically works but no one really understands how" sense) such as dependency injection (the random and now parameters in the last class) and the way it's fixtures work (the use of autouse and scope, for example).

This series also makes use of `mock` to make the tests more reliable, it's a library that I've found incredibly helpful for writing reliable tests, and is part of the core library (as `unittest.mock`) in 3.3+.

### DIFF
--- a/tests/helper_test.py
+++ b/tests/helper_test.py
@@ -1,9 +1,29 @@
 # encoding=utf-8
+# Copyright © 2016-2017 Dylan Baker
+# Copyright © 2017 Lucas Hoffman
+
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """Test suite for alot.helper module."""
+
 from __future__ import absolute_import
 
+import datetime
+import random
 import unittest
+
+import mock
 
 from alot import helper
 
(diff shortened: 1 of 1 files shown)


lucc:
As we started to use unittest already I would say let's remove the pytest commit and merge this.

pazz:
the unittest version you mean?

lucc:
@pazz yes

dcbaker:
@lucc or @pazz, I've dropped the pytest patch and updated the commit message, would one of you like to review these for merge?

lucc:
About naming test classes and methods: I was trying to follow [pep8 naming conventions](https://www.python.org/dev/peps/pep-0008/#prescriptive-naming-conventions), I suggest you do that too with your class names. And I have read somewhere (don't remember where) that one should use verbose and expressive method/test names in order to hint what is tested and expected. See my #975 for an example.  I am not sure if there exist some official recommendations for this.

I will start a thorough review later.

tests/helper_test.py
```diff
@@ -0,0 +1,235 @@
...
+
+    # TODO: yesterday
+    # TODO: yesterday > now > a year
+    # TODO: last year
+    # XXX: when can the last else be hit?
+
+    @staticmethod
+    def __future_expected(test):
+        if test.strftime('%P'):
+            expected = test.strftime('%I:%M%P')
```
lucc: If I understand my manpage on strftime correctly the capital `%P` is a GNU extension and only the lower case `%p` is mentioned in the python docs on strftime: https://docs.python.org/2/library/datetime.html#strftime-strptime-behavior

So it might be better to use `%p` and call `.lower()` on the result.

tests/helper_test.py
```diff
@@ -0,0 +1,235 @@
...
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Test suite for alot.helper module."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
dcbaker: I can drop them.

__future__ imports are not like other imports though, they don't actually import code, they're directives to the interpreter that just happen to use the `import` syntax, but don't add code, they change how the interpret treats certain language constructs.

tests/helper_test.py
```diff
@@ -0,0 +1,235 @@
...
+        self._test(base, expected)
+
+    def test_unicode(self):
+        base = u'echo "foo";sleep 1'
+        expected = ['echo "foo"', 'sleep 1']
+        self._test(base, expected)
+
+    def test_escapes(self):
+        base = 'echo "foo \"bar\"";'
+        expected = ['echo "foo \"bar\""']
```
dcbaker: Nope, it's testing that the esacped `"` characters are not treated like the unescaped ones. Have a look at the first three lines of the split_commandline function and I think you'll see what I'm trying to test (though maybe I'm not doing a good job). It probably does deserve a comment though.

tests/helper_test.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class Test_pretty_datetime(unittest.TestCase):
+
+    # TODO: Currently these tests use the ampm format based on whether or not
+    # the testing machine's locale sets them. To be really good mock should be
+    # used to change the locale between an am/pm locale and a 24 hour locale
+    # and test both scenarios.
+
+    __patchers = []
```
dcbaker: It's a private class variable to be used by the class level fixtures (`setUpClass` and `tearDownClass`), and shouldn't be touched elsewhere. This is completely okay via PEP8, it invokes python's name mangling (in CPython this will be changed to `__Test_pretty_datetime_patchers`, which is the closest thing python has to a private attribute/method.

It's down near the bottom of [this section of PEP8](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles)

tests/helper_test.py
dcbaker: It might be better to not merge the tests for 'pretty_datetime' just yet, that one is only half-baked. I'll send an updated pull request with that test removed, and I'll clean it up for later. Apart from that problem (which was copied from alot) there are a number of cases marked as TODO, and a few other bits that need to be fixed.

dcbaker:
@pazz, @lucc. I know this has the potential to devolve into bikeshedding, but, here goes. Should we name tests using PEP8 style religiously, ie, a test for `my_func` would be called `TestMyFunc`, or should we be more clear about what is being tested and call it `Test_my_func`, even though we generally use PEP8 naming style?

tests/helper_test.py
lucc: I think I see what you want to test but I think you are confused by the escaping of double quotes in single quoted strings. Try to paste this into you interpreter: `'foo " bar' == 'foo \" bar'` (I get `True`). I also entered the debugger at line 55 and had a look at the variables after they are defined and there are no backslashes saved in the variables `base` or `expected`.

tests/helper_test.py
lucc: Yea I know about name mangling but [according to PEP8](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables) it should be used to protect attributes from being overwritten y subclasses. But will this class ever be subclassed?

tests/helper_test.py
dcbaker: ahhh, I always get bit by this. Too many years of ruby and shell scripts I guess. Yeah, it's not right

tests/helper_test.py
dcbaker: Probably not, but it also marks that this is private, don't depend on it and don't mess with it after initialization. I usually store mock patches in a mangled name so that if other people come along after me and want to muck with it after initialization they'll know not to, and that way we can have a `_patchers` attribute that is instance level, and is initated per test rather than once for all tests.

tests/helper_test.py
lucc: vim is even worse, try `:echo ''''` :)

tests/helper_test.py
lucc: I would be ok with merging the TODOs. (@pazz?)

tests/helper_test.py
dcbaker: I hate vimscript with a passion, It makes perl, php, and javascript seem nice

dcbaker: Actually, I think this test is bogus. Since we set `shlex.whitespace = ';'` and `shlex.split_whitespace = True` then the only case of spilts will be `;`, it will never split on tabs, spaces or newlines.

tests/helper_test.py
```diff
@@ -0,0 +1,233 @@
...
+        actual = helper.string_decode(base, encoding)
+        self.assertEqual(actual, expected)
+
+    def test_ascii_bytes(self):
+        base = u'test'.decode('ascii')
+        expected = u'test'
+        self._test(base, expected)
+
+    def test_utf8_bytes(self):
+        base = u'test'.decode('utf-8')
```
lucc: Maybe we can have a better string to be tested here, as `u'test'.decode('ascii') == u'test'.decode('utf-8')`

Also we should probably test with something other than unicode strings. They all hit the `except TypeError:` in line 111 in helpers.py. 

You can try it yourself:
```python
>>> from alot.helpers import string_decode
>>> from pdb import set_trace
>>> a = u'foobar'
>>> a == a.decode('utf-8')
True
>>> set_trace(); string_decode(a)
…
```

tests/helper_test.py
lucc: ok so lets leave it.

tests/helper_test.py
dcbaker: Gahh, python2 and it's enfuriating `str.encode` and `unicode.decode`! those should be `encode`, I've fixed them locally.

The only case we're not hitting (or should be hitting) is the malformed encoding case that causes the LookupError. And I'm not convinced that we should be handling that case really, and if we are going to handle it, we should probably use magic to guess the encoding rather than assuming that it's ascii.

tests/helper_test.py
```diff
@@ -76,3 +97,209 @@ def test_megabyte_numbers_are_converted_with_precision_1(self):
...
+        expected = expected.decode('utf-8')
+        return expected
+
+    def test_future_seconds(self):
+        test = self.now + datetime.timedelta(seconds=30)
+        actual = helper.pretty_datetime(test)
+        expected = self._future_expected(test)
+        self.assertEqual(actual, expected)
+
+    # Returns 'just now', instead of 'from future or something similar
```
lucc: the comment (each of these comments) is missing a closing single quote after "future".

tests/helper_test.py
```diff
@@ -76,3 +97,209 @@ def test_megabyte_numbers_are_converted_with_precision_1(self):
...
+
+
+class TestStringDecode(unittest.TestCase):
+
+    def _test(self, base, expected, encoding='ascii'):
+        actual = helper.string_decode(base, encoding)
+        self.assertEqual(actual, expected)
+
+    def test_ascii_bytes(self):
+        base = u'test'.decode('ascii')
```
lucc: Should this be encode or decode?

dcbaker: encode of course. It's not end of frustration that python to implements `str.decode`, which does absolutely nothing. Since a linter can't catch that it in fact does nothing.

lucc:
I approved this now. We can always add more tests for `string_decode` when bugs pop up.

PR #929, opened 2016-12-17
Use with blocks to write to temp files

lucc:
Like #922

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -375,12 +375,11 @@ class EditCommand(Command):
         old_tmpfile = None
         if self.envelope.tmpfile:
             old_tmpfile = self.envelope.tmpfile
-        self.envelope.tmpfile = tempfile.NamedTemporaryFile(delete=False,
-                                                            prefix='alot.',
-                                                            suffix='.eml')
-        self.envelope.tmpfile.write(content.encode('utf-8'))
-        self.envelope.tmpfile.flush()
-        self.envelope.tmpfile.close()
+        with tempfile.NamedTemporaryFile(
+                delete=False, prefix='alot.', suffix='.eml') as tmpfile:
+            tmpfile.write(content.encode('utf-8'))
+            tmpfile.flush()
+            self.envelope.tmpfile = tmpfile
         if old_tmpfile:
             os.unlink(old_tmpfile.name)
         cmd = globals.EditCommand(self.envelope.tmpfile.name,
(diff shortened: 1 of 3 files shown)


alot/db/utils.py
```diff
@@ -314,13 +314,12 @@ def extract_body(mail, types=None, field_key='copiousoutput'):
...
-                                                          prefix=prefix,
-                                                          suffix=suffix)
-                    # write payload to tmpfile
-                    tmpfile.write(raw_payload)
-                    tmpfile.close()
-                    tempfile_name = tmpfile.name
+                    with tempfile.NamedTemporaryFile(
+                            delete=False, prefix=prefix, suffix=suffix) \
+                            as tmpfile:
+                        # write payload to tmpfile
```
lucc: I think this comment is also not needed. Opinions?

alot/commands/envelope.py
```diff
@@ -375,12 +375,13 @@ def openEnvelopeFromTmpfile():
...
-                                                            suffix='.eml')
-        self.envelope.tmpfile.write(content.encode('utf-8'))
-        self.envelope.tmpfile.flush()
-        self.envelope.tmpfile.close()
+        with tempfile.NamedTemporaryFile(delete=False, prefix='alot.',
+                                         suffix='.eml') as tmpfile:
+            tmpfile.write(content.encode('utf-8'))
+            # Is this needed if the file is closed at the end of the with
+            # block?
+            tmpfile.flush()
```
lucc: Does one of you know if that is needed?

alot/db/utils.py
dcbaker: it seems pretty obvious to me what's going on here, I don't think we need the comment.

alot/commands/thread.py
```diff
@@ -944,13 +944,10 @@ def apply(self, ui):
...
 
-                tmpfile = tempfile.NamedTemporaryFile(delete=False,
-                                                      prefix=prefix,
-                                                      suffix=suffix)
-
-                tempfile_name = tmpfile.name
-                self.attachment.write(tmpfile)
-                tmpfile.close()
+                with tempfile.NamedTemporaryFile(delete=False, prefix=prefix,
+                                                 suffix=suffix) as tmpfile:
```
dcbaker: You're wrapped this and the one after in the opposite manner, from a consistancy point of view I'd prefer to see one or the other.

alot/commands/envelope.py
dcbaker: I would leave it, flush is basically a blocking operation to ensure that the file moves out of the harddrive cache and is actually written to disk (which is basically a  no-op on SSDs and NVMe).

lucc:
@dcbaker I amended with your comments and force pushed.

pazz:
thanks guys

PR #894, opened 2016-09-14
Add togglemimetree command in thread buffer

lucc:
This is an attempt to implement #862.

The new command `:togglemimetree` will (similar to `:togglesource` and friends) change the displayed message.  It will toggle between the previous used representation and a text representation of the mime tree.

TODO list (also see #862)
- [x] implement basic command logic
- [x] extract a mime tree from the message
- [x] add nicer tree visualisation (line drawings)
  - [ ] nicer indent, formatting and coherence with other ArrowTrees
- [ ] make the mime parts in the tree accessible for other commands (see below)
- [ ] add a command to display a mime part with an external program (mailcap entry)
- [ ] add a command to save a single mime part to disk
- [ ] add a command to pipe a mime part to a shell command

For the line drawing I could use something like https://github.com/jml/tree-format but as we already do display line drawings for the thread tree I suppose there is some code in alot that can maybe help me.  @ Everybody: do you know where it might be?

PR #933, opened 2016-12-20
Switch to container infrastructure on travis

lucc:
Travis advises to prefer them and they start up much quicker compared to full
VMs.

This is now possible because
https://github.com/travis-ci/apt-package-whitelist/issues/3895 has been
resolved.


lucc:
Yes we can try to use that two. Feel free to try it. I will try to work on #894 first.

PR #934, opened 2016-12-20
More cleanups

dcbaker:
This rather wondering series cleans up a number of issues I spotted by inspection, as a well as a number reported by pylint. The majority of the series is pretty trivial stuff, adding instance attributes to constructors, raw strings, dropping unneeded parenthesis, etc. And quite a few tests which optimize dictionaries by using iter methods or dropping the use of keys() altogether.

There are a couple of patches in here that refactor a method that stood out to me as more complicated than necessary, and add tests (since refactoring without tests is always a sketchy business).

### DIFF
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -144,7 +144,7 @@ class EnvelopeBuffer(Buffer):
         hidden = settings.get('envelope_headers_blacklist')
         # build lines
         lines = []
-        for (k, vlist) in self.envelope.headers.items():
+        for (k, vlist) in self.envelope.headers.iteritems():
             if (k not in hidden) or self.all_headers:
                 for value in vlist:
                     lines.append((k, value))
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -106,7 +106,7 @@ class BufferlistBuffer(Buffer):
 
     def get_selected_buffer(self):
         """returns currently selected :class:`Buffer` element from list"""
-        (linewidget, pos) = self.bufferlist.get_focus()
+        linewidget, _ = self.bufferlist.get_focus()
         bufferlinewidget = linewidget.get_focus().original_widget
         return bufferlinewidget.get_buffer()
 
@@ -279,7 +279,7 @@ class SearchBuffer(Buffer):
         returns curently focussed :class:`alot.widgets.ThreadlineWidget`
         from the result list.
         """
-        (threadlinewidget, size) = self.threadlist.get_focus()
+        threadlinewidget, _ = self.threadlist.get_focus()
         return threadlinewidget
 
     def get_selected_thread(self):
@@ -668,6 +668,6 @@ class TagListBuffer(Buffer):
 
     def get_selected_tag(self):
         """returns selected tagstring"""
-        (cols, pos) = self.taglist.get_focus()
+        cols, _ = self.taglist.get_focus()
         tagwidget = cols.original_widget.get_focus()
         return tagwidget.get_tag()
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -225,13 +225,13 @@ class SearchBuffer(Buffer):
     def __str__(self):
         formatstring = '[search] for "%s" (%d message%s)'
         return formatstring % (self.querystring, self.result_count,
-                               's' * (not (self.result_count == 1)))
+                               's' if self.result_count > 1 else '')
 
     def get_info(self):
         info = {}
         info['querystring'] = self.querystring
         info['result_count'] = self.result_count
-        info['result_count_positive'] = 's' * (not (self.result_count == 1))
+        info['result_count_positive'] = 's' if self.result_count > 1 else ''
         return info
 
     def cleanup(self):
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -303,8 +303,7 @@ class SearchBuffer(Buffer):
     def focus_last(self):
         if self.reversed:
             self.body.set_focus(0)
-        elif (self.result_count < 200) or \
-                (self.sort_order not in self._REVERSE.keys()):
+        elif self.result_count < 200 or self.sort_order not in self._REVERSE:
             self.consume_pipe()
             num_lines = len(self.threadlist.get_lines())
             self.body.set_focus(num_lines - 1)
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -85,7 +85,7 @@ class BufferlistBuffer(Buffer):
             self.isinitialized = True
 
         lines = list()
-        displayedbuffers = filter(self.filtfun, self.ui.buffers)
+        displayedbuffers = [b for b in self.ui.buffers if self.filtfun(b)]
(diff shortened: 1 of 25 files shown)


pazz:
wow! you've been busy :)

alot/buffers.py
```diff
@@ -85,7 +85,7 @@ def rebuild(self):
             self.isinitialized = True
 
         lines = list()
-        displayedbuffers = filter(self.filtfun, self.ui.buffers)
```
pazz: for those changes in this patch that replace the `map`s, OK, I kind of see your point that it may seem more natural in the way you proposed. But IMO `filter(filtfun, list)` is looks much cleaner than `[t for t in list]`..

alot/db/thread.py
```diff
@@ -151,13 +151,14 @@ def get_authors(self):
         :rtype: list of (str, str)
         """
         if self._authors is None:
+            # Sort messages with date first (by date descending), and those
+            # without a date last.
+            msgs = sorted(
```
pazz: very elegant! I'll have to test this but having only read the code, I like the change.

alot/buffers.py
dcbaker: I can go either way on the ``filter`` and ``map`` calls that don't have a lambda (the comprehensions are quite a bit faster if we can drop a lambda), I personally like the consistancy of doing it one way, but they are more verbose. There is one thing to think about as we look to python 3, ``map`` and ``filter`` return iterators in python 3 (they're basically ``itertools.imap`` and ``itertools.ifilter``), while comprhensions are the same between the two (other than that list comrehensions leak the last value calculated in python 2).

pazz:
regarding the `abc` stuff in f764d0a: is this part of the stdlib or a new dependency?
Secondly, should be do this in all abstract base classes then? (AddressBook and Account comes to mind)

alot/buffers.py
pazz: I'm OK with eliminating lambdas as you do. Where we use already defined functions to filter I'd rather use them in combination with filter instead of list comprehensions.
But this is really rather based on my personal stylistic preferences.
@lucc: you decide

dcbaker:
It's part of the standard library, so no new external dependencies.

As to whether we should use ABCMeta for other abstract base classes, yes, we probably should. I'll update the pull request.

alot/buffers.py
```diff
@@ -225,13 +225,13 @@ def __init__(self, ui, initialquery='', sort_order=None):
     def __str__(self):
         formatstring = '[search] for "%s" (%d message%s)'
         return formatstring % (self.querystring, self.result_count,
-                               's' * (not (self.result_count == 1)))
+                               's' if self.result_count > 1 else '')
```
lucc: Thank you for this!! I was nearly fainting when I just read the old code :)

lucc:
I have never before used abc so I have no idea about it (sorry pazz you have to decide on your own :P )

+1 for adding a test.

I had a look at the changes but have to stop now. Note to self, continue here: https://github.com/pazz/alot/pull/934/files#diff-f54191cc4a415fba8ab2eb8d203ccbfeL152

@dcbaker What is the difference between items() and iteritems() and keys() and iterkeys()?

dcbaker:
@lucc keys, values, and items return a copy of the underlying elements, while the iter versions return an iterator, which is a series of pointers back to the original object. The advantage is that the memory and CPU use is considerably less than creating a copy, the disadvantage is that iterators break in odd ways if the underlying object changes size while iterating over them.

alot/buffers.py
pazz: :D

alot/buffers.py
pazz: OK then, let's go with this.

0x64746b: Ftr only:

> Comments from GvR:  filter and map should die and be subsumed into list comprehensions

https://www.python.org/dev/peps/pep-0279/

If the expressions are too complex to be concisely represented in a comprehension, using the `itertools` alternatives mentioned by @dcbaker instead of the built-ins seems to be recommended [citation needed].

lucc:
I see some changes in author ordering in a search buffer. Is it possible to add the test at the beginning of this patch series so that we can see the point where it changes?

Two example messages I see this with have these headers:
```
Date: Sun, 11 Dec 2016 16:24:50 +0100
From: Me
```
and 
```
Date: Sun, 11 Dec 2016 15:50:08 +0000
From: You
```
On the current master branch they are shown as "Me, You" and on this PR it's "You, Me".

lucc:
PS: I have not set the autor order option in my config.

pazz:
Quoting Lucas Hoffmann (2016-12-20 23:13:09)
>    I have never before used abc so I have no idea about it (sorry pazz you
>    have to decide on your own :P )
> 
>    +1 for adding a test.

Is this some generic testing syntax or did you guys now agree on a
particular unit test suite?
should this test patch really be part of this cleanup series?

lucc:
no we did not agree on anything. It uses `unittest`.

I just like the fact he added a test as I think this is what we should do in the future. We can also decide to not merge the test now and maybe cherry-pick it later. I still would like to use it to debug the change I mentioned above.

pazz:
OK, cool with me

dcbaker:
@lucc I'm having trouble reproducing your problem, could you bisect which commit introduces the bad change? I'm assuming it's either ``db/thread: don't create multiple lists to sort`` or ``db/thread: Avoid intermediate data structure``. Either way it should be pretty trivial to bisect.

dcbaker:
@lucc nevermind. I can reproduce this. Turns out testing minutes and hours as well as days shows the error pretty readily.

dcbaker:
@lucc I've reordered the tests to the begining of the series as you requested, and I've hopefully fixed the issue you were seeing before.

lucc:
Yes I tried and the change in ordering is gone. Good work!

lucc:
Now we only have to decide if we merge the test with this or if we leave it out and cherry-pick it later if we have decided on a test framework.

Honestly I have no experience with any python testing framework so I will happily learn to use whichever you choose.

I suppose it is not even such a big problem to just start with unittest as pytest can run unittest tests. So if we decide to switch we can use all the unittest tests until we ported them.

dcbaker:
My thoughts on the matter is that porting a few tests from unittest to another framework (if that's what we decide) is easier than a wholesale port. Before we merge 100 tests we should decide what framework we're using. I think the benefits of having tests (even if we decide to change framework) in the repo is worth it, just to help catch any bugs, and to have people submitting tests with new features and refactors.

pazz:
I have no qualified opinion on tests suits of course, but I'm OK with using unittest for now.
We seem to already have some of those flying around and its certainly not some obscure unknown framework without long term future..

dcbaker:
I'm going to go ahead and merge this, with the unittests. I don't think adding a few fairly simple unittest based tests is going to create unimaginable amounts of work down the road, they're wont be that bad to port if we decide to port them.

PR #936, opened 2016-12-22
use unittest.mock instead of homebrew Mock modules

pazz:
.. in sphinx docs. see also
https://read-the-docs.readthedocs.io/en/latest/faq.html?highlight=autodocs#where-do-i-need-to-put-my-docs-for-rt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,8 @@ install:
   - |
     set -e
     if [[ $JOB = docs ]]; then
+      # install docs dependencies
+      pip install '.[docs]'
       pip install configobj twisted python-magic urwidtrees
       # Mock all "difficult" dependencies of alot in order to be able to import
       # alot when rebuilding the documentation.  Notmuch would have to be
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,28 +1,24 @@
 # -*- coding: utf-8 -*-
 # alot documentation build configuration file
 
-import sys, os
+import sys
+import os
 
 ###############################
 # readthedocs.org hack,
 # needed to use autodocs on their build-servers:
 # http://readthedocs.org/docs/read-the-docs/en/latest/faq.html?highlight=autodocs#where-do-i-need-to-put-my-docs-for-rtd-to-find-it
 
-class Mock(object):
-    def __init__(self, *args, **kwargs):
-        pass
+try:
+    from unittest.mock import MagicMock
+except ImportError:
+    from mock import Mock as MagicMock
 
-    def __call__(self, *args, **kwargs):
-        return Mock()
 
+class Mock(MagicMock):
     @classmethod
     def __getattr__(cls, name):
-        return Mock() if name not in ('__file__', '__path__') else '/dev/null'
-
-class MockModule(object):
-    @classmethod
-    def __getattr__(cls, name):
-        return Mock if name not in ('__file__', '__path__') else '/dev/null'
+            return MagicMock()
 
 MOCK_MODULES = ['twisted', 'twisted.internet',
                 'twisted.internet.defer',
@@ -34,12 +30,9 @@ MOCK_MODULES = ['twisted', 'twisted.internet',
                 'gpgme',
                 'configobj',
                 'validate',
-                'argparse']
-MOCK_DIRTY = ['notmuch']
-for mod_name in MOCK_MODULES:
-    sys.modules[mod_name] = MockModule()
-for mod_name in MOCK_DIRTY:
-    sys.modules[mod_name] = Mock()
+                'notmuch',
+                ]
+sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
 
 # end of readthedocs.org hack
 ##############################
--- a/setup.py
+++ b/setup.py
@@ -34,8 +34,13 @@ setup(name='alot',
         'configobj>=4.7.0',
         'pygpgme>=0.2'],
       tests_require=[
-          'mock',
+        'mock',
       ],
+      extras_require={
+        'docs': [
+            'mock',
+        ],
+      },
       provides=['alot'],
       test_suite="tests",
       )


dcbaker:
I really like this idea, but you'll need to ensure that mock is installed from 2.7 for this to work (for 3.3+ we can just use unitest.mock)

pazz:
Do you think it's alright to add simply add 'mock' to ` install_requires` in the setup.py?

dcbaker:
I think the right way to do that is to add something like this:
```python
setup(
    ...
    extra_requires={
        'docs': [
            'mock',
        ],
    },
    ...
)
```

then it can be installed via ``pip install '.[docs]'``, which we could add to the .travis.yml script.

With my distro-maintainer hat on: The problem with putting it in install_requires is that it's neither a build-time nor run-time dependency, it's needed to build the docs, which isn't something that a normal user is going to do when they pip install (or that I'm going to do for my users as a distro packager), so I'd carry a patch to remove the dep, as would most distros I think.

.travis.yml
```diff
@@ -98,6 +98,8 @@ install:
 script: |
   set -e
   if [[ $JOB = docs ]]; then
+    # install docs dependencies
+    pip install '.[docs]'
```
lucc: Can you move that to the install step above?

pazz:
meh! It looks like our homebrew variant was more versatile after all

docs/source/conf.py
```diff
@@ -1,28 +1,25 @@
 # -*- coding: utf-8 -*-
 # alot documentation build configuration file
 
-import sys, os
+import sys
+import os
 
+from alot import __version__,__author__
```
lucc: Will this find the correct files on disk? It should find the development code in ../../alot if we are in most cases (git checkout or tarball extract). But I assume it will find a installed version of alot in the system or user dirs first.

pazz: i guess you're right, yes.

pazz: (this import was here before, i just moved it up)

pazz:
I don't really see why there is a mock class involved at all at this point.


```
/home/pazz/projects/alot/docs/source/api/interface.rst:41: WARNING: autodoc: failed to import class u'UI' from module u'alot.ui'; the following exception was raised:
  File "/usr/lib/python2.7/dist-packages/sphinx/ext/autodoc.py", line 519, in import_object
    __import__(self.modname)
  File "/home/pazz/.local/lib/python2.7/site-packages/alot-0.4.dev0-py2.7.egg/alot/ui.py", line 13, in <module>
    from .commands import globals
  File "/home/pazz/.local/lib/python2.7/site-packages/alot-0.4.dev0-py2.7.egg/alot/commands/globals.py", line 41, in <module>
    class ExitCommand(Command):
  File "/home/pazz/.local/lib/python2.7/site-packages/alot-0.4.dev0-py2.7.egg/alot/commands/__init__.py", line 143, in __call__
    prog=self.name, add_help=False)
  File "/home/pazz/.local/lib/python2.7/site-packages/mock/mock.py", line 1062, in __call__
    return _mock_self._mock_call(*args, **kwargs)
  File "/home/pazz/.local/lib/python2.7/site-packages/mock/mock.py", line 1121, in _mock_call
    result = next(effect)
  File "/home/pazz/.local/lib/python2.7/site-packages/mock/mock.py", line 109, in next
    return _next(obj)
StopIteration
```

I also get this error locally, but removing `argparse` from `MOCK_MODULES` fixes he issue.
I suspect that the reason i mocked it originally was related to readthedocs.. teesting...

.travis.yml
pazz: can I just extend line 86?
I'm a little confused why this isnt `pip install mock` instead.

pazz:
right, so this seems to fix the travis build as well.
In fact, the builds on readthedocs showed this warning as well, but did not consider it as failing.

For some reason however, the (rtfd) build for the docs in this mock branch hang... I'll wait and see

docs/source/conf.py
lucc: Yes and at the place where it was before there was a modification to sys.path just one line above. The relevant directory to find the local version of alot was added but it is probably not in sys.path up here.

.travis.yml
lucc: I think you can. I assume it is more versatile like this. If we tell travis to `pip install mock` we have to change it if we ever change our requirements. If we tell travis to `pip install .[docs]` pip will look at our setup.py file and check the extra_requires stuff for a "docs" key (I assume). In this way we only have to change the setup.py if we change our requirements and it will this be already communicated to travis and any other (end) user using pip/setup.py (so hopefully everybody).

pazz: thanks that makes sense. So this means that we should only do `install '.[docs]'\
and put all the other extra requirements into the setup.py as well?
I moved this line up for now.

docs/source/conf.py
pazz: uh silly me. thanks. updated.

pazz:
there are conflicts? I've just rebased this to master!?

pazz:
aha, got it now :)

pazz:
@dcbaker it is `extras_require` (no trailing s) by the way:
http://peak.telecommunity.com/DevCenter/setuptools#declaring-extras-optional-features-with-their-own-dependencies

pazz:
i wonder why github still lists old commits here, that were removed or outdated by a force push

.travis.yml
lucc: I don't really understand you last question here. Are you just confirming and rewording what I tried to say or is there some opinion of mine that you query?

The new changes you made seem good.

lucc:
@pazz that is similar for me: Sometimes I have to refresh github tabs when I force push to a PR. Until I do so I see all the old and new commits in the discussion. So for me github seems to only half-update the PR page on a force push.

.travis.yml
pazz: oh i was refering to the other `pip install` in  https://github.com/pazz/alot/blob/master/.travis.yml#L77  this can be replaced? The current patch just adds another `pip install` command

pazz:
I do not want to merge this yet: this seems to break building the docs on readthedocs.net.
For some reason this branch does not finish the build:
https://readthedocs.org/projects/alot/builds/4831158/

I don't know what to do here

pazz:
I think I'm withdrawing this whole PR: with unittest.Mock,
the widgets are not correctly (not at all) documented via sphinx's `automodule` directive on readthedocs. See http://alot.readthedocs.io/en/moock/api/interface.html#widgets

This does work with the previous homebrew aproach. I cannot explain why and decided to just give up on this.

toogley:
@pazz But please don't close this PR, maybe someone finds the time to fix this PR someday. (Unfortunately, i can't help you here)

.travis.yml
lucc: If I understand the syntax for this `pip install .[whatever]` correctly then it is just a dict lookup in python. So we could add "alot" of names for all kind of dependency sets. But to add more special entries to the extra_requires dict does only make sense if that are groups of dependencies that we want to expose to users/package maintainers/...  So I would only do it for obvious things like "docs" and "tests".

IMHO The line 77 that you mention is something that is specific to our travis workflow. We only want to build the docs at this point and do not care to really run alot. Therefor we could actually mock all its dependencies by writing dummy python files. But then we would have to take care to define some empty classes and functions in these files which are imported by name from within alot. My approach was to mock the modules with annoying C dependencies which are not easily available on travis (compare line 78) and install all the other dependencies normally (but list them by hand because notmuch and gpgme are missing from the list). 

We could merge line 77 into line 84 though.

pazz:
Quoting Lucas Hoffmann (2016-12-27 11:14:51)
>    @lucc commented on this pull request.
> 
>    --------------------------------------------------------------------------
> 
>    In [1].travis.yml:
> 
>  > @@ -98,6 +98,8 @@ install:
>   script: |
>     set -e
>     if [[ $JOB = docs ]]; then
>  +    # install docs dependencies
>  +    pip install '.[docs]'
> 
>    If I understand the syntax for this pip install .[whatever] correctly then
>    it is just a dict lookup in python. So we could add "alot" of names for
>    all kind of dependency sets. But to add more special entries to the
>    extra_requires dict does only make sense if that are groups of
>    dependencies that we want to expose to users/package maintainers/... So I
>    would only do it for obvious things like "docs" and "tests".

Yes I agree
 
>    IMHO The line 77 that you mention is something that is specific to our
>    travis workflow. We only want to build the docs at this point and do not
>    care to really run alot. Therefor we could actually mock all its
>    dependencies by writing dummy python files. But then we would have to take
>    care to define some empty classes and functions in these files which are
>    imported by name from within alot. My approach was to mock the modules
>    with annoying C dependencies which are not easily available on travis
>    (compare line 78) and install all the other dependencies normally (but
>    list them by hand because notmuch and gpgme are missing from the list).

No: The thing is that the docs are build using sphinx's autodoc feature:
It imports the modules and extracts the docs from the docstrings.
This seems to fail for classes that are derived from things you mock out!
Have you actually had a look at those docs build on travis? I highly
suspect that the API docs are incomplete.
 
>    We could merge line 77 into line 84 though.

Since I'm abandoning this PR, there is nothing to do here of course :)

PR #939, opened 2016-12-27
Cache pip packages

dcbaker:
This saves having to rebuild the pip packages that require building (like twisted) every time a job is run. This can save a bit of time in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: python
 
+# Cache pip objects to speed up next build
+cache: pip
+
 # We need a newer version of zlib to build notmuch from source as is available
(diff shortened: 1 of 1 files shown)


lucc:
Does it also cache alot itself? Is it clever enough to recognize newer versions available or do we have to `install -U` everything now?

dcbaker:
> Does it also cache alot itself? 

No, only things installed from pypi.

> Is it clever enough to recognize newer versions available or do we have to pip install -U everything now?

Yes, it still goes out to pypi to ask for the latest version, but pip always checks the cache to see if such a version exists before downloading and building it.

PR #940, opened 2016-12-27
Even more cleanups

dcbaker:
The majority of the changes in this series are whitespace related, and after this series pylint's complaints are down basically to missing docstrings a few other small things here and there. However, this series also contains a few other cleanups and optimizations, including the use of more iterators and fewer (intermediate) concrete objects, not calling dunder methods directly when not necessary, removing useless calls, renaming non-PEP8 variables, and adding TODO's for future cleanups and optimizations (mostly those that are invasive enough that I didn't feel comfortable implementing without unittests.

There are two refactors in here as well, one replaces the `__cmp__` protocol with the rich comparison protocol (`__eq__`, `__lt__`, etc). This is both a performance optimization and is required for python 3, which doesn't have `__cmp__`. The second remove unnecessary concrete instances, and refactors to make use of `enumerate` and `reversed`.

I don't think any of this is too controversial, except for maybe the giant whitespace cleanup patch.

### DIFF
--- a/alot/db/message.py
+++ b/alot/db/message.py
@@ -246,10 +246,11 @@ class Message(object):
 
                 if cd.lower().startswith('attachment'):
                     if ct.lower() not in ['application/pgp-encrypted',
-                                  'application/pgp-signature']:
+                                          'application/pgp-signature']:
                         self._attachments.append(Attachment(part))
                 elif cd.lower().startswith('inline'):
-                    if filename is not None and ct.lower() != 'application/pgp':
+                    if (filename is not None and
+                            ct.lower() != 'application/pgp'):
                         self._attachments.append(Attachment(part))
         return self._attachments
 
--- a/alot/db/message.py
+++ b/alot/db/message.py
@@ -3,6 +3,7 @@
 # For further details see the COPYING file
 import email
 import email.charset as charset
+import functools
 from datetime import datetime
 
 from notmuch import NullPointerError
@@ -16,6 +17,7 @@ from ..settings import settings
 charset.add_charset('utf-8', charset.QP, charset.QP, 'utf-8')
 
 
+@functools.total_ordering
 class Message(object):
 
     """
@@ -60,10 +62,20 @@ class Message(object):
         """needed for sets of Messages"""
         return hash(self._id)
 
-    def __cmp__(self, other):
-        """needed for Message comparison"""
-        res = cmp(self.get_message_id(), other.get_message_id())
-        return res
+    def __eq__(self, other):
+        if isinstance(other, type(self)):
+            return self.get_message_id() == other.get_message_id()
+        return NotImplemented
+
+    def __ne__(self, other):
+        if isinstance(other, type(self)):
+            return self.get_message_id() != other.get_message_id()
+        return NotImplemented
+
+    def __lt__(self, other):
+        if isinstance(other, type(self)):
+            return self.get_message_id() < other.get_message_id()
+        return NotImplemented
 
     def get_email(self):
         """returns :class:`email.Message` for this message"""
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -110,7 +110,7 @@ def message_from_file(handle):
 
     # handle OpenPGP signed data
     if (m.is_multipart() and
-        m.get_content_subtype() == 'signed' and
+            m.get_content_subtype() == 'signed' and
             p.get('protocol') == app_pgp_sig):
         # RFC 3156 is quite strict:
         # * exactly two messages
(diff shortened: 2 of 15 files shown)


alot/db/message.py
```diff
@@ -60,10 +62,20 @@ def __hash__(self):
         """needed for sets of Messages"""
         return hash(self._id)
 
-    def __cmp__(self, other):
-        """needed for Message comparison"""
-        res = cmp(self.get_message_id(), other.get_message_id())
-        return res
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
```
lucc: Maybe `type(self)`? You gave the reason in your "remove dunder calls" commit.

alot/db/utils.py
```diff
@@ -110,7 +110,7 @@ def message_from_file(handle):
 
     # handle OpenPGP signed data
     if (m.is_multipart() and
-        m.get_content_subtype() == 'signed' and
+            m.get_content_subtype() == 'signed' and
```
lucc: I think if there is stuff after the opening brace then everything should align with the opening brace+1. That would mean that line 114 would have to take the old indent of 113 and not the other way round. Because that would result in exactly 4 spaces that's ugly because it aligns with the default indent of the inner block of the if.

Not really sure what to do here :/

alot/settings/theme.py
```diff
@@ -106,11 +106,11 @@ def matches(sec, thread):
         candidates = self._config['search'].sections
         for candidatename in candidates:
             candidate = self._config['search'][candidatename]
-            if candidatename.startswith('threadline') and\
-               (not candidatename == 'threadline') and\
-               matches(candidate, thread):
-                    match = candidate
-                    break
+            if (candidatename.startswith('threadline') and
+                    (not candidatename == 'threadline') and
```
lucc: mentioned before: strange 4 space align after `if (`. :/

docs/source/usage/modes/search.rst
```diff
@@ -94,10 +94,7 @@ The following commands are available in search mode
 
 .. describe:: toggletags
 
-    flip presence of tags on this thread.
-    A tag is considered present if at least one message contained in this
-    thread is tagged with it. In that case this command will remove the tag
-    from every message in the thread.
+    flip presence of tags on this thread. A tag is considered present if at least one message contained in this thread is tagged with it. In that case this command will remove the tag from every message in the thread.
```
lucc: I vote for some line length around 80 chars even in rst files.

alot/db/message.py
```diff
@@ -246,10 +246,11 @@ def get_attachments(self):
 
                 if cd.lower().startswith('attachment'):
                     if ct.lower() not in ['application/pgp-encrypted',
-                                  'application/pgp-signature']:
+                                          'application/pgp-signature']:
                         self._attachments.append(Attachment(part))
                 elif cd.lower().startswith('inline'):
-                    if filename is not None and ct.lower() != 'application/pgp':
+                    if (filename is not None and
+                            ct.lower() != 'application/pgp'):
```
lucc: same as with the other `if (`.

alot/db/utils.py
dcbaker: this is one of those wierd corners of PEP8, if you indent 4 spaces after an `if` you end up with this:
```python
if (foo and
    bar):
    print 'yes!'
```
PEP8 (https://www.python.org/dev/peps/pep-0008/#indentation) chooses not to wade into the mess too much, and says either make them flat as above, or push the second line of the if statement in and additional 4 spaces (total of 8). Personally, I think they're both ugly, btu the patch is much more readable than the example in this comment. Applies to the next comment of yours too.

docs/source/usage/modes/search.rst
dcbaker: So, the problem with that is it makes the docs in the python code and the rst not the same, which makes the docs tests fail. Maybe the right solution is to drop this change and find a better way to test the docs?

alot/db/message.py
dcbaker: Yes, that's much better. I'll change that.

docs/source/usage/modes/search.rst
dcbaker: either that or put the coresponding python code back? (or closer to the original). I don't really care which solution we go for.

lucc: I just saw that this file is auto generated. Forget my comment.

(damn I hate all these auto generated files that we have to keep in git)

alot/db/utils.py
lucc: ok, so lets leave it at you original change

pazz:
I didn't know about rich comparisons. Thanks.

Issue #942, opened 2016-12-28
Command line arguments are broken!


pazz:
I found that the `-n` switch seems problematic:

```
alot -n idontexist
Could not read idontexist and/or /home/pazz/.local/lib/python2.7/site-packages/alot-0.4.dev0-py2.7.egg/alot/settings/../defaults/notmuch.rc.spec
```

Also, the `-d` switch for the debug log level fails
```
alot -d debug
Traceback (most recent call last):
  File "/home/pazz/.local/bin/alot", line 11, in <module>
    load_entry_point('alot==0.4.dev0', 'console_scripts', 'alot')()
  File "/home/pazz/.local/lib/python2.7/site-packages/alot-0.4.dev0-py2.7.egg/alot/__main__.py", line 116, in main
    args.parseOptions()  # When given no argument, parses sys.argv[1:]
  File "/usr/lib/python2.7/dist-packages/twisted/python/usage.py", line 253, in parseOptions
    self._dispatch[optMangled].dispatch(optMangled, arg)
  File "/usr/lib/python2.7/dist-packages/twisted/python/usage.py", line 59, in dispatch
    value = self.coerce(value)
TypeError: debuglogstring() takes exactly 2 arguments (1 given)
```

lucc:
I did a git bisect and it seems to be 581ed2987bd456d2894637a30bb5a14a4caa5f9b that introduced this.

I did 
```sh
git bisect start master 0.4
git bisect run sh -c 'PYTHONPATH=. python2 -m alot -l out -d debug'
```

dcbaker:
I think just reverting my patch is the easiest solution. The changes were meant to be cosmetic only, if we want to fix them again we can revist them later.

pazz:
this was quick, thanks.
I wouldn't mind replacing the `twisted.usage` option parsing by argparse. The only reason why I originally didn't is that at that point (still?) argparse wasn't able to model subcommands (`alot search --help`..)

How would we test this?

dcbaker:
It's a bit of a functional test, but probably with `subprocess.check_call`.

argparse does handle subcommands, but there are some corners. If you're interested in moving to argparse I'd be willing to do that work. I've done a lot of work with argparse.

pazz:
@dcbaker sure, I'd be interested. Alot already uses argparse for commands internally, and I'm generally not a huge fan of twisted anyway. It's not a priority of course.

lucc:
[...]

PR #944, opened 2016-12-29
Replace twisted command line parsing with argparse

lucc:
As discussed in #942.

The command line interface is copied as directly as possible.  But at least the help output is formatted differently.  We still have to test this a litte bit to find out if I did not mess up some corner case with optional subcommands or some defaults for options.

I have some more ideas how to improve this and will try to add some more commits soon.  I will take care not to rebase so you can already have a look at what I put up already.

### DIFF
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -22,12 +22,12 @@ def main():
                         version=alot.__version__)
     parser.add_argument('-r', '--read-only', action='store_true',
                         help='open db in read only mode')
-    parser.add_argument('-c', '--config', type=argparse.FileType('r'),
-                        help='config file')
-    parser.add_argument('-n', '--notmuch-config', type=argparse.FileType('r'),
-                        default=os.environ.get(
+    parser.add_argument('-c', '--config', help='config file',
+                        type=lambda x: argparse.FileType('r')(x).name)
+    parser.add_argument('-n', '--notmuch-config', default=os.environ.get(
                             'NOTMUCH_CONFIG',
                             os.path.expanduser('~/.notmuch-config')),
+                        type=lambda x: argparse.FileType('r')(x).name,
                         help='notmuch config')
     parser.add_argument('-C', '--colour-mode',
                         choices=(1, 16, 256), type=int, default=256,
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -62,9 +62,8 @@ def main():
         root_logger.removeHandler(log_handler)
     root_logger = None
     numeric_loglevel = getattr(logging, options.debug_level.upper(), None)
-    logfilename = os.path.expanduser(options.logfile)
     logformat = '%(levelname)s:%(module)s:%(message)s'
-    logging.basicConfig(level=numeric_loglevel, filename=logfilename,
+    logging.basicConfig(level=numeric_loglevel, filename=options.logfile,
                         filemode='w', format=logformat)
 
     # locate alot config files
@@ -78,10 +77,11 @@ def main():
         alotconfig = options.config
 
     # locate notmuch config
-    notmuchpath = os.environ.get('NOTMUCH_CONFIG', '~/.notmuch-config')
     if options.notmuch_config:
-        notmuchpath = options.notmuch_config
-    notmuchconfig = os.path.expanduser(notmuchpath)
+        notmuchconfig = options.notmuch_config
+    else:
+        notmuchconfig = os.environ.get('NOTMUCH_CONFIG',
+                                       os.path.expanduser('~/.notmuch-config'))
 
     try:
         settings.read_config(alotconfig)
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -39,8 +39,10 @@ def main():
                         help='logfile [default: %(default)s]')
     # We will handle the subcommands in a seperate run of argparse as argparse
     # does not support optional subcommands until now.
+    subcommands = ('search', 'compose', 'bufferlist', 'taglist', 'pyshell')
     parser.add_argument('command', nargs=argparse.REMAINDER,
-                        help="possible subcommands are 'search' and 'compose'")
+                        help='possible subcommands are {}'.format(
+                            ', '.join(subcommands)))
     options = parser.parse_args()
     if options.command:
         # We have a command after the initial options so we also parse that.
@@ -48,10 +50,9 @@ def main():
         # command that will back this subcommand.
         parser = argparse.ArgumentParser()
         subparsers = parser.add_subparsers(dest='subcommand')
-        subparsers.add_parser('search',
-                              parents=[COMMANDS['global']['search'][1]])
-        subparsers.add_parser('compose',
-                              parents=[COMMANDS['global']['compose'][1]])
+        for subcommand in subcommands:
+            subparsers.add_parser(subcommand,
+                                  parents=[COMMANDS['global'][subcommand][1]])
         command = parser.parse_args(options.command)
     else:
         command = None
@@ -104,7 +105,7 @@ def main():
             cmdstring = settings.get('initial_command')
         except CommandParseError as err:
             sys.exit(err)
-    elif command.subcommand in ('compose', 'search'):
+    elif command.subcommand in subcommands:
         cmdstring = ' '.join(options.command)
 
     # set up and start interface
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -78,30 +78,20 @@ def main():
                         filemode='w', format=logformat)
 
     # locate alot config files
-    configfiles = [
-        os.path.join(os.environ.get('XDG_CONFIG_HOME',
-                                    os.path.expanduser('~/.config')),
-                     'alot', 'config'),
-    ]
-    if options.config:
-        expanded_path = os.path.expanduser(options.config)
-        if not os.path.exists(expanded_path):
-            msg = 'Config file "%s" does not exist. Goodbye for now.'
-            sys.exit(msg % expanded_path)
-        configfiles.insert(0, expanded_path)
+    if options.config is None:
+        alotconfig = os.path.join(
+            os.environ.get('XDG_CONFIG_HOME', os.path.expanduser('~/.config')),
+            'alot', 'config')
+        if not os.path.exists(alotconfig):
+            alotconfig = None
+    else:
+        alotconfig = options.config
 
     # locate notmuch config
     notmuchpath = os.environ.get('NOTMUCH_CONFIG', '~/.notmuch-config')
     if options.notmuch_config:
         notmuchpath = options.notmuch_config
     notmuchconfig = os.path.expanduser(notmuchpath)
-
-    alotconfig = None
-    # read the first alot config file we find
-    for configfilename in configfiles:
-        if os.path.exists(configfilename):
-            alotconfig = configfilename
-            break  # use only the first
 
     try:
         settings.read_config(alotconfig)
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -1,9 +1,10 @@
 # Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
 # This file is released under the GNU GPL, version 3 or a later revision.
 # For further details see the COPYING file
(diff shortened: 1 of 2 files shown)


lucc:
@pazz I added two more commits.

Yes the fixup commit should be squashed before merging (that's why I used the fixup prefix like `git rebase -i`) But as I did just promise that I will not rebase I thought it would be better like this. Will do before merge.

alot/__main__.py
```diff
@@ -13,115 +14,67 @@
...
-        args.parseOptions()  # When given no argument, parses sys.argv[1:]
-    except usage.UsageError as errortext:
-        print '%s' % errortext
-        print 'Try --help for usage details.'
-        sys.exit(1)
+    # set up the parser to parse the command line options.
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('-v', '--version', action='version',
+                        version='%(prog)s, {}'.format(alot.__version__))
```
dcbaker: This changes the behavior of the version flag, which currently just prints "0.4". I don't konw if it's a big deal, but I'm pointing it out.

alot/__main__.py
```diff
@@ -13,115 +14,67 @@
...
+    parser.add_argument('-r', '--read-only', action='store_true',
+                        help='open db in read only mode')
+    parser.add_argument('-c', '--config', type=argparse.FileType('r'),
+                        help='config file')
+    parser.add_argument('-n', '--notmuch-config', type=argparse.FileType('r'),
+                        help='notmuch config')
+    parser.add_argument('-C', '--colour-mode', choices=(1, 16, 256), type=int,
```
dcbaker: I'm going to get computer-sciency for a minute, appologies. This is not what a tuple is for, please use a list or set. Tuples are heterogenous collections, ie, tuples have structure; lists on the other hand are homogeneous, they are all of the same type (or at least they all quack) and have an order.

alot/__main__.py
```diff
@@ -13,115 +14,67 @@
...
+                        help='terminal colour mode.')
+    parser.add_argument('-p', '--mailindex-path', #type=directory,
+                        help='path to notmuch index')
+    parser.add_argument('-d', '--debug-level', default='info',
+                        choices=('debug', 'info', 'warning', 'error'),
+                        help='debug log [default: info].')
```
dcbaker: let's use the default flag and not put default in the help message. the default flag is wonderful :)

alot/__main__.py
dcbaker: lets also use the default flag.

alot/__main__.py
```diff
@@ -13,115 +14,67 @@
...
+    parser.add_argument('-l', '--logfile', default='/dev/null',
+                        type=lambda x: argparse.FileType('w')(x).name,
```
dcbaker: When I implemented this I didn't use FileType, the functions we're passing to accept strings. I really don't like the argparse.FileType because it's not clear when these are going to be released to gc.

alot/__main__.py
```diff
@@ -13,115 +14,67 @@
...
+                        help='logfile [default: /dev/null]')
+    parser.add_argument('command', nargs=argparse.REMAINDER)
```
dcbaker: rather than this, why not use parser.parse_known_args()?

alot/__main__.py
```diff
@@ -13,115 +14,67 @@
...
+    options = parser.parse_args()
+    print(options)
```
dcbaker: looks like a debug print got left behind.

alot/__main__.py
```diff
@@ -13,115 +14,67 @@
...
+        compose.add_argument('--to', help='recipients')
+        compose.add_argument('--cc', help='copy to')
+        compose.add_argument('--bcc', help='blind copy to')
+        compose.add_argument('--template', type=argparse.FileType('r'),
+                             help='path to template file')
+        compose.add_argument('--attach', type=argparse.FileType('r'),
+                             help='files to attach')
+
+        command = parser.parse_args(options.command)
+        print(command)
```
dcbaker: And another one.

pazz:
@dcbaker: what do you think of reusing the parsers used in command decorations?
Any technical reservations?
I kind of like the idea of avoiding code replication and also being
able to make global commands available on the command line.

alot/__main__.py
lucc: about default: what is the default here? I was assuming that it is dependent on the terminal type we detect and I hoped that urwid would do the right thing if we pass it None.

About being computer-sciency: I like that, don't apologize :) (but I did not get everything yet, see below)

About set: A set is a cool idea here. Will use it.

About the difference between list and tuples: I did not get that in total. I currently assume that python lists and tuples are essentially both linearly ordered aggregations/sequences/lists/you-give-it-a-mathematically-well-defined-name of arbitrary python objects, that allow for iteration (in for loops), random access with integer indices (in contrast to dicts which allow arbitrary indices), repeated access (contrast to generators, so they are stored and kept in memory and not generated on the fly), storing objects more than once in different positions (contrast to sets which allow each object only once). Possibly more? The difference being that lists are mutable (add/replace/remove elements) and tuples are not. So in order to relate this to you comment: how are lists more homogeneous than tuples? What kind of structure do tuples have that lists don't? Don't they both have an order? What does it mean that all elements of lists "quack"? (If this getting to long for a code-comment I am happy to move the discussion about lists vs. tuples to IRC.)

alot/__main__.py
lucc: I am using the default flag here. I was using a different help formatter before that did also print the default for every option defined so it was telling the end user that some options where set to `None` by default. I deemed that to be of no interest to the end user so I removed this help formatter. But for this and the logfile option I wanted to display the default to the user (like it was also done before) so I put it here manually. 

Do you agree that this is the right thing to do or can/should it be done differently/more elegantly?

alot/__main__.py
lucc: I used the argparse.FileType because it integrates nicely with usage messages if the user supplies bad files (not writeable/readable or so).

About your concerns: At least in the interactive python interpreter I get this result:
```python
>>> import argparse
>>> x = argparse.FileType('r')('somefile')
>>> # If I now `lsof -p $PID_OF_PYTHON` I see that somefile is opend
>>> del x
>>> # no lsof does not list it any longer
```
And as the FileType object is only used in the lambda to get the file name (as a string) I assume that the object is deleted afterwards. Is that right for lambdas? I know the lambda itself is kept in the parser object. But the object generated during its run should not be. The following test in the python interpreter seems to support this asumptions:
```python
>>> xx = lambda x: argparse.FileType('r')(x).name
>>> xx('somefile')
'somefile'
>>> # at this point lsof does not list somefile
>>> xx
<function <lambda> at 0x7f1596b51398>
```

alot/__main__.py
lucc: I will have a look at this and check the differences.

alot/__main__.py
lucc: done

alot/__main__.py
lucc: my bad, will change it if nobody comes forth and demands the new output to stay.

lucc:
@pazz did you find other commands that you want to have exposed?

alot/__main__.py
lucc: I just found a bug in the colourmode setting. It seems that the value from the config file is not used. I set it to 1 but still get colours, equally on master and this PR, so we might come back to that later.

pazz:
@lucc: well you are right, search and compose are the most important/sensible commands to expose.
maybe also `bufferlist,taglist,pysell`, but those are not so important and one could argue that they are better left hidden. I have no strong feelings about this. In case of doubt leave it as is..

The config file parameter still fails

```
alot -n ~/.notmuch-config
  File "/home/pazz/.local/bin/alot", line 11, in <module>
    load_entry_point('alot==0.4.dev0', 'console_scripts', 'alot')()
  File "/home/pazz/.local/lib/python2.7/site-packages/alot-0.4.dev0-py2.7.egg/alot/__main__.py", line 84, in main
    notmuchconfig = os.path.expanduser(notmuchpath)
  File "/usr/lib/python2.7/posixpath.py", line 254, in expanduser
    if not path.startswith('~'):
AttributeError: 'file' object has no attribute 'startswith'
```

alot/__main__.py
lucc: I found the problem with parser.parse_known_args(): With the REMAINDER stuff I can run `alot seach -h` and will get the help message of the search subcommand. With parse_known_args I will get the global help message instead. There will not be any possibly to force an option to be parsed by the subcommand parser as parse_known_args will eat *all* args it understands not only up to the first one it does not understand.

lucc:
@pazz do we even need to `expanduser()` the value given on the command line? I think we only need to do that for the default value where we use `~`. Right?

pazz:
I think we do need this, for some people (me) do type `alot -n ~/.notmuch.rc` and expect it to work.
If I understand the trace well, we somehow call `.startswith` on a non-string, maybe a file descriptor already?
So If the argument is declared a file then I would expect argparse to take care of the environment variable expansions etc., yes.

lucc:
It is a file object, yes. But that means that argparse did find a string on the command line which points to a readable file (argparse checks that). Hence your shell did already expand the ~ (which most UNIX shells do)

lucc:
@pazz I added a commit to expose the commands you mentioned so you can have a look.

lucc:
@dcbaker I tried to answer your comments and include some of the changes you suggested. What do you think?

About the help output difference: In general I like your version better but we have to write an extra help text for every command that we expose (they just became more). Also your output suggests that the subcommands are mandatory.

I would like to keep my handling of the alot config path as alot can run fine without a config file but your patch chokes on a missing alot config file. The notmuch config file is required so I could move the default value into the argparse parser setup. That would change the error message when the notmuch config file is missing from
```
Could not read /home/luc/.notmuch-config and/or alot/settings/../defaults/notmuch.rc.spec
```
to
```
usage: __main__.py [-h] [-v] [-r] [-c CONFIG] [-n NOTMUCH_CONFIG]
                   [-C {16,1,256}] [-p MAILINDEX_PATH]
                   [-d {debug,info,warning,error}] [-l LOGFILE]
                   ...
__main__.py: error: argument -n/--notmuch-config: can't open '/home/luc/.notmuch-config': [Errno 2] No such file or directory: '/home/luc/.notmuch-config'
```
Which error message looks better (I like the *code* for the second one better).

I notice that using `set()`s for the choices argument reorders the choices in the help output. Should we go back to tuples or lists?

dcbaker:
Sorry for the long reply. I'll start with the easy one. `set` will reorder the outputs (since it's an unordered type) if order is important `list` is the right type to use (As an aisde I really wish python had bulitin ordered sets, they're so useful).

I like the first better (though maybe it would be good to add the program name to the `ArgumentParser` instance so that it says 'alot' instead of '__main__.py'?) but if the code is ugly the second is fine.

I glanced over your changes and they seem reasonable.

lucc:
I squashed some of the small commits and rebased on current master. I removed the `set` for choices.

I added two new commits: the first one results in the second help output on missing notmuch config files and the second one releases all open config files after parsing.

@pazz any comment on the exposed command set?

pazz:
@lucc I honestly have no strong opinion on the exposed commands as I won't be using them a lot
(sorry for proposing extra work initially)

My only comments are:
- you might want to fix a typo in the commit msg of 1075a3f: "unneded" should by "unneeded",
- does the help stil lrefer to `__main__.py` instead of `alot`  as the runtime?

lucc:
Sorry I forgot to tell you about the program name in the help output. I tested it and the correct name is given if I install alot. The name `__main__` is only given if I run it from the git repository with `PYTHONPATH=. python2 -m alot --help`.

I will fix the typo.

pazz:
cool.
One more thing: please update docs/source/usage/synopsis.rst .
AFAIR, sphinx can even automagically generate this from argparse somehow.
But hardcoding it should be ok for now.

lucc:
simple enough

pazz:
nice work :)

PR #945, opened 2016-12-29
Remove some irc notifications by travis

lucc:
I did not find a setting to force every master branch build to send notifications. (maybe I missed something: https://docs.travis-ci.com/user/notifications/#IRC-notification) Is that still acceptable?

If we want we can also shorten the message that is send.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,3 +107,4 @@ script: |
 
 notifications:
     irc: "chat.freenode.net#alot"
+    on_success: change


dcbaker:
Just for posterity: I found this, which indicates that the travis-ci people are aware of the issue, but thy haven't implemented their proposed solution. There was one solution to secure the channel name, but I don't think that is really a good plan for an opensource project.

Issue #874, opened 2016-07-01
Trailing comma when group replying


josch:
Hi,

I'm running alot from git. When I group reply to an email with the following headers:

```
From: Santiago Vila <sanvila@unex.es>
To: Johannes Schauer <josch@debian.org>
cc: 825359@bugs.debian.org, deller@gmx.de
```

Then the following email will be generated:

```
Cc: 825359@bugs.debian.org, deller@gmx.de
To: Santiago Vila <sanvila@unex.es>,
From: Johannes Schauer <josch@debian.org>
```

Notice the trailing comma in the produced `To` header. The email I initially mentioned is publically archived here (this is also why I did not obfuscate the headers): https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=825359#29 By clicking the mbox link you can download the full email to see this effect yourself.

The trailing comma confuses some email clients (alpine). It is not necessary and should not be generated.

pazz:
I agree that comma shouldn't be there.
To find out where the error is introduced,
It'd help to add some debug logs to alot/commands/thread.py,
around line 231, to check the content of the `recipients` variable..

Quoting josch (2016-07-01 09:49:14)

josch:
Awesome! Thanks to your help, I got it now. The following code block:

```python
                # append To addresses if not replying to self sent message
                if sender_address not in my_addresses:
                    cleared = self.clear_my_address(
                        my_addresses, mail.get_all('To', []))
                    recipients.append(cleared)
```

wil append an empty string if `clear_my_address()` returns one. The problem would elegantly be solved if `clear_my_address()` would not return a string of already concatenated addresses but a list of addresses. Then one could just do a `recipients.extend(cleared)` and the problem would be fixed.

Incidentally, this is exactly what I proposed in issue #818 and thus the patch I proposed in #818 would also automatically fix this bug.

pazz:
Quoting josch (2016-07-01 10:18:10)
>    Incidentally, this is exactly what I proposed in issue [1]#818 and thus
>    the patch I proposed in [2]#818 would also automatically fix this bug.


Thanks. I will happily push these changes if you send a PR.
IIRC, #818 was loosely dependent on #823, which was solved (by you) but not
pushed yet :)

josch:
[...]

Issue #818, opened 2015-12-28
Make addresses in To header unique when group replying


josch:
Hi,

imagine an email with headers like this:

```
From: foo@bar.com
To: blub@bla.com, plim@blub.com
Reply-To: foo@bar.com, blub@bla.com
```

If I'd now hit reply-all for such an email, I'd get a message with the following To header:

```
foo@bar.com, blub@bla.com, plim@blub.com, foo@bar.com, blub@bla.com
```

As you can see, foo@bar.com and blub@bla.com are repeated. This is because there is no attempt at sending the list of addresses through a `sorted(set())` call but even with making such call, the addresses would not be unique-y-fied because they get pushed into the resulting list as groups of addresses and not as individual addresses.

The following patch sketches a possible solution:

``` patch
--- alot/commands/thread.py 2014-08-02 18:26:46.000000000 +0200
+++ alot/commands/thread.py.bak 2015-12-28 03:46:23.144234416 +0100
@@ -181,7 +184,7 @@
                 % recipients
             logging.debug(emsg)
         else:
-            recipients = [sender]
+            recipients = self.clear_my_address([], [sender])

         if self.groupreply:
             # make sure that our own address is not included
@@ -189,8 +192,8 @@
             MFT = mail.get_all('Mail-Followup-To', [])
             followupto = self.clear_my_address(my_addresses, MFT)
             if followupto and settings.get('honor_followup_to'):
-                logging.debug('honor followup to: %s', followupto)
-                recipients = [followupto]
+                logging.debug('honor followup to: %s', ', '.join(followupto))
+                recipients = followupto
                 # since Mail-Followup-To was set, ignore the Cc header
             else:
                 if sender != mail['From']:
@@ -200,15 +203,27 @@
                 if sender_address not in my_addresses:
                     cleared = self.clear_my_address(
                         my_addresses, mail.get_all('To', []))
-                    recipients.append(cleared)
+                    recipients.extend(cleared)

                 # copy cc for group-replies
                 if 'Cc' in mail:
                     cc = self.clear_my_address(
                         my_addresses, mail.get_all('Cc', []))
-                    envelope.add('Cc', decode_header(cc))
+                    envelope.add('Cc', decode_header(', '.join(cc)))

-        to = ', '.join(recipients)
+        def make_unique(value):
+            res = dict()
+            for name, address in getaddresses(value):
+                res[address] = name
+            new_value = []
+            for address, name in res.items():
+                if name != '':
[...]

PR #947, opened 2016-12-30
Make addresses in To header unique when group replying 

pazz:
This is just relaying the patch proposed in #818. related: #874.

`make_unique` should at least be documented properly, maybe made into a classmethod.
 The docstring of `clear_my_address` need to be changed.

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -231,7 +231,20 @@ class ReplyCommand(Command):
                         my_addresses, mail.get_all('Cc', []))
                     envelope.add('Cc', decode_header(', '.join(cc)))
 
-        to = ', '.join(recipients)
+        def ensure_unique_address(recipients):
+            """
+            clean up a list of name,address pairs so that
+            no address appears multiple times.
+            """
+            res = dict()
+            for name, address in getaddresses(recipients):
+                res[address] = name
+            urecipients = ['"{}" <{}>'.format(n, a) if n != ''
+                           else a
+                           for a, n in res.iteritems()]
+            return sorted(urecipients)
+
+        to = ', '.join(ensure_unique_address(recipients))
         logging.debug('reply to: %s', to)
 
         if self.listreply:
@@ -295,7 +308,7 @@ class ReplyCommand(Command):
                     new_value.append('"%s" <%s>' % (name, address))
                 else:
                     new_value.append(address)
-        return ', '.join(new_value)
+        return new_value
 
 
 @registerCommand(MODE, 'forward', arguments=[
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -204,7 +204,7 @@ class ReplyCommand(Command):
                 % recipients
             logging.debug(emsg)
         else:
-            recipients = [sender]
+            recipients = self.clear_my_address([], [sender])
 
         if self.groupreply:
             # make sure that our own address is not included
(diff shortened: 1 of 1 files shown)


alot/commands/thread.py
```diff
@@ -221,15 +221,27 @@ def apply(self, ui):
...
                         my_addresses, mail.get_all('Cc', []))
-                    envelope.add('Cc', decode_header(cc))
+                    envelope.add('Cc', decode_header(', '.join(cc)))
+
+        def make_unique(value):
+            res = dict()
+            for name, address in getaddresses(value):
+                res[address] = name
+            new_value = []
+            for address, name in res.items():
```
dcbaker: you should use `res.iteritems()` instead of `res.items()`

alot/commands/thread.py
```diff
@@ -221,15 +221,27 @@ def apply(self, ui):
...
+                if name != '':
+                    new_value.append('"%s" <%s>' % (name, address))
+                else:
+                    new_value.append(address)
```
dcbaker: What about using a comprehension?
```python
sorted(['"{}" <{}>'.format(n, a) if n != '' else a for n, a in res.iteritems()])
```
Or do you think having the ternary in the comprehension is too complex?

pazz: i wonder  why he used `sorted` here anyway..

alot/commands/thread.py
pazz: ok

alot/commands/thread.py
pazz: i used your comprehension now, please check if the formatting makes sense.

pazz:
oh wait! something is wrong! the current branch here mixes up address and names for a subset of the recipients when group replying!

pazz:
ah ok, (iter)items() returns pairs key,value not the other way around.
I'm happy with this now.

pazz:
the docs tests seem to fail here, and i cannot see why. This should be totally unrelated. Bug in the travis setup?

```
Warning, treated as error:

/home/travis/build/pazz/alot/docs/source/usage/commands.rst:27: WARNING: usage/modes/global is already assigned section numbers (nested numbered toctree?)
```

lucc:
I can not reproduce this locally. You can try to change something in line 102 of the travis file. One thing could be to `make cleanall html` instead of just `html` and you could increase the verbosity of sphinx-build with one or more `-v` options.

pazz:
I also cannot reproduce this locally. the verbose debug message does not really tell me anything as well.
I've split the commit into two smaller ones, but each one makes travis fail the docs build.
This looks completely unrelated to me!

pazz:
@lucc: the craziest thing is that the docs also fail on a new test branch (off master, same HEAD), whereas not on master itself.

lucc:
Maybe there is something broken in the cache? Can you clear the cache?

dcbaker:
I deleted the cache on one of the branches and restarted it. We'll see if that fixes the issue.

dcbaker:
There's a new version of sphinx. I wonder if that's the problem? 1.5.2 vs 1.5.1

dcbaker:
To be clear, clearing the cache did not solve the problem

lucc:
I had a look at the actual sphinx warning 
```
SphinxWarning: /home/travis/build/pazz/alot/docs/source/usage/commands.rst:27: WARNING: usage/modes/global is already assigned section numbers (nested numbered toctree?)
```
and the lines 27+ in this file are not generating any html output as far as I can see. So I removed them but then I get another warning:
```
SphinxWarning: /home/luc/vcs/alot/docs/source/usage/modes/bufferlist.rst:: WARNING: document isn't included in any toctree
```
Maybe there is something wrong with this part of the file. But I don't know rst or sphinx so I hope you have an idea how to fix that.

pazz:
OK, we can treat the travis/docs issue separately: I just triggered a rebuild on the master branch and get the same problem. I believe this was indeed related to the cache somehow hiding the fact that an earlier commit already resulted in a fail. #994

pazz:
@dcbaker I addressed your previous review, can you have another look and OK/merge this one then?

dcbaker:
lgtm

pazz:
thanks

Issue #664, opened 2013-10-31
Mouse event support via urwid


sjackso:
Urwid can deliver mouse and scroll wheel events.  It would be neat if alot could accept these.

@pazz, would you be open a pull request for such a feature?

dcbaker:
@pazz. Do you want to leave this open or close it with #952 merged?

pazz:
Let's close it, I'm quite happy with the current patch and
it clearly hasn't been a pressing issue in the past years :)

#close

PR #952, opened 2017-01-09
adding support for mouse scrolling

samdmarshall:
This adds basic support for scrolling with the mouse wheel, addressing #664.

The changes to ui.py on line 160 feels super hacky but it seems that the format of the events from urwid change based on if they are keyboard bindings or from a mouse. I think that should be able to handle all of the use cases (i wasn't able to get it to crash, and the tests passed after I ran `python setup.py test`).


alot/ui.py
```diff
@@ -157,6 +157,8 @@ def fire(_, cmdline):
                     return [cmdline[5:]]
 
             key = keys[0]
+            if len(key) > 1 and key[0].startswith('mouse'):
```
dcbaker: you can simplify this:
`if key and key[0].starswith('mouse'):`

alot/ui.py
```diff
@@ -157,6 +157,8 @@ def fire(_, cmdline):
...
+                key = key[0]+' %i' % key[1]
```
dcbaker: The formatting on this isn't quite right, we use spaces between operators.
so: `key = key[0] + ' %i' % key[1]`. Although I think that you can simplify this to `' '.join(key[:2])`

dcbaker:
The changes look good, thanks for making them. I'd like to see the fixes squashed into the original patch, and then I'll be happy to merge it.

samdmarshall:
How exactly would you like me to do that (squashing commits is not something do myself)?

pazz:
Quoting Samantha Marshall (2017-01-09 20:25:00)
>    How exactly would you like me to do that (squashing commits is not
>    something do myself)?

You can use `git rebase -i` to do that, and force push to your PR branch
(it will update automatically here then).

I haven't tested this so just one question to make sure:
does this patch prevent you from selecting text from the terminal
screen? I vaguely remember that this was the reasoning behing removing
mouse support initially..

dcbaker:
> I haven't tested this so just one question to make sure: does this patch prevent you from selecting text from the terminal screen? I vaguely remember that this was the reasoning behing removing mouse support initially..

I didn't test that. I will before we merge.

dcbaker:
> I didn't test that. I will before we merge.

Okay, I've tested and you can in fact select text with this branch.

samdmarshall:
Ok, I think this is ready to go.

dcbaker:
Looks good, thanks!

pazz:
@dcbaker which terminal are you using? I can now confirm that this in fact disables on-screen selection for me (urxvt). So unless we can separate mouse events 4/5 from the rest I personally would use the
setting proposed in #957 to disable this.

lucc:
I can confirm that mouse scrolling works but selecting text **not** (needs shift pressed). In xterm, urxvt & konsole (all with and without tmux).

pazz:
Quoting Lucas Hoffmann (2017-01-11 10:33:17)
>    I can confirm that mouse scrolling works but selecting text not (needs
>    shift pressed). In xterm, urxvt & konsole (all with and without tmux).

Uh! great find! I didn't know that it works with the shit key.
I could actually live with that..
Nevertheless, this should be configurable.

lucc:
The shift key thing is a standard in all (?) terminal applications. Either you can select text and they do not support mouse events or you can not select text and they do support mouse events.

I am living with this with at least these programs: vim/neovim, tmux, tig, ncmpcpp

PR #957, opened 2017-01-10
Settings for enabling mouse

fnurl:
Enabling mouse support disables the possibility to select text using the mouse in the terminal window. This commit adds a setting, `handle_mouse`, that can be used to turn off mouse support which restores this behaviour.

Setting `handle_mouse = False` in the config file disables mouse support in `alot`. The default value is `True` as not to conflict with the previous commit that enabled mouse support.

*Updated*: Making this a PR for the mouse setting only. Also, holding `<shift>` seems to be a workaround for selecting text in a terminal that is running a program tracking the mouse.

### DIFF
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -96,7 +96,7 @@ class UI(object):
 
         # set up main loop
         self.mainloop = urwid.MainLoop(self.root_widget,
-                                       handle_mouse=True,
+                                       handle_mouse=settings.get('handle_mouse'),
                                        event_loop=urwid.TwistedEventLoop(),
                                        unhandled_input=self._unhandeled_input,
                                        input_filter=self._input_filter)
--- a/alot/defaults/alot.rc.spec
+++ b/alot/defaults/alot.rc.spec
@@ -35,6 +35,9 @@ themes_dir = string(default=None)
 # name of the theme to use
 theme = string(default=None)
 
+# enable mouse support - mouse tracking will be handled by urwid
+handle_mouse = boolean(default=True)
+
 # headers that get displayed by default
 displayed_headers = force_list(default=list(From,To,Cc,Bcc,Subject))
 
--- a/docs/source/configuration/alotrc_table
+++ b/docs/source/configuration/alotrc_table
@@ -289,6 +289,16 @@
     :default: "Fwd: "
 
 
+.. _handle-mouse:
+
+.. describe:: handle_mouse
+
+     enable mouse support - mouse tracking will be handled by urwid
+
+    :type: boolean
+    :default: True
+
+
 .. _history-size:
 
 .. describe:: history_size


fnurl:
@pazz I'm not sure what the failure was, to me it looks like the docs were generated nicely? What should I change?

dcbaker:
These are two separate changes that do two completely different things, please split them into two commits, one for the mouse change, and one for the completion_match_anywhere change.

Also, is there any reason that we can't tell urwid to simply only capture mouse 4 and 5 (the scroll wheel) and ignore any other mouse input? I think it's better if possible to not have two paths if we can have just one.

Also the docs thing is an auto-generated file, you just need to copy your changes to the docs.

fnurl:
@dcbaker ok, thank. I agree that it would be better if urwid could ignore all events except for mouse 4 and 5. Will see if I can look into that.

Will also split the request and re-request. Closing this for now then.

alot/ui.py
```diff
@@ -71,6 +71,9 @@ def __init__(self, dbman, initialcmdline):
         # alarm handle for callback that clears input queue (to cancel alarm)
         self._alarm = None
 
+        # does the user want to use a mouse? Default to True
+        self._enable_mouse = settings.get('enable_mouse')
```
pazz: why not move this directly into line 102?

dcbaker:
I probably should have been more clear with what I wanted you to do, because its rather simple actually (espcially since Pazz likes the mouse changes and wants to merge them). We can look into ignoring mouse events that alot doesn't care about later

This should get you what I'd like, and then you can force push the branch:
```sh
git reset HEAD~
git add -p  # at this point you will be prompted to add hunks for the first commit (which can be whichver you like) you may need to use 's' to split the hunks if the changes are together
git commit
git commit -a
```

pazz:
regarding the failed test: I believe it checks that you checked in the generated docs in addition to the inline docs when you introduced the new config settings.

This can be fixed by doing a `make html` inside the docs directory and also adding the generated files
to your commit.

I cannot remember why we decided to have these files in the repository though. If we really do, then this should be better documented in the docs, under "contributing" or so, in order to prevent further confusion.

dcbaker:
@pazz, we should figure out why we decided including generated docs in git, I have found that it always ends up being more pain than it's worth to check them in and both @lucc and I have run into this same issue.

pazz:
It could have been due to restrictions on running arbitrary code on
readthedocs. I'll test a branch without those files and see if it runs
smothly..

fnurl:
@pazz regarding the hardcoded ones, I intentionally didn't touch those since I thought they were supposed to override the value set for performance reasons :)

@dcbaker Ok, thanks for the instructions. Haven't contributed to projects really, mostly forked and modified for my own personal use. I'll make them now.

PR #964, opened 2017-01-12
alot/completion: revert premature optimization

### LINKED ISSUE #962: envelope: set command removed by auto-complete
Steps to reproduce:
Start new mail, save, press `c` to add Cc, start typing an email, press `tab`. The name and email will be completed, but the `Cc` will be removed.

dcbaker:
The change this is partially reverting removed a variable assignment
that was actually needed.

Fixes #962

### DIFF
--- a/alot/completion.py
+++ b/alot/completion.py
@@ -417,7 +417,8 @@ class CommandCompleter(Completer):
                         def f((completed, pos)):
                             return ('%s %s' % (header, completed),
                                     pos + len(header) + 1)
-                        logging.debug(f(r) for r in res)
+                        res = [f(r) for r in res]
+                        logging.debug(res)
 
                 elif self.mode == 'envelope' and cmd == 'unset':
                     plist = params.split(' ', 1)


pazz:
just tested this and there is another (related?) problem:

do what you did, hit `c` to get a prompt for Cc but then hit escape to cancel.
I get the following printed on screen:
```
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 393, in callback
    self._startRunCallbacks(result)
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 501, in _startRunCallbacks
    self._runCallbacks()
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 587, in _runCallbacks
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1241, in gotResult
    _inlineCallbacks(r, g, deferred)
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1185, in _inlineCallbacks
    result = g.send(result)
  File "/home/pazz/.local/lib/python2.7/site-packages/alot-0.4.dev0-py2.7.egg/alot/commands/globals.py", line 147, in apply
    raise CommandCanceled()

```

pazz:
this happens on master as well btw.

dcbaker:
I looked into it, and I can replicate that all the way back to 3.6 (the compose option changed to it's current semantics in 3.6), so while it's definately related I don't think it's the same bug or introduced at the same time. If you can verify that my findings are correct we should open a different bug for that.

pazz:
thanks, please do that. I'll merge this here

PR #966, opened 2017-01-12
alot/ui: Handle Exceptions in UI.apply_command callbacks

### LINKED ISSUE #965: Unhandled CommandCancled in defered
pazz noted in #964 that there is another bug in the `set Cc` command. if `set Cc` is canceled, then the `CommandCanceled` will be unhandled.

dcbaker:
This ports the same handling used in UI.apply_commandline to
UI.apply_command. This means that raising a CommandCanceled exception
trhough UI.apply_command will end up with a notification instead of a
stacktrace.

Fixes #965

### DIFF
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -112,6 +112,19 @@ class UI(object):
         # start urwids mainloop
         self.mainloop.run()
 
+    def _error_handler(self, failure):
+        """Default handler for exceptions in callbacks."""
+        if failure.check(CommandParseError):
+            self.notify(failure.getErrorMessage(), priority='error')
+        elif failure.check(CommandCanceled):
+            self.notify("operation cancelled", priority='error')
+        else:
+            logging.error(failure.getTraceback())
+            errmsg = failure.getErrorMessage()
+            if errmsg:
+                msg = "{}\n(check the log for details)".format(errmsg)
+                self.notify(msg, priority='error')
+
     def _input_filter(self, keys, raw):
         """
         handles keypresses.
@@ -229,19 +242,8 @@ class UI(object):
         for cmdstring in split_commandline(cmdline):
             d.addCallback(apply_this_command, cmdstring)
 
-        # add sequence-wide error handler
-        def errorHandler(failure):
-            if failure.check(CommandParseError):
-                self.notify(failure.getErrorMessage(), priority='error')
-            elif failure.check(CommandCanceled):
-                self.notify("operation cancelled", priority='error')
-            else:
-                logging.error(failure.getTraceback())
-                errmsg = failure.getErrorMessage()
-                if errmsg:
-                    msg = "%s\n(check the log for details)"
-                    self.notify(msg % errmsg, priority='error')
-        d.addErrback(errorHandler)
+        d.addErrback(self._error_handler)
+
         return d
 
     @staticmethod
@@ -666,7 +668,7 @@ class UI(object):
             prehook = cmd.prehook or (lambda **kwargs: None)
             d = defer.maybeDeferred(prehook, ui=self, dbm=self.dbman, cmd=cmd)
             d.addCallback(call_apply)
-            d.addCallback(call_posthook)
+            d.addCallbacks(call_posthook, self._error_handler)
             return d
 
     def handle_signal(self, signum, frame):


dcbaker:
I'm a little bit fuzzy on how Twisted's `defered`'s work with Errback. Since `UI.apply_commandline` is just a wrapper around `UI.apply_command`, do we need the handler attached to both defered's, or only the one in `UI.apply_command`? This definitely fixes the issue and makes things work, but I'm wondering if it's 100% correct.

pazz:
I'm not 100% sure this is correct as well:
IIRC, if a deferred has an errback, this will be used as exception handler in case an exception is raised,
and unless it re-raises an exception it will be considered dealt with.
So in particular, if you have (as we have) a chain of callbacks for the trivial deferred constructed in apply_commandline, and an error happens in a callback in the middle, it should be handeled by the errback of the outermost deferred because that will prevent the other callbacks (which in our case fire the following commands in the sequence).
However, if the inner callbacks (constructed within `UI.apply_command`) have their own errbacks, then the global sequence would not be interrupted when some command fails in the middle.
So I'd expect this unwanted behaviour in your version.
To test, raise some exception in the apply method of some commandX and some log message in some commandY. Then call `commandX;commandY` from the alot command prompt.
You should see X's error message but not Y's log message.

dcbaker:
Looking at it, `apply_commandline` creates a defered that calls `apply_command`, which in turn creates a defered, so the call chain looks like this (I think) apply_commandline -> defered -> apply_command -> defered, and currently in master the only error handler is in `apply_commandline`, so if an exception is raised in the defered created by `apply_command` then it wont be handled, which is what's happening currently.

Based on that reasoning the right solution is to move the handler into `apply_command` and out of `apply_commandline`, correct?

dcbaker:
Okay, after reading more of the twisted documentation, a bit of hacking, and a lot of really painful reasoning about how asynchronous programming works (or at least, how I think it works) I think what happens is this:

This assumes that apply_command returns a `Deferred`, not `None`:
```
apply_commandline -> (cb: apply_command, eb: errorHandler) -> apply_command -> 
((cb: call_apply), (cb: call_posthook))
```

Now, callbacks and errbacks are pairs, that is each after each callback or errback, if it was successful (no error was raised and the return value is not a Failure) then the next callback will be called, otherwise the next errback will be called. This only applies to callbacks and errbacks registered via `Deferred.addCallback` or `Deferred.addErrback`, not `Deferred`s returning additional `Deferred`s.

Therefore, I think we actually need to pass the error handler to at last the `call_apply` callback (as this patch does) since it's a completely separate callback chain that can also raise a CommandCanceled exception.

Does that seem right?

pazz:
Quoting Dylan Baker (2017-01-13 23:28:31)
>    Does that seem right?

Sorry, it's been a while since I've read about those things. If you just
have then you are the expert :)
I cannot point to any "obvious" flaws in your argument though.

As long as this patch does not mess with the fact that command sequences
get interrupted as soon as one command fails, I'm happy to push the
change.

dcbaker:
So I did some more thinking about it, and while this is "correct" I think I have a better idea of how to fix it. I'll push an update.

dcbaker:
Okay, I can't get my better idea (which was to refactor to get a single deferred rather than a deferred that returns deferreds) working. So if no one objects I think I would like to say this seems correct and fixes the issue, so, can we merge this?

pazz:
can you just confirm that command sequences are correctly interrupted as mentioned above?
This will be crucial for instance if you fire `send; bclose` in an envelope buffer: If the send fails, the buffer must not be closed.

dcbaker:
Well, that shows yet another bug (actually, I found two). 'send; bclose' works both with and without the patch, but it will (on both master and the feature branch) send the email, then prompt to send without closing.

The other bug I found is that I typo'd 'bclose' as 'blcose', and it didn't stop until it hit that command. We probably want to at least verify that each command exists before firing them.

pazz:
Quoting Dylan Baker (2017-01-18 17:49:28)
>    Well, that shows yet another bug (actually, I found two). 'send; bclose'
>    works both with and without the patch, but it will (on both master and the
>    feature branch) send the email, then prompt to send without closing.

Yikes!

Actually, I was referring to this sequence where send fails.

> 
>    The other bug I found is that I typo'd 'bclose' as 'blcose', and it didn't
>    stop until it hit that command. We probably want to at least verify that
>    each command exists before firing them.

Yes. I think this should be checked by `ui.commandline` before adding the
callback. But we should only instanciate the Command (interpret the
commandstring) in the callback, i.e., after all previous commands
were executed!

dcbaker:
err, that should be 'close without sending', not 'send without closing'. I haven't had enough coffee yet apparently.

dcbaker:
oops, got too zealous deleting branches...

pazz:
Sorry, are you still working on this?

dcbaker:
Yes. Where are we at on this, did you have more changes you wanted me to make?

pazz:
Come to think of it: that send command actually "suceeds" in the above case. It tries to send, fails, sends a notification and crucially, does not raise an exception.
This should be independent of this patch, which looks good to me.

dcbaker:
Cool! 0.5 release happening soon hopefully.

PR #967, opened 2017-01-13
Add hooks to all created commands?

fnurl:
Here something I made in order to be able to add pre and post hooks to all commands that are run, not only those initiated from the command line. I don't know if this restriction is by design or if it is a bug. Below follows a description of my line of thoughts. There might be a better way to do this, but this is what I could think of with my limited experience with the code base.

Please do comment and tell me what you think.

**Description**
Even though commands have pre and post hooks, this has only been true for commands initiated from the command line.

This commit changes where the hooks are added. Rather than being added in `commands.commandfactory()`, they are now added in `commands.Command.__init__()`. This because not all commands are created using `commands.commandfactory()`.

In order to implement this, a top level dictionary, `REVERSE_COMMANDS`, has been added using command classes as keys and a tuple containing `cmdname` and `mode` as value. Reverse lookup is done using `reverse_lookup_command(klass)` which takes a command class and returns the mode and name of the command.

**Use-case**
The use-case that motivated this change was that I wanted to create a hook that added `format="flowed"` to the `Content-type` header when replying and composing messages. However, it was not possible to do this, since only the  `reply` and `compose` command's hooks were triggerered which do not have access to the message itself.

The chain of commands being queued and applied include `commands.envelope.EditCommand`, but it the command is created from `commands.globals.ComposeCommand`, which in turn is called from the originating `commands.thread.ReplyCommand`.

Neither the `globals.ComposeCommand` nor the `envelope.EditCommand` have any pre or post hook attached to them since they were not created using `commandfactory` which is made for creating commands from the command line.

### DIFF
--- a/alot/commands/__init__.py
+++ b/alot/commands/__init__.py
@@ -10,19 +10,52 @@ import os
 import re
 
 from ..settings import settings
-from ..helper import split_commandstring, string_decode
+from ..helper import split_commandstring, string_decode, classproperty
 
 
 class Command(object):
 
     """base class for commands"""
     repeatable = False
+    mode = ""
+    cmdname = ""
+
+    @classproperty
+    def id(cls):
+        """Mode and command name of instance as string."""
+        return "{}.{}".format(cls.mode, cls.cmdname)
 
     def __init__(self):
-        self.prehook = None
-        self.posthook = None
+        # set to False if no attempt to fetch hooks from settings. After
+        # looking for hooks, they will contain a callable or None if no
+        # hook was found.
+        self._prehook = False
+        self._posthook = False
+
         self.undoable = False
         self.help = self.__doc__
+
+    @property
+    def prehook(self):
+        """Command pre hook."""
+        if self._prehook is False:
+            get_hook = settings.get_hook
+            mode = type(self).mode
+            cmdname = type(self).cmdname
+            self._prehook = (get_hook('pre_%s_%s' % (mode, cmdname)) or
+                             get_hook('pre_global_%s' % cmdname))
+        return self._prehook
+
+    @property
+    def posthook(self):
+        """Command post hook."""
+        if self._posthook is False:
+            get_hook = settings.get_hook
+            mode = type(self).mode
+            cmdname = type(self).cmdname
+            self._posthook = (get_hook('post_%s_%s' % (mode, cmdname)) or
+                              get_hook('post_global_%s' % cmdname))
+        return self._posthook
 
     def apply(self, caller):
         """code that gets executed when this command is applied"""
@@ -33,6 +66,7 @@ class CommandCanceled(Exception):
     """ Exception triggered when an interactive command has been cancelled
     """
     pass
+
 
 COMMANDS = {
     'search': {},
@@ -134,15 +168,17 @@ class registerCommand(object):
         self.forced = forced or {}
         self.arguments = arguments or []
 
-    def __call__(self, klass):
-        helpstring = self.help or klass.__doc__
+    def __call__(self, class_):
+        helpstring = self.help or class_.__doc__
         argparser = CommandArgumentParser(description=helpstring,
                                           usage=self.usage,
                                           prog=self.name, add_help=False)
         for args, kwargs in self.arguments:
             argparser.add_argument(*args, **kwargs)
-        COMMANDS[self.mode][self.name] = (klass, argparser, self.forced)
-        return klass
+        COMMANDS[self.mode][self.name] = (class_, argparser, self.forced)
+        class_.mode = self.mode
+        class_.cmdname = self.name
+        return class_
 
 
 def commandfactory(cmdline, mode='global'):
@@ -188,13 +224,6 @@ def commandfactory(cmdline, mode='global'):
 
     # create Command
     cmd = cmdclass(**parms)
-
-    # set pre and post command hooks
-    get_hook = settings.get_hook
-    cmd.prehook = get_hook('pre_%s_%s' % (mode, cmdname)) or \
-        get_hook('pre_global_%s' % cmdname)
-    cmd.posthook = get_hook('post_%s_%s' % (mode, cmdname)) or \
-        get_hook('post_global_%s' % cmdname)
 
     return cmd
 
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -631,3 +631,29 @@ def email_as_string(mail):
                            as_string, flags=re.MULTILINE)
 
     return as_string
+
+
+class classproperty(object):
+    """Decorator to emulate 'class properties' using class level methods.
+
+    The decorator is applied to a method to make it a class property. E.g.
+
+    .. code-block
+
+        class Foo(object):
+
+            _some_property = 0.5
+
+            @classproperty
+            def some_property(cls):
+                return cls._some_proprerty
+    """
+
+    def __init__(self, func):
+        """Store method to be used to return property value."""
+        self.__func__ = func
+
+    def __get__(self, instance, cls):
+        """Return value of property."""
+        return self.__func__(cls)
+
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -659,7 +659,7 @@ class UI(object):
             def call_posthook(_):
                 """Callback function that will invoke the post-hook."""
                 if cmd.posthook:
-                    logging.info('calling post-hook')
+                    logging.info('calling post hook for "%s"', cmd.id)
                     return defer.maybeDeferred(cmd.posthook,
                                                ui=self,
                                                dbm=self.dbman,
@@ -669,7 +669,11 @@ class UI(object):
             def call_apply(_):
                 return defer.maybeDeferred(cmd.apply, self)
 
-            prehook = cmd.prehook or (lambda **kwargs: None)
+            if cmd.prehook:
+                logging.info('calling pre hook for "%s"', cmd.id)
+                prehook = cmd.prehook
+            else:
+                prehook = lambda **kwargs: None
             d = defer.maybeDeferred(prehook, ui=self, dbm=self.dbman, cmd=cmd)
             d.addCallback(call_apply)
             d.addCallbacks(call_posthook, self._error_handler)


alot/commands/__init__.py
```diff
@@ -22,6 +22,20 @@ def __init__(self):
         self.undoable = False
         self.help = self.__doc__
 
+        mode, cmdname = reverse_lookup_command(self.__class__)
```
dcbaker: please use `type(self)` instead of accessing the dunder directly. it would probably be best to just store the result for both class and the name. This aplies to the rest of the patch.

alot/commands/__init__.py
```diff
@@ -22,6 +22,20 @@ def __init__(self):
...
+        logging.debug("fetching commands for: %s, %s, %s",
+                self.__class__.__name__, mode, cmdname)
+
+        # fetch and set pre and post command hooks
+        get_hook = settings.get_hook
+        self.prehook = get_hook('pre_%s_%s' % (mode, cmdname)) or \
```
dcbaker: Instead of using `\\` to escape the newline, wrap the right side in parens so it can be lined up:
```python
self.prehook = (get_hook('pre_%s_%s' % (mode, cmdname)) or
                get_hook('pre_global_%s' % cmdname))
```
This applies to posthook too.

alot/commands/__init__.py
```diff
@@ -41,6 +55,23 @@ class CommandCanceled(Exception):
...
+    returns mode and commandname for a commandclass
+
+    :param klass: the class used to look up the cmdname and mode
+    :type klass: class
+    :rtype: (str, str)
+    """
+    if klass in REVERSE_COMMANDS:
+        return REVERSE_COMMANDS[klass]
+    else:
+        return (None, None)
```
dcbaker: please use `class_` instead of `klass` for the name, this is in keeping with PEP8 standards:
```
* single_trailing_underscore_ : used by convention to avoid conflicts with Python keyword, e.g.

Tkinter.Toplevel(master, class_='ClassName')
```

You can also simplify this function by using `dict.get()` instead:
```python
return REVERSE_COMMANDS.get(class_, (None, None))
```

alot/commands/__init__.py
```diff
@@ -41,6 +55,23 @@ class CommandCanceled(Exception):
...
+REVERSE_COMMANDS = {}
+
+
+def reverse_lookup_command(klass):
+    """
+    returns mode and commandname for a commandclass
+
+    :param klass: the class used to look up the cmdname and mode
+    :type klass: class
+    :rtype: (str, str)
```
dcbaker: This is not true, The tuple will be `(None, None)` in the case that there is no defined hook.

alot/commands/__init__.py
fnurl: Ok, thanks, I'll do that.

alot/commands/__init__.py
fnurl: Will change this.

fnurl:
Should I add a new commit or reset, change and commit to keep the number of commits in this request to a minimum? Will fix the commit message as well.

dcbaker:
Generally when you make changes for review requests you squash the changes into the original commit. There are two ways to do that. If you're squashing into the commit that is currently the tip of your development branch you use `git commit --amend` (or git `commit -a --amend`), that's what you'll probably want to do in this case.

For longer more complicated series you use `git commit --fixup` and `git commit --squash` which will create commits called 'fixup! <my commit message> ' or 'squash! <my commit message>' respectively. Then you can use `git rebase --autosquash` to merge fixes into the original patches. Then all you do is force push and we'll look at the changes.

The only time you need to do the 'reset, change, commit' dance is when you want to take one patch and split it into two or more patches.

alot/commands/__init__.py
fnurl: > it would probably be best to just store the result for both class and the name

Do you mean that it would be better to leave the `mode` out? That would make the lookup ambiguous since the same command name can exist in different modes. Or did I misunderstand?

fnurl: ... ah, or do you mean store them locally to make the code cleaner? I'll do that.

dcbaker: the later, so we can avoid calling `type` over and over again.

fnurl:
Sorry for the multiple pushes, kept seeing minor details. I'll fix my workflow so I do better diff-checks locally in the future.

Btw, I noticed a slight inconsistency in the order or `mode` and `cmdname` in which is more "important": hooks use `mode_cmdname` but in ui/__init__.py we have `lookup_command(cmdname, mode)`.

One of the changes among my pushes was that I decided to go with the order used in ui/__init__.py

pazz:
The only thing that comes to mind about these patches is that the naming "REVERSE_COMMANDS" etc is not great: We often get people asking about making the effects of commands reversible in order to implement some kind of "undo" of tagging etc. I don't want people getting confused about your changes being in any way related tho this stuff..

fnurl:
Yes, I agree. It was named `REVERSE_COMMANDS_LOOKUP` or something similar previously, but then I renamed it to match the `COMMANDS` dictionary. But it might be better not to have the word "reverse" at all then?

Perhaps `COMMAND_NAME_LOOKUP`? 

Another alternative would be to have the whole dictionary stored under e.g. `COMMANDS['reverse']` or `COMMANDS['cmdclasses']  but that wouldn't really be a mode..

dcbaker:
I'm marking this approved, but please resolve the comments from @pazz before merging.

alot/commands/__init__.py
```diff
@@ -17,11 +17,24 @@ class Command(object):
...
 
     def __init__(self):
-        self.prehook = None
-        self.posthook = None
         self.undoable = False
         self.help = self.__doc__
 
+        class_ = type(self)
+        class_name = class_.__name__
+        cmdname, mode = reverse_lookup_command(class_)
```
pazz: it seems that you only need the cmd name and mode string to look up the hook here. Why not turn `reverse_lookup_command` into a function `get_hook` that takes `class_` and returns the callable hook or None?

alot/commands/__init__.py
```diff
@@ -136,15 +163,16 @@ def __init__(self, mode, name, help=None, usage=None,
...
+        helpstring = self.help or class_.__doc__
         argparser = CommandArgumentParser(description=helpstring,
                                           usage=self.usage,
                                           prog=self.name, add_help=False)
         for args, kwargs in self.arguments:
             argparser.add_argument(*args, **kwargs)
-        COMMANDS[self.mode][self.name] = (klass, argparser, self.forced)
-        return klass
+        COMMANDS[self.mode][self.name] = (class_, argparser, self.forced)
+        REVERSE_COMMANDS[class_] = (self.name, self.mode)
```
pazz: How about turning `REVERSE_COMMANDS` into a dict `HOOKS`,
that maps `class_` to a pair `(prehook,posthook)`, and which gets populated here?
The `get_hook` function I suggested above is then just `HOOKS[class_]`.

fnurl: Yeah, that sounds cleaner. I'll modify the patch so it does that instead!

fnurl: What about having `prehook` and `posthook` as class variables instead that get set when a class is registered as a command?

``` python
class Command(object):
    prehook = None
    posthook = None
```

and then in the registerCommand decorator

``` python
class_.prehook = prehook
class_.posthook = posthook
```

then no function at all is needed, wouldn’t that be more pythonic? or am I off?

fnurl: It looks like `mode` and `cmdname` are needed one way or another.

Hooks are accessed via `settings.get_hook(key)` where `key` is the name of the hook function. Since names of hook functions are constructed as `{pre|post}_<mode>_<cmdname>`, both `mode` and`cmdname` are needed to retrieve the hook from `settings.get_hook()`. This is why I stored `mode` and `cmdname` in the original global dictionary.

The only place where `mode` and `cmdname` are available without "storing" them first, is in the `registerCommand` decorator. However, the decorator is run on the classes *before* the settings have been loaded, so there are never any hooks to retrieve.

I pushed another solution where I store `mode` and `cmdname` as class variables, then making `prehook` and `posthook` properties of `Command` instead with only a getter.

So with this solution, no extra "lookup" structure is needed, and hooks are fetched when needed.

What do you think?

alot/commands/__init__.py
```diff
@@ -15,25 +15,44 @@ class Command(object):
...
-        # fetch and set pre and post command hooks
-        # they are set to None if not defined in the hooks file
-        get_hook = settings.get_hook
-        self.prehook = (get_hook('pre_%s_%s' % (mode, cmdname)) or
-                        get_hook('pre_global_%s' % cmdname))
-        self.posthook = (get_hook('post_%s_%s' % (mode, cmdname)) or
-                         get_hook('post_global_%s' % cmdname))
+        # set to False if no attemtp to fetch hooks from settings
+        self._prehook = False 
+        self._posthook = False
```
dcbaker: You should use `None` here instead of `False`.

alot/commands/__init__.py
```diff
@@ -15,25 +15,44 @@ class Command(object):
...
 
-        logging.debug("%s prehook: %s", class_name, str(self.prehook))
-        logging.debug("%s posthook: %s", class_name, str(self.posthook))
+    @property
+    def id(self):
+        """Mode and command name of instance as string."""
+        class_ = type(self)
+        return class_.mode + '.' + class_.cmdname
```
dcbaker: Gah, python3 can have `property` and `classmethod` used together, which would be nice. You could implement a `classproperty` in the helpers module that would make this nicer. Something like:

```python
class classproperty(object):
    """A decorator like `property` but works on classes."""
    def __init__(self, func):
        self.__func = func

    def __get__(self, instance, cls):
        return self.__func(cls)
```

Then you could implement this method as:
```python
@classproperty
def id(cls):
    return '{}.{}'.format(cls.mode, cls.cmdname)
```
However, this could be done as a followup instead of in this patch.

alot/commands/__init__.py
```diff
@@ -15,25 +15,44 @@ class Command(object):
...
+
+    @property
+    def prehook(self):
+        """Command pre hook."""
+        if self._prehook == False:
```
dcbaker: Two things:
1) Don't use `==` with singletons (`None`, `True`, `False`, and `NotImplemented` are the most common), use `is` (and use `is not` instead of `!=`), It's faster
2) You can just write this as `if not self._prehook`

alot/commands/__init__.py
fnurl: I had them `False` from the start since they might be set to `None` there is no hook for the particular command. That way, only a single call to `settings.get_hook()` has to be made, not that there would be very many though.

alot/commands/__init__.py
fnurl: The reason for using `==` here was to distinguish between `None` and `False`. If set to `False`, `settings.get_hook()` has not been called. If `self._prehook` is `None`, `settings.get_hook()` has been called but no hook was found, so if it is `None`, that is actually what is supposed to be returned.

The property will not be read that many times, but the idea was to avoid the extra calls to `settings.get_hook()`. What do you think?

alot/commands/__init__.py
```diff
@@ -15,25 +15,44 @@ class Command(object):
...
-        cmdname, mode = reverse_lookup_command(class_)
-
-        # fetch and set pre and post command hooks
-        # they are set to None if not defined in the hooks file
-        get_hook = settings.get_hook
-        self.prehook = (get_hook('pre_%s_%s' % (mode, cmdname)) or
-                        get_hook('pre_global_%s' % cmdname))
-        self.posthook = (get_hook('post_%s_%s' % (mode, cmdname)) or
-                         get_hook('post_global_%s' % cmdname))
+        # set to False if no attemtp to fetch hooks from settings
```
dcbaker: s/attemtp/attempt :)

alot/commands/__init__.py
dcbaker: Okay, I see why you're using False. I guess that's okay, but please do use `is` instead of `==`

alot/commands/__init__.py
dcbaker: Okay, I see why you used False. Let's leave it then.

alot/helper.py
```diff
@@ -631,3 +631,29 @@ def email_as_string(mail):
...
 
     return as_string
+
+
+class classproperty(object):
+    """Decorator to emulate 'class properties' using class level methods.
+    
+    The decorator is applied to a method to make it a class property. E.g.
+    
+    .. code-block:: python
```
fnurl: I using 

    ``` python
    stuff
    ```

in the docstring triggered an error so I am using `.. code-block:: python` instead. Was I correct to do so?

fnurl:
I'm not sure if I am generating the travis error or there is something else responsible? I fixed a docstring issue that popped up (switched to using `.. code-block:: python` instead of triple backticks.

I tried doing a `make html` in the docs dir, but nothing new was generated or updated.

alot/commands/__init__.py
```diff
@@ -10,20 +10,53 @@
...
+    @property
+    def prehook(self):
+        """Command pre hook."""
+        if self._prehook is False:
+            get_hook = settings.get_hook
+            mode = type(self).mode
+            cmdname = type(self).cmdname
+            self._prehook = (get_hook('pre_%s_%s' % (mode, cmdname)) or
+                             get_hook('pre_global_%s' % cmdname))
+        return self._prehook
```
dcbaker: Is there a reason that the pre and post hooks should be applied to the instance rather than the class? I think that we actually want to attach these to the class.

alot/helper.py
```diff
@@ -631,3 +631,29 @@ def email_as_string(mail):
...
 
     return as_string
+
+
+class classproperty(object):
+    """Decorator to emulate 'class properties' using class level methods.
+
+    The decorator is applied to a method to make it a class property. E.g.
+
+    .. code-block:: python
```
dcbaker: you don't need the `python` bit, it's the default domain in sphinx.

dcbaker:
Also, don't worry about the docs failing at the moment, thats a bigger issue not related to your changes, see #994.

lucc:
You can force the docs to rebuild with `make cleanall html`.

alot/commands/__init__.py
fnurl: Yes, I agree that it would be better if hooks were class level function objects, but I couldn’t get it to work previously since the hooks kept getting called as bound methods.

However, I now have a working patch using a wrapper class - `Hook` which wrap the hook functions. Should I commit it as an amend to this PR?

alot/helper.py
fnurl: ok!

alot/commands/__init__.py
dcbaker: Why can't you use the classproperty object here? Since they return a function it should be invisible to a user that they're calling a property and not a function.

fnurl: @dcbaker I tried that first, but the function is referenced from a class attribute (I had them `_prehook` and `_posthook` on the same level as `mode` and `cmdname`).

When the hooks were called, Python treated them as if the hook was a function defined on that level, i.e. expected them to receive `self` as their first argument which is not how the hook functions are defined.

Here is my working solution:

- Hook class: https://github.com/fnurl/alot/blob/hook-always-class-store/alot/settings/manager.py#L467
- hooks as `classproperties`: https://github.com/fnurl/alot/blob/hook-always-class-store/alot/commands/__init__.py#L26
- wrapping the hooks: https://github.com/fnurl/alot/blob/hook-always-class-store/alot/settings/manager.py#L311

dcbaker: hmmmmmmm. I think we can handle this as a follow up, I'm not convinced that you're Hook approach is right, subclassing `__new__` is almost always the wrong thing to do, there are so many corners involved. So lets leave this as-is for now and we can handle it in the next go around.

dcbaker:
Can you rebase this on master so that the CI will pass (assuming there aren't docs changes that need to be checked in). Assuming that all goes right I'll go ahead and approve the merge.

If you could add one additional patch, this is worth putting in the NEWS file for the next release :)

@pazz, anything else you need to see change before this lands?

alot/commands/__init__.py
fnurl: Will do! However, overriding `__new__` isn't really necessary. I removed it from the current commit: https://github.com/fnurl/alot/blob/hook-always-class-store/alot/settings/manager.py#L472

fnurl:
Done! Looks like the CI passed too.

@dcbaker Not sure what you mean by adding an additional patch though? Are you referring to moving the hooks from instances to classes in the followup? Very happy to be able to contribute to software I use!

dcbaker:
Just add something to the NEWS file that the next release has hooks for all commands.

alot/commands/__init__.py
```diff
@@ -34,6 +67,7 @@ class CommandCanceled(Exception):
     """
     pass
 
+
```
pazz: why the extra newline?

pazz:
Ah! not so fast! Sorry but I've only now realised that you store the command name inside the Comman object. These are properties of the instances, not the class right? So this stores the command name that was used to create this particular instance?
Because otherwise, this could introduce trouble for Commands that cover several command names, like
`thread.ChangeDisplaymodeCommand`.

+1 for the use of `@property`s !

alot/commands/__init__.py
fnurl: That's how I usually follow PEP8 on the top level. Two blank lines between "stuff". In this case, the end of the `CommandCanceled` class and the top level `COMMANDS` dictionary initialization. There already was two blank lines after the `COMMANDS` dictionary initialization before `def lookup_command(cmdname, mode)`, so I added an extra to the spacing above.

fnurl:
They are defined in the `Command` class, but are never stored in the `Command` class, rather they are stored in the derived classes when they are registered using the `@registerCommand` decorator. E.g. in `commands/envelope.py`:

``` python
MODE = 'envelope'


@registerCommand(
    MODE, 'attach',
    arguments=[(['path'], {'help': 'file(s) to attach (accepts wildcads)'})])
class AttachCommand(Command):
    """attach files to the mail"""
```

The decorator then saves it in the specific class that gets registered. Here is the code from `registerCommand`:

``` python
    def __call__(self, class_):
        helpstring = self.help or class_.__doc__
        argparser = CommandArgumentParser(description=helpstring,
                                          usage=self.usage,
                                          prog=self.name, add_help=False)
        for args, kwargs in self.arguments:
            argparser.add_argument(*args, **kwargs)
        COMMANDS[self.mode][self.name] = (class_, argparser, self.forced)
        class_.mode = self.mode
        class_.cmdname = self.name
        return class_
```

I checked to see if the same class could be used for multiple combinations of`mode` and `cmdname`, but it didn’t seem like it. E.g. rather than the same `TagCommand` class being used in *search* and *thread* mode, there are two separate `Command` subclasses, one in `commands/search.py` and one in `commands/thread.py`. As it stands, it is the `AttachCommand` class that gets `AttachCommand.mode` set to `'envelope'` in this case, and `AttachCommand.cmdname` set to `'attach'`.

fnurl:
Given the use of multiple decorators and the way commands are applied from other commands, there are several cases where the mode and command name cannot be established. E.g. here:

``` python
@registerCommand(MODE, 'select')
class ThreadSelectCommand(Command):

    """select focussed element. The fired action depends on the focus:
        - if message summary, this toggles visibility of the message,
        - if attachment line, this opens the attachment"""
    def apply(self, ui):
        focus = ui.get_deep_focus()
        if isinstance(focus, AttachmentWidget):
            logging.info('open attachment')
            ui.apply_command(OpenAttachmentCommand(focus.get_attachment()))
        else:
            ui.apply_command(ChangeDisplaymodeCommand(visible='toggle'))
```

I have four possible solutions I would like to propose: 

- **Alt 1**: add two keywords to the signature of the `Command` class constructor: `mode` and `cmdnamn` in order to supply these when possible/wanted. If they are not provided, the command is not "hookable".
- **Alt 2**: add two keywords to the `Command` class constructor as above, but if no `mode` and `cmdname` is provide when creating the instance, a generic `"subcommand"` mode is assumed, and the name of the command class is used as `cmdname`. In the case above, we would have `mode='subcommand'` and `cmdname='ChangeDisplaymodeCommand'`.
- **Alt 3**: a combination of *Alt 1* and *Alt 2*: Rather than using the name of the class as default `cmdname`, a generic name is defined in each class definition. This alternative hides the actual `Command` class names from the user and also allow for making commands unhookable.
- **Alt 4**: deprecate the use of multiple `registerCommand` decorators on a command. If the same class can be used in multiple contexts, wrapper classes are used and registered separately. This way, each mode+command combination has a 1-1 mapping to the class that is run and hooks can be stored on a class level.

I really like the idea of being able to hook into all commands, but the case might perhaps be that not all applied commands make sense in being able to hook into. One thing that makes the deprecation route better in one sense is that it could be implemented "gracefully". If multiple registrations are done, the command is made unhookable, but this change does not effect any other part of the code. This seems like the "simplest" solution?

pazz:
Honestly, my take on it is to leave it as it is:
"commands" were originally intended to represent "user instructions", as
given through the command prompt. All those go through commandfactory
and get the respective hooks already.
The fact that some are implemented by instanciating other `Command`
classes is besides the point, as this is implementation detail and should be invisible to the user.

Alternative 4 would make it really difficult to simply map "select"
(enter) to behave differently on thread buffers, depending on the
current focus. You'd have one command per operation, so you'd need to
map different keys to all those. Not nice.
If we *really* want this feature, then I'd opt for alternative 1.

alot/commands/__init__.py
lucc: I agree with @fnurl here.

fnurl:
Ok, I understand the design decision of only exposing commands that are instructions from the command prompt.

The original reason for me sending in this patch was that the data exposed to the pre and post hooks when running the `reply` command in `thread` mode was insufficient to modify the headers of a message (envelope is inaccessible).

I think that being able to modify the header of a message before editing it in the text editor is a reasonable hook to supply. Three special hooks are provided for manipulation other parts of the message: `reply-prefix`, `text-quote` and `reply_subject`.

A less general patch would be to add a specific hook that either exposes all headers or specific headers.

What do you think?

pazz:
there are also hooks `pre|post_edit_translate`...

fnurl:
Thanks for the `pre|post_edit_translate` tip, that gives access. I had looked it up in the documentation, but never tried it since the documentation said only the body text was provided.

I am fine not having this PR accepted based on the issues that have popped up. I have enjoyed reading the code and appreciate the effort you put into alot and the comments here. I think I will make a last commit to this PR with an implementation of alt 1 and can close the PR without any merge after that if you want to?

pazz:
Thanks for this anyway. As I said, I'd be OK with dropping this, but let's see what the others say.
And yes, the docs regarding these `pre|post_edit_translate` hooks could be improved.

Issue #969, opened 2017-01-13
__future__ imports


dcbaker:
In order to get ready for python3.x support, it would be usful to enable some of the __future__ imports in alot, there are 3 that should be fairly trivial to implement correctly: `__future__.division`, `__future__.absolute_import` and `__future__.print_statement`.

lucc:
I have a patch for \_\_future\_\_.print_**function**. I did not know there was a print_statemant in `__future__` as well. But if there is such a thing I still think it is better to use print_funtion as it will make the transition easier later.

lucc:
It is #971.

In general I am all for splitting the py3 work across multible PRs and even merge all the stuff that is also py2 compatible as soon as possible! So good idea.

I did some work on imports not so long ago in fe748a7f47fb61d46dd88c270c2329e6ff19a895 and 571422a079ce235eba738ccfeac1ca2f043ece23. Maybe that will make the absolute_import stuff as simple as adding one line to each file that imports something. (Or should we only add the line to files that import stuff from our code base or to all files, even empty ones?)

dcbaker:
Lol, yes print_function

I'm on mobile, please excuse autocorrect fail.

pazz:
Right: we don't use prints, future division is #1012 and future.absolute_imports is #984.
Can we close this issue?

lucc:
According to https://docs.python.org/2/library/__future__.html there also is the bytes literal stuff. Do we have any use for that?

(The 3 @dcbaker mentioned where only the easy ones as I read it, but it might also be all of them)

pazz:
BTW: I created a new milestone for the py3k stuff. Since I don't really know what else is necessary to port, feel free to add issues to that milestone as you see fit.

dcbaker:
I don't think that we want to use unicode_literals. It's nice for pojects that need to support python 2 and python 3 and want to have python2 behave like python 3. I'm going to close this since I think this is done.

PR #800, opened 2015-12-03
[RFC] another python 3 port

geier:
This python 3 port should make alot python 3.3+ compatible.

I will now start to use alot with python 3 and fix any issues I come across, but please do leave comments as well (or give it a go under python 3).

PR #779, opened 2015-07-18
Python3

tlevine:
I got as far as I could until getting stuck on this error, in both Python 2.7 and Python 3.4.
```
$ python2.7 -c 'import alot.init as i; i.main()'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "alot/init.py", line 12, in <module>
    from alot.db.manager import DBManager
  File "alot/db/__init__.py", line 6, in <module>
    from .thread import Thread
  File "alot/db/thread.py", line 7, in <module>
    from .message import Message
  File "alot/db/message.py", line 8, in <module>
    from notmuch import NullPointerError
  File "/Library/Python/2.7/site-packages/notmuch/__init__.py", line 54, in <module>
    from .database import Database
  File "/Library/Python/2.7/site-packages/notmuch/database.py", line 24, in <module>
    from .globals import (
  File "/Library/Python/2.7/site-packages/notmuch/globals.py", line 27, in <module>
    raise ImportError("Could not find shared 'notmuch' library.")
ImportError: Could not find shared 'notmuch' library.
```

PR #971, opened 2017-01-14
Use print function from `__future__`

lucc:
Related #969, #779, #800
label: py3k

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -142,17 +142,17 @@ def shorten_author_string(authors_string, maxlength):
         between first and next authors is added.
 
     >>> authors = u'King Kong, Mucho Muchacho, Jaime Huerta, Flash Gordon'
-    >>> print shorten_author_string(authors, 60)
+    >>> print(shorten_author_string(authors, 60))
     King Kong, Mucho Muchacho, Jaime Huerta, Flash Gordon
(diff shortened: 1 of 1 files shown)


pazz:
This looks totally harmless to me. The file `extra/tagsections_convert.py ` can actually be removed completely: it was a converter script for users of an earlier version with incompatible config format.
I think it's safe to assume that this will never be used anymore.

The other stuff is in the docstrings only, and I'm ok with making this change for consistency with further py3k ports.

regarding putting `__future__` import everywhere: OK, but I think it only makes sense in those places where we actually use those imports. I'd like to avoid imporing all kinds of unused stuff.

lucc:
So if I see that correctly we don't use print in any real code any more. Only in this one doctest.  Hence I vote that we drop this PR and the print_function import and port *all* doctest to unittests instead.

pazz:
fine with me

lucc:
@pazz and what about the file extra/theme_convert.py?

pazz:
Quoting Lucas Hoffmann (2017-01-17 12:16:56)
>    [1]@pazz and what about the file extra/theme_convert.py?

Same thing. I forgot to mention that, thx.

lucc:
I withdraw this PR in favor of #975.

PR #975, opened 2017-01-17
Porting old doctests to unittest

lucc:
There are some doctests in the code base.  This ports most of them to the current unittest based test suite.  The ones that are not covered here are in #928.  (This idea was born from #971)

I did not mock anything in these tests as I don't know if it is necessary here.  Any opinions?

Also I did not add a GPL header or other comments yet.  What should be our default?

Should this be merged after #928 as there will be conflicts?

### DIFF
--- a/tests/commands/init_test.py
+++ b/tests/commands/init_test.py
@@ -1,0 +1,28 @@
+# encoding=utf-8
+
+"""Test suite for alot.commands.__init__ module."""
+
+import argparse
+import unittest
+
+from alot import commands
+from alot.commands import thread
+
+
+class TestLookupCommand(unittest.TestCase):
+
+    def test_look_up_save_attachment_command_in_thread_mode(self):
+        cmd, parser, kwargs = commands.lookup_command('save', 'thread')
+        # TODO do some more tests with these return values
+        self.assertEqual(cmd, thread.SaveAttachmentCommand)
+        self.assertIsInstance(parser, argparse.ArgumentParser)
+        self.assertDictEqual(kwargs, {})
+
+
+class TestCommandFactory(unittest.TestCase):
+
+    def test_create_save_attachment_command_with_arguments(self):
+        cmd = commands.commandfactory('save --all /foo', mode='thread')
+        self.assertIsInstance(cmd, thread.SaveAttachmentCommand)
+        self.assertTrue(cmd.all)
+        self.assertEqual(cmd.path, u'/foo')
--- a/tests/db/envelop_test.py
+++ b/tests/db/envelop_test.py
@@ -1,0 +1,14 @@
+# encoding=utf-8
+
+import unittest
+
+from alot.db import envelope
+
+
+class TestEnvelopeMethods(unittest.TestCase):
+
+    def test_setitem_stores_text_unchanged(self):
+        "Just ensure that the value is set and unchanged"
+        e = envelope.Envelope()
+        e['Subject'] = u'sm\xf8rebr\xf8d'
+        self.assertEqual(e['Subject'], u'sm\xf8rebr\xf8d')
--- a/tests/helper_test.py
+++ b/tests/helper_test.py
@@ -1,0 +1,78 @@
+# encoding=utf-8
+
+"""Test suite for alot.helper module."""
+
+import unittest
+
+from alot import helper
+
+
+class TestHelperShortenAuthorString(unittest.TestCase):
+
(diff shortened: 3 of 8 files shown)


pazz:
very nice! Do those actually pass ? :)

lucc:
I don't get what you mean? The old doctests or the new ones? I modified them a bit in some places.

But travis likes them: https://travis-ci.org/pazz/alot/jobs/192661458

pazz:
Quoting Lucas Hoffmann (2017-01-17 12:32:42)
>    I don't get what you mean? The old doctests or the new ones? I modified
>    them a bit in some places.

I was referring to the old doctests. AFAIK, they were never really
systematically checked.
 
>    But travis likes them: [1]https://travis-ci.org/pazz/alot/jobs/192661458

A cool! I didn;t realize that they already run on travis without extra
magic.

lucc:
The "magic" was in the travis file since 8fb47b645ae58a1bed7a262a9e8e0f4063fccf15. (And already really used in 1e4229ad52fc6db4f2a3141cbe72d7f42658ca06)

pazz:
It looks good to me but I'll leave it to @dcbaker to review/merge as he had some experience with testing and some dislike for unittest, whereas [I have no qualified opinion](https://github.com/pazz/alot/pull/934#issuecomment-268881982) on the matter.
As far as I understand, the upshot of our previous discusion on unittests vs pytest etc (#917) was that
we will go with unittest for now in order to get things going. So as far as I'm concerned, this can go live.

tests/commands/init_test.py
```diff
@@ -0,0 +1,27 @@
...
+        # TODO do some more tests with these return values
+        self.assertEqual(cmd, SaveAttachmentCommand)
+
+
+class TestCommandFactory(unittest.TestCase):
+
+    def test_2(self):
+        cmd = commandfactory('save --all /foo', mode='thread')
+        self.assertEqual(
+                cmd.__repr__()[:56],
```
dcbaker: repr(cmd)[:56]

tests/db/envelop_test.py
```diff
@@ -0,0 +1,13 @@
...
+import unittest
+
+from alot.db.envelope import Envelope
+
+
+class TestEnvelopeMethods(unittest.TestCase):
+
+    def test_setitem(self):
+        e = Envelope()
+        e['Subject'] = u'sm\xf8rebr\xf8d'
```
dcbaker: it's probably worhtleavinga comment about the bizare unicode string, otherwise everyone who reads this is going to wonder if that's something speicial, or if subject can be set to anything.

tests/commands/init_test.py
```diff
@@ -10,18 +11,18 @@
...
 
 class TestCommandFactory(unittest.TestCase):
 
-    def test_2(self):
+    def test_create_save_attachment_command_with_arguments(self):
         cmd = commandfactory('save --all /foo', mode='thread')
-        self.assertEqual(
-                cmd.__repr__()[:56],
-            "<alot.commands.thread.SaveAttachmentCommand object at 0x")
+        self.assertIsInstance(cmd, SaveAttachmentCommand)
```
dcbaker: I like all of the changes in this patch, but could you squash them backinto the original commits?

tests/helper_test.py
```diff
@@ -0,0 +1,32 @@
...
+
+from alot.helper import shorten_author_string
+
+
+class TestHelperShortenAuthorString(unittest.TestCase):
+
+    authors = u'King Kong, Mucho Muchacho, Jaime Huerta, Flash Gordon'
+
+    def test_high_maxlength_keeps_string_intact(self):
+        self.assertEqual(shorten_author_string(self.authors, 60), self.authors)
```
dcbaker: I'm terrible about doing this myself, but... It's better to assign the value of the function being tested than to pass it directly into the assert, since that makes it much easier to inspect with a debuger like pdb if the test is failing.

tests/db/envelop_test.py
lucc: Do you have an idea about what to write in the comment. I just copied that from the doctest.

dcbaker: maybe something like "just ensure that the value is set and unchanged"?

lucc:
@dcbaker I tried to work through your comments. Do you think the stuff you said about debuggers is ok like this or should I also add an "expected" variable?

dcbaker:
Looks good, thanks. I think it's fine without an expected variable, that's easy enough to view through inspection.

PR #980, opened 2017-01-17
Pr/fix 968 and 979

dcbaker:
This fixes some small bugs/quirks around the way BufferCloseCommand and ExitCommands interact with `global.quit_on_last_bclose`, both when it's set and when it's not.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -414,8 +414,23 @@ class BufferCloseCommand(Command):
 
     @inlineCallbacks
     def apply(self, ui):
+        def last_buffer_no_close():
+            """Helper that prints a message to the log and notifies if this is
+            the last buffer and the settings don't allow closebuffer to
+            exit."""
+            msg = ('not closing last remaining buffer as '
+                   'global.quit_on_last_bclose is set to False')
+            logging.info(msg)
+            ui.notify(msg, priority='error')
+
         if self.buffer is None:
             self.buffer = ui.current_buffer
+
+        # If there is only one buffer and the settings don't allow using
+        # closebuffer to exit, then just stop.
+        if len(ui.buffers) == 1 and not settings.get('quit_on_last_bclose'):
+            last_buffer_no_close()
+            return
 
         if (isinstance(self.buffer, buffers.EnvelopeBuffer) and
                 not self.buffer.envelope.sent_time):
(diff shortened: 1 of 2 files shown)


alot/commands/globals.py
```diff
@@ -414,9 +414,24 @@ def __init__(self, buffer=None, force=False, redraw=True, **kwargs):
...
+            msg = ('not closing last remaining buffer as '
+                   'global.quit_on_last_bclose is set to False')
+            logging.info(msg)
+            ui.notify(msg, priority='error')
+
         if self.buffer is None:
             self.buffer = ui.current_buffer
 
+        # If there is only one buffer and the settings don't allow using
+        # closebuffer to exit, then just stop.
```
pazz: The formulation of this docstirng is rather confusing:
actually, the settings spoecify not to exit the UI when/after closing the last buffer.

This also means that we *should* close the last buffer but keep the ui running (without open buffers), so we do not want to abort here..

dcbaker: Is it possible to close the last buffer without closing? I tested closing the last buffer without exiting, and I'm getting a stack trace because the code assumes that there will always be one buffer open. The current behavior is to not close the last buffer, but without a notification of why.

Maybe it would be better to have a comment like:
Alot always needs to have at least one buffer opened, if this is the last buffer and the configuration is not quit when the last buffer closes then don't close the buffer, and provide a message as to why.

pazz:
Quoting Dylan Baker (2017-01-18 17:39:19)
>    Is it possible to close the last buffer without closing? I tested closing
>    the last buffer without exiting, and I'm getting a stack trace because the
>    code assumes that there will always be one buffer open.

Ah! I remember that this worked at some point, and the interface would
simply show an "empty buffer", i.e., blank screen, and only accept
global commands.

dcbaker:
So, that's another bug we should fix?

pazz:
Quoting Dylan Baker (2017-01-18 17:50:18)
>    So, that's another bug we should fix?

We need to decide if we want to allow a bufferless interface or not.
If not, adding a notification here as you suggest would suffice.
If yes, we need to make sure that nothing incorrectly assumes
that there is always a buffer.
I have no opinion on this really. Adding a note seems easier. Your call.

dcbaker:
Well, I think in the short term this doesn't change the behavior, and is a strict improvement since it provides a message as to why the last buffer doesnt close (the message was oringally only written to the log, but I copied it into ui notification). If we want to allow a bufferless interface that's a bigger patch and change, and we could implement that as a later patch, we probably want to open an issue for that if we do.

pazz:
OK agreed. I'll push this then. thanks.

PR #981, opened 2017-01-18
helper: Don't rely on GNU extensions for strftime

dcbaker:
While the reality is that most alot users probably have glibc as their C
library and can therefore rely on %P, it is possible that there are
users or potential users who run a *BSD or other unix-like OS. This
leaves only non-GNU extensions, though some of the extensions are
specified in later POSIX, unix or C specifications.


dcbaker:
@lucc pointed this out in my unittests, if it's worth fixing there, it's probably worth fixing in alot core too.

PR #982, opened 2017-01-18
travis: Add ccache

dcbaker:
This will speed up the builds of notmuch, saving about 40 seconds of
time in travis.


dcbaker:
Note that the first build in a branch will be slower, since the cache will be empty, but each subsequent build will be faster.

PR #983, opened 2017-01-18
Some random cleanup

lucc:


alot/db/message.py
```diff
@@ -107,13 +107,11 @@ def get_thread_id(self):
         return self._thread_id
 
     def get_message_parts(self):
-        """returns a list of all body parts of this message"""
+        """yield all body parts of this message"""
         # TODO really needed? email  iterators can do this
```
pazz: perhaps we rather want to use those email iterators in that for loop?

lucc: I didn't see how any of [this](https://docs.python.org/2/library/email.iterators.html) fits the bill. So maybe we should remove the comment? Or am I missing something?

pazz: yes you are right. A quick "fix" would be to just remove the comment.
But this function is actually only used once, and there we could instead also
use `https://docs.python.org/2/library/email.iterators.html#email.iterators.typed_subpart_iterator`..

lucc: Sorry I don't understand how this iterator can be used. It seems to me that it can only return parts of a specific type but our if-else dance checks for several different content types. So we would have to iterate over the email several times?

dcbaker: I'm with @lucc on this. we'd have to call email.iterators.typed_subpart_iterator multiple times (since mainpart doesn't take a `*` argument).

lucc:
I removed the comment.

lucc:
I ran the following shell command  on master to find unused functions  (like in the first commit here)
```sh
sed -n 's/ *def \([a-zA-Z0-9_]*\)(.*):/\1/p' alot/**/*.py | \
  sort -u | \
  while read line; do
    grep --no-filename $line alot/**/*.py | \
      grep -vq '^ *def ' || echo $line
  done
```
and it returns this list:
```
__call__
collapse_all
collapse_if_matches
__eq__
expand_all
focus_next_matching
focus_prev_matching
get_headers_string
__getitem__
get_main_addresses
get_messagetree_positions
get_prev
_get_theme
__hash__
has_replies
kill_search_processes
__lt__
messagetree_at_position
__ne__
processEnded
replace_bodytext
set_focussed
__setitem__
set_unfocussed
```
Might it be possible to remove some of these? I assume things like `__lt__` are never called because they implement comparison with `<` so I will not find the call like this. But are some of these names special to urwid or some other thing we subclass and will not be called directly in our code base but in the urwid main loop for example?

pazz:
just checked `db.manager.Manager.kill_search_processes`: It is indeed never used, as the search buffers will individually kill their subprocesses on exit.

I suspect that some of these will fall into the category "makes sense to have" for a consistent API to be used in hooks etc. For instance `focus_next_matching`, although I'm surprised it;s not used by the `move` commands in threadbuffers.

lucc:
I am against keeping unused code in general.

If we want to keep some of this code to be used in hooks (by users I suppose) it *needs* to be discoverable. But if it is mentioned in the docs or the wiki as "stuff you can use in hooks for convenience" it effectively becomes part of the public API. And then we have to add a comment to the code saying "don't remove this it is not used but part of the public API" or so.

Or we could move the code that is interesting for user hooks into a separate module `hook_helpers.py` (but that might be harder with methods or we have to wrap the original method with a function in the hook_helper.py).

But the fact that we keep code around just "it might be useful to somebody somehow" is very unsatisfactory to me. Especially if we have to ensure it functions correctly when we don't even know what "correctly" means in this case as somebody else has intensions to use it and not we.

pazz:
Alright, let's remove unnecessary code where possible. But: note the TODOs in thread buffer's `move` command regarding `move next matching`. This could actually be implemented using the methods I mentioned above. So I'd rather do that in this particular case.

dcbaker:
I agree with removing unused code or marking that it is part of the public API. As you note that code won't correctly indentify unused dunder methods. Fortunately there are few enough of those manually looking at them is probably fine.

lucc:
I merged this. Further removal of some of these functions can happen in another PR.

PR #988, opened 2017-01-19
Add `move next matching` command in thread buffer

lucc:
This is a POC and discussion of the `move next matching` and `move previous matching` commands mentioned in #983.  I am able to use the command like this:
```
$ PYTHONPATH=. python2 -m alot
:search some thread with many messages
:select
:pyshell
>>> ui.current_buffer.querystring = u'some short text that will match somewhere in the thread'
>>> <CTRL-D>
:move next matching
```

What kind of syntax/command do we want use to set the querystring in a thread buffer? Maybe `:move next matching /foo/` or so. Or we could extend the `next matching` command to prompt for a search string if none was defined.  And we could map `/` and `n`.


pazz:
The move command is not properly documented anyway:
Both in search and thread buffers, these inherit from `globals.MoveCommand`, so also allow for say, `move halfpage up`.
Maybe we want to distinguish `move` from `focus`? the former just moves up/down etc without a "target", and the latter moves *to* some defined target like next/prev mail, sibling, or msg matching some query?

In any case, syntaxwise, I think I'd like something like `move next <query>`, which would focus on the DFO-next message matching that query. The current behaviour is then just `move next *`.
The same can be implemented for previous, (sibling, child)... Thoughts?

lucc:
The syntax of `move next search-term` is quite nice, I think. We could implement the parser for the others by just recognizing the special search terms `sibling`, `unfolded` and so on.

lucc:
I tried to add some "documentation". Is there more that can be done, and where?

I also changed the syntax to the suggested `move next query...` version. Should this also implement some kind of xapian parser? The comments at the end of `alot.commands.thread.MoveFocusCommand.apply` suggest some like this to me:
- `move next date monday`
- `move next date 2017`
- `move next date january`
- or maybe rather `move next date:foo`
- and also `move next tag:inbox` or `move next tag inbox`

Or can some of these queries really be passed to notmuch? Can a message be checked against a query?

What else do these comments mean? What is the idea with the "original query in threadbuffers"?

I am unsure about the focus vs move distinction from a usability perspective. Move would be the global command and focus the one specific to thread and search buffers? What was your idea why this is usefull?

pazz:
Quoting Lucas Hoffmann (2017-08-04 07:25:59)
> I tried to add some "documentation". Is there more that can be done, and where?

I agree with tha changes in 57bb6c1f622cd33becd341db4d12dda4784133f0
All movements should be documented somewhere and this looks as good a
place as any other.


> I also changed the syntax to the suggested move next query... version. Should
> this also implement some kind of xapian parser?

I'd really like to avoid reimplementing anything that the notmuch lib
does for us, including parsing queries. So in particular it's a bad idea
to start changing the query syntax for use only within alot.
Some early notmuch GUIs did that and its a messy, slippery road to
oblivion IMHO.

> The comments at the end of alot.commands.thread.MoveFocusCommand.apply
> suggest some like this to me:
> 
>   • move next date monday
>   • move next date 2017
>   • move next date january

no (see above)

>   • or maybe rather move next date:foo
>   • and also move next tag:inbox or move next tag inbox

Yes! these suffixes look like valid notmuch queries.
`date:monday` is valid notmuch syntax.
 
> Or can some of these queries really be passed to notmuch? Can a message be
> checked against a query?

Yes:
http://alot.readthedocs.io/en/latest/api/database.html#alot.db.Message.matches
 
> What else do these comments mean? What is the idea with the "original query in
> threadbuffers"?

These were note-to-selfs, so please excuse the imprecision.
It was meant to reming me of the idea of remembering, in the
threadbuffer, the query string used in its "parent" searchbuffer that
was used to open the thread.

Then "next matching" would focus the next message in this thread, which
matches the searchbuffer-query (remember that a thread matches a query
if and only if at least one of its messages does).

At the moment, the threadbuffer does not know its parent's query string,
so this focus movement is not possible.

I was (and am still not) conviced that implementing such "strong
coupling" is a good idea and this focus feature did not seem important
enough to me at that point.


> I am unsure about the focus vs move distinction from a usability perspective.
> Move would be the global command and focus the one specific to thread and
> search buffers? What was your idea why this is usefull?

that was the idea yea. I was more concearned with a sensible code
structure than with usability. The fact that `move` is global, which i
think makes sense, means that it is tricky to implement and document
all the buffer specific focus options..
I'm open for suggestions though.

lucc:
I just had a look at the code of `alot.buffers.ThreadBuffer.focus{next,prev}_matching` and that already uses `alot.db.Message.matches` internally. And I tested it, yes the current implementation can already deal with `:move next is:signed and date:2017-07-16`.  A fix for `or` is comming up.

lucc:
I think the updated docs make it clear what is possible now. Can we remove these "note to self" comments? Are we still missing something?

I do not really care for the strong coupling you mentioned.

docs/source/usage/modes/thread.rst
```diff
@@ -40,7 +40,7 @@ The following commands are available in thread mode
     move focus in current buffer
 
     argument
-        up, down, page up, page down, first, last
+        up, down, [half]page up, [half]page down, first, last,                   parent, first reply, last reply,                   next sibling, previous sibling, next, previous,                   next unfolded, previous unfolded,                   next NOTMUCH_QUERY, previous NOTMUCH_QUERY
```
pazz: The spacing looks funky here.. was this intended?

pazz:
what does `move next` do after this change exactly? will it act just like `move next *` and focus on the next message in depth-first order of the thread?

docs/source/usage/modes/thread.rst
lucc: This is autogenerated from the docstring. I tried to format the docstring and did not care about this. This whitespace will become irellevant in the html output.

lucc:
The code and functionallity for `move next` is unchaged. So yes it work like `move next *`.

PR #989, opened 2017-01-20
Add docstrings to some methods and functions.

lucc:
I randomly picked some of the places where it was easy to make out what to write.  Please have a look if I captured the semantics of the code correctly.

### DIFF
--- a/alot/account.py
+++ b/alot/account.py
@@ -178,6 +178,7 @@ class Account(object):
 class SendmailAccount(Account):
     """:class:`Account` that pipes a message to a `sendmail` shell command for
     sending"""
+
     def __init__(self, cmd, **kwargs):
         """
         :param cmd: sendmail command to use for this account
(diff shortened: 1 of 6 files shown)


alot/account.py
```diff
@@ -20,6 +20,7 @@ class StoreMailError(Exception):
 
 
 class Account(object):
+
```
pazz: why do you insert empty new lines here (and below)?

lucc: The templates I use when writing new code have that so I'm used to it... Will remove them.

lucc: What about the empty line below the docstring? (I assume you meant the other docstring when you said "below")

pazz: yes, I meant the other docstring. I don't really mind, just wanted to make sure we stick to one convention.

lucc: I fixed the newline above the docstrings so you can unblock this.

pazz:
thanks

PR #992, opened 2017-01-21
widgets/globals: Fix typo which would lead to AttributeError

dcbaker:


pazz:
good catch! I wonder why this never cause any problems so far. Did it for you?

dcbaker:
Nope, one of the syntax checkers highlighted it while i was in there for something else.

Issue #889, opened 2016-07-28
bug: parse address list from prompt


pazz:
When I prompt for a list of recipients in envelope mode and type in 

>  "X X" x@x.com, "Y Y" y@y.com

The `To` header gets assigned the value

>  "X X x@x.com, Y Y" y@y.com

which is malformed and leads to mail delivery errors.

pazz:
note to self: there should be a test for this

lucc:
I can only reproduce this half way. I use `msmtp --file /home/luc/.config/msmtp/msmtprc --read-recipients --read-envelope-from` as my sendmail_command and I can send a mail just fine if a set the to field to `"Luc, der Hoffmann" <my-vaid-email@inter.net>`. I get this error popup in alot though:
```
Can not find key for '"Luc'.
Can not find key for 'my-vaid-email@inter.net'.
```
but I assume that is due to the autoencrypt option. To me this shows that we indeed do something undesirable to the to field but also that the to field is not passed to the sendmail_command (or is it just my msmtp --read-recipients that does not care about the arguments given?)

What kind of delivery errors do you get exactly?

pazz:
Sorry, the initial statement wasn't too clear.
The point is that `"X X x@x.com, Y Y" y@y.com` has a different semantics than what I intended (and typed). This describes a single recipient with address ` y@y.com` and name `X X x@x.com, Y Y`.

lucc:
Yes I get the result that you intend (one recipient with a strange name). What do you get?

pazz:
I get the same. but what I intended (and typed) was two recipients!

pazz:
fixed in #993

Issue #990, opened 2017-01-20
missing quotes for auto-completed contact


klement:
Hi,
I'm using the notmuch address approach for getting contacts completion with the regexp per docs:
```
regexp = (\"?(?P<name>.+)\"?)?\s*<(?P<email>.*@.+?)>
```
but when I auto-complete the address, instead of getting `"name" <address>` it completes as `name" <address>` so the header looks like this:
```
To: name" <address>
```
instead of:
```
To: "name" <address>
```

lucc:
The behavior you see is consistent with the regex. The first quote will not be caught into the group `<name>`. The second quote will always end up in that group as it is matched by `.*` (remember `*` catches as much as it can) and the second `\"?` is never matched as it is optional. If the quotes were not optional the `.*` would be forced not to match it but then the whole thing would only match if there are quotes around the realname, and the whole idea behind the double groups and optional quotes is to match email addresses with and without realnames and the former in turn with and without quotes.

So the conclusion is that the regex given in the docs is buggy. We should try to fix it but that might be very hard, because matching the realname part of an email address with regex is hard (it might for example contain backslash escaped quotes and so on).

PS: I formatted your post for you. Please use [markdown](https://guides.github.com/features/mastering-markdown/) to format your posts yourself in the future.  They are much easier to read if you do.

klement:
hmm, maybe the whole regexp could be designed as:

> "a"<b>|<b>

with quotes being mandatory in the first alternative...

lucc:
This might be to restrictive (at least for my database, but I use another command altogether so it might still work for you). If I search my database like this: `notmuch address '*'` I get
* emails without a real name, they have no `<` and `>` around them
* emails with a real name, without quotes around the real name
* emails with either single or double quotes around the realname
* emails with single or double quotes around the realname and the other type of quotes in the realname
* surprisingly no quotes of the same type inside a quoted realname, even though I believe I have seen this before, somewhere.
* all emails that have a realname part have at least one space after the realname

klement:
while I don't see the describe behaviour, it seems the real problem here is the lack of structur

`notmuch address --format=json` seems to be the answer

the output looks like this:

`{"name": "", "address": "<email>", "name-addr": "<email>"}`
 for emails only

`{"name": "<name>", "address": "<email>", "name-addr": "\"<name>\" <email>"}`

for addresses containing both name and address...

I tried using:

[...]

PR #993, opened 2017-01-21
[RFC] Fix addressbook completion

lucc:
There are some fine corner cases when generating a TO header from an address book.  Characters like `,'"` might need special treatment.  I opend this PR early (the code is not yet finished) in order to discuss some of the things with you.

Related: #990, #889

Points to check, discuss and decide:
* [x] look up the RFC how To headers have to be formatted
* [ ] write test cases for all corner cases
* [x] decide how we want to treat which special chars in To headers and if we should add quotes in cases where they are not needed, "just to have it look uniform"
* [x] implement our decisions in alot.completion.AbooksCompleter
* [ ] check the test for proper coding style (@dcbaker is this how I should write the tests and the mock stuff?)
* [ ] more tests?
* [ ] you name it …

At the moment I just wrote some tests to show you what I think are the cases we have to take care of.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import
 import argparse
 import code
 import email
+import email.utils
 import glob
 import logging
 import os
@@ -805,7 +806,7 @@ class ComposeCommand(Command):
             accounts = settings.get_accounts()
             if len(accounts) == 1:
                 a = accounts[0]
-                fromstring = "%s <%s>" % (a.realname, a.address)
+                fromstring = email.utils.formataddr((a.realname, a.address))
                 self.envelope.add('From', fromstring)
             else:
                 cmpl = AccountCompleter()
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -287,7 +287,16 @@ class ReplyCommand(Command):
 
     @staticmethod
     def clear_my_address(my_addresses, value):
-        """return recipient header without the addresses in my_addresses"""
+        """return recipient header without the addresses in my_addresses
+
+        :param my_addresses: a list of my email addresses (no real name part)
+        :type my_addresses: list(str)
+        :param value: a list of recipient or sender strings (with or without
+            real names as taken from email headers)
+        :type value: list(str)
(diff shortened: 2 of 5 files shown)


pazz:
(I copy my previous comment here so as to close the other issue.)

I have one recipient of the form `"lastname, firstname" <x@y.com>`.  And notmuch address outputs

```
{"name": "lastname, firstname", "address": "x@y.com", "name-addr": "\"lastname, firstname\" <x@y.com>"}
```

Notice the missing quotes in the name part. They should be there, because otherwise the comma will signal the end of the recipient, when using this string as value for the `To` header.

> Maybe we should check that the name does not mess up our To header in general?

yes, I think this should go here:
https://github.com/pazz/alot/blob/master/alot/completion.py#L233
I thought about just quoting the realname bit, so using `\"%s\" <%s>` there.
But actually, what happens if the name contains an odd number of quotes itself?

lucc:
One of the last days I was wondering if our problem was not already solved and really, it is: [email.utils.formataddr()](https://docs.python.org/2/library/email.util.html#email.utils.formataddr). 

I would say that is the correct way to do the quoting (so I checked the point in my initial post, uncheck and comment if you disagree)

pazz:
Great find! So this is a small matter of wrapping this heper function around the the line I linked.
Thanks!

lucc:
I added some commits that should implement this. I also added commits to use email.utils.farmataddr in other places. I wanted to use it in alot/db/utils.py line 411 as well but am not sure about the encode stuff there. So for now I left it. I you can comment on this with more insight I can also use the formataddr there.

I think we don't have to actually look up the RFC any more, I trust python on this, so I checked that item :)

pazz:
OK, let's rename that 'input' variable before pushing to make the codacity bot happy.

lucc:
fixed the comments

pazz:
thanks, LGTM

lucc:
I rebased and removed the `expectedFailure`s.

lucc:
@pazz does this fix #889?

pazz:
yes it does. I'll merge this and close #889 as well. Thanks!

Issue #994, opened 2017-01-23
fix docs build on travis


pazz:
see #947 
the docs build on travis fails on the current master branch (22af804982)
with

```
/home/travis/build/pazz/alot/docs/source/usage/commands.rst:27: WARNING: usage/modes/global is already assigned section numbers (nested numbered toctree?)
```

dcbaker:
I can verify that it is a sphinx issue. I ran `make -C docs cleanll html` with sphinx 1.5.1 and everything ran fine, then I updated my virtualenv to 1.5.2 and I got the errors we're seeing in read the docs.

dcbaker:
This might not be our bug, it might be sphinx's bug.

dcbaker:
Complete sphinx 1.5.2 output:
```
Running Sphinx v1.5.2
making output directory...
loading pickled environment... not yet created
loading intersphinx inventory from http://notmuch.readthedocs.org/en/latest/objects.inv...
intersphinx inventory has moved: http://notmuch.readthedocs.org/en/latest/objects.inv -> http://notmuch.readthedocs.io/en/latest/objects.inv
loading intersphinx inventory from http://docs.python.org/objects.inv...
intersphinx inventory has moved: http://docs.python.org/objects.inv -> https://docs.python.org/2/objects.inv
loading intersphinx inventory from http://urwid.readthedocs.org/en/latest/objects.inv...
intersphinx inventory has moved: http://urwid.readthedocs.org/en/latest/objects.inv -> http://urwid.readthedocs.io/en/latest/objects.inv
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 32 source files that are out of date
updating environment: 32 added, 0 changed, 0 removed
reading sources... [100%] usage/synopsis                   
looking for now-outdated files... /home/dylan/source/alot/docs/source/usage/commands.rst:27: WARNING: usage/modes/global is already assigned section numbers (nested numbered toctree?)
/home/dylan/source/alot/docs/source/usage/commands.rst:27: WARNING: usage/modes/search is already assigned section numbers (nested numbered toctree?)
/home/dylan/source/alot/docs/source/usage/commands.rst:27: WARNING: usage/modes/thread is already assigned section numbers (nested numbered toctree?)
/home/dylan/source/alot/docs/source/usage/commands.rst:27: WARNING: usage/modes/envelope is already assigned section numbers (nested numbered toctree?)
[...]

PR #999, opened 2017-01-24
travis: work around sphinx-1.5.2 problem

dcbaker:
This forces travis to install sphinx 1.5.1 which doesn't display the
warnings 1.5.2 does. This is not a proper solution, but a work around so
that the docs test works and remains useful. We still need to implement
a proper fix.

I am not marking this as a fix for the problem, we need to figure out whether we're doing something wrong, or whether this is a bug in sphinx, and address that accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ install:
       printf '%s = None\n' NotmuchError NullPointerError > notmuch.py
       touch gpgme.py
       # install sphinx for building the html docs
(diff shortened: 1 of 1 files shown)


dcbaker:
This is a work around for #994

dcbaker:
@lucc @pazz Does this look acceptable to merge in the short term so lucc and I can merge things that otherwise wouldn't pass the 'docs' job?

PR #1000, opened 2017-01-24
argparse validators

dcbaker:
This series creates a custom action type for argparse, a validated store action. Arguments created with `argparse.add_argument` that set their action to `ValidatedStore` gain a new validator keyword, which takes a callable, this callable will be used to test the input, and will raise an `argparse.ArgumentError` if the validation fails. It also adds validators for all of the arguments used in `__main__.py` that either previously used `type` for validation, or that had comments marking that they should be validated.

This approach has a couple of advantages over using `type` for validation. The first is that it frees `type` to be used for it's actual purpose, chaning a the string that argparse gets into another type, like an `int` or `float`. Second, it fixes the use of pipes or character special devices (like /dev/null) as ouput.

It comes complete with unittests and passes travis (including doc tests when not built with sphinx 1.5.2)

### DIFF
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -15,6 +15,7 @@ from alot.db.manager import DBManager
 from alot.ui import UI
 from alot.commands import *
 from alot.commands import CommandParseError, COMMANDS
+from alot.utils import argparse as cargparse
 
 
 _SUBCOMMANDS = ['search', 'compose', 'bufferlist', 'taglist', 'pyshell']
@@ -27,23 +28,29 @@ def parser():
                         version=alot.__version__)
     parser.add_argument('-r', '--read-only', action='store_true',
                         help='open db in read only mode')
-    parser.add_argument('-c', '--config', help='config file',
-                        type=lambda x: argparse.FileType('r')(x).name)
+    parser.add_argument('-c', '--config',
+                        action=cargparse.ValidatedStoreAction,
+                        validator=cargparse.require_file,
+                        help='config file')
     parser.add_argument('-n', '--notmuch-config', default=os.environ.get(
                             'NOTMUCH_CONFIG',
                             os.path.expanduser('~/.notmuch-config')),
-                        type=lambda x: argparse.FileType('r')(x).name,
+                        action=cargparse.ValidatedStoreAction,
+                        validator=cargparse.require_file,
                         help='notmuch config')
     parser.add_argument('-C', '--colour-mode',
                         choices=(1, 16, 256), type=int, default=256,
                         help='terminal colour mode [default: %(default)s].')
-    parser.add_argument('-p', '--mailindex-path', #type=directory,
+    parser.add_argument('-p', '--mailindex-path',
+                        action=cargparse.ValidatedStoreAction,
+                        validator=cargparse.require_dir,
                         help='path to notmuch index')
     parser.add_argument('-d', '--debug-level', default='info',
                         choices=('debug', 'info', 'warning', 'error'),
                         help='debug log [default: %(default)s]')
     parser.add_argument('-l', '--logfile', default='/dev/null',
-                        type=lambda x: argparse.FileType('w')(x).name,
+                        action=cargparse.ValidatedStoreAction,
+                        validator=cargparse.optional_file_like,
                         help='logfile [default: %(default)s]')
     # We will handle the subcommands in a seperate run of argparse as argparse
     # does not support optional subcommands until now.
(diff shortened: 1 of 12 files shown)


alot/__main__.py
```diff
@@ -14,56 +14,72 @@
...
     else:
         command = None
 
+    return options, command
+
+
+def main():
+    """The main entry point to alot.  It parses the command line and prepares
+    for the user interface main loop to run."""
+    # set up the parser to parse the command line options.
```
lucc: This comment has to be moved.

alot/commands/actions.py
```diff
@@ -0,0 +1,54 @@
...
+    """Exception raised when Validation fails in a ValidatedStoreAction."""
+    pass
+
+
+class ValidatedStore(argparse.Action):
+    """An action that allows a validation function to be specificied.
+
+    The validator keyword must be a function taking exactly one argument, that
+    argument is a list of strings or the type specified by the type argument.
+    It must raise ValidationFailed when with a message when validation fails.
```
lucc: remove the first "when"

alot/commands/validators.py
```diff
@@ -0,0 +1,85 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Functions for validating command options.
+
+For use with the alot.commands.utils.Validator*Action.
```
lucc: That should be `alot.commands.actions.ValidatedStore`.

tests/commands/validators_test.py
```diff
@@ -0,0 +1,131 @@
...
+            validators.require_file(f.name)
+
+    def test_char_special(self):
+        with self.assertRaises(ValidationFailed):
+            validators.require_file('/dev/null')
+
+    def test_fifo(self):
+        with temporary_directory() as d:
+            path = os.path.join(d, 'fifo')
+            subprocess.call(['mkfifo', path])
```
lucc: Why not `os.mkfifo(path)`? That would allow you to ditch `subprocess` and would probably use the system call directly instead of using the external program.

pazz:
1000! party

Seriously though, good idea.
I'm not sure about burying this inside `alot/commands` when we use it in the main file as well.

alot/commands/validators.py
dcbaker: oops, I guess I didn't check all of the docstrings well enough after I finished renaming things.

tests/commands/validators_test.py
dcbaker: Yup, that sounds much better.

dcbaker:
@pazz I don't really care where we put them, but since they're about commands I thought alot/commands seemed like a good place. I was thinking that some follow up work for this would be to add validators to commands create with `registerCommand`, and have a handler there that could send the result to ui as an error if something didn't validate, so command seemd like a good place for them.

pazz:
quick question: what is the rationale behind using actions, and not (special) types here?
As far as I understand, you could also define a function from string to "validated string" and use that in combination with argparse's `type` parameters, right? 

Regarding where to put it: we could somehow put it together with the custom checks for configobject, which currently live in `alot/settings/checks` and do essentially the same job for validating config settings.
Ah: and there is also another argparse action in `utils/booleanaction.py`.
So how about putting all argparse extensions into `utils/argparse`, and the configobj stuff into `utils/configobj` ?

dcbaker:
Why use an action? I think the crux of it is that validation is a different action than setting the type of the argument. You can overload type to do what you want, but it's a hack. This is what a custom action is for, and it's clear, conscise, and easy to extend.

Sure. I'll move that stuff around.

dcbaker:
@pazz does that look good?

dcbaker:
@pazz other than codacy complaining about unittest being stupid or implementing well defined interfaces, are you okay with merging this?

pazz:
just wen through the code and it looks good to me, thanks.
By the way: we can tell the codacity bot not to complain about certain classes of issues. I believe you should have write access through your github account @dcbaker . Feel free to remove stuff you feel should go. That bot is a bit too slow to be useful anyways..

lucc:
So I conclude that we are ready for 0.5 and this is indeed the [party thread](https://github.com/pazz/alot/pull/1000#issuecomment-274916254) :D

pazz:
I'll make the release on the weekend, unless you'd like to..

lucc:
I'll leave it to you.

pazz:
OK

PR #1005, opened 2017-01-25
Add a hook that runs periodically

dcbaker:
This adds the much sough after functionality to run arbitrary commands on a timer. This is acheived using twisted's event loop (the reactor), and a `LoopingCall`. Basic error handling is provided as well (errors are sent to the log).

### DIFF
--- a/alot/defaults/alot.rc.spec
+++ b/alot/defaults/alot.rc.spec
@@ -248,6 +248,9 @@ reply_account_header_priority = force_list(default=list(From,To,Cc,Envelope-To,X
 #           history file might get *very* long.
 history_size = integer(default=50)
 
+# The number of seconds to wait between calls to the loop_hook
+periodic_hook_frequency = integer(default=300)
+
 # Key bindings
 [bindings]
     __many__ = string(default=None)
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -311,8 +311,7 @@ class SettingsManager(object):
     def get_hook(self, key):
         """return hook (`callable`) identified by `key`"""
         if self.hooks:
-            if key in self.hooks.__dict__:
-                return self.hooks.__dict__[key]
+            return getattr(self.hooks, key, None)
         return None
 
     def get_mapped_input_keysequences(self, mode='global', prefix=u''):
--- a/docs/source/configuration/hooks.rst
+++ b/docs/source/configuration/hooks.rst
@@ -205,3 +205,10 @@ Apart from command pre- and posthooks, the following hooks will be interpreted:
     :type suffix: str
     :returns: tuple of `prefix` and `suffix`
     :rtype: (str, str)
+
+.. py:function:: loop_hook(ui=None)
+
+    Run on a period controlled by :ref:`_periodic_hook_frequency <periodic-hook-frequency>`
+
+    :param ui: the main user interface
+    :type ui: :class:`alot.ui.UI`
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,7 @@ next release:
 * enable variable interpolation in config file
 * Add encryption to CC addresses
 * Add bufferlist, tablist and pyshell subcommands to the command line interface
+* Add hook that runs periodically.
 
 0.4:
 * signal: refresh current buffer on SIGUSR1
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -7,7 +7,7 @@ import logging
 import os
 import signal
 
-from twisted.internet import reactor, defer
+from twisted.internet import reactor, defer, task
 import urwid
 
 from .settings import settings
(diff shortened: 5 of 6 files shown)


docs/source/configuration/hooks.rst
```diff
@@ -194,3 +194,12 @@ Apart from command pre- and posthooks, the following hooks will be interpreted:
     :type suffix: str
     :returns: tuple of `prefix` and `suffix`
     :rtype: (str, str)
+
+.. py:function:: loop_hook(ui=None)
+
+    Run on a period controlled by settings:loop_period.
```
pazz: you can use 
```
:ref:`loop_hook <loop-hook>`
```
to refer to your new config option here.

docs/source/configuration/hooks.rst
```diff
@@ -194,3 +194,12 @@ Apart from command pre- and posthooks, the following hooks will be interpreted:
...
+
+    All errors are logged by default.
```
pazz: I don't think this is necessary to say. kind of obvious no?

alot/defaults/alot.rc.spec
```diff
@@ -248,6 +248,9 @@ reply_account_header_priority = force_list(default=list(From,To,Cc,Envelope-To,X
 #           history file might get *very* long.
 history_size = integer(default=50)
 
+# The number of seconds to wait between calls to the loop_hook
+loop_period = integer(default=300)
```
pazz: I'm not a fan of the name: How about "periodic_hook_frequency"?
although technically this describes the period, not a frequency.
By the way. Are you sure that this is in seconds, not milliseconds?

alot/settings/manager.py
```diff
@@ -311,8 +311,7 @@ def colourpick(triple):
     def get_hook(self, key):
         """return hook (`callable`) identified by `key`"""
         if self.hooks:
-            if key in self.hooks.__dict__:
-                return self.hooks.__dict__[key]
+            return getattr(self.hooks, key, None)
```
lucc: Isn't None the default?

dcbaker: No, with `getattr` if you don't provide a fallback then an `AttributeError` is raised.
https://docs.python.org/2/library/functions.html#getattr

alot/defaults/alot.rc.spec
dcbaker: what about "loop_hook_frequency" or "loop_hook_period"?

Yes, `LoopingCall`'s interval argument is defined in seconds.

pazz:
Quoting Dylan Baker (2017-01-26 18:36:20)
>    what about "loop_hook_frequency" or "loop_hook_period"?

I was trying to avoid "loop" because it sounds like "main loop" to me..
Ho about "periodic_hook_sleeps" ?
In any case, it's not the end of the world for me if we go with "loop"
:)

More important is the possibility to switch this feature off. Maybe by
setting this config setting to 0?

>    Yes, LoopingCall's interval argument is defined in seconds.

OK thanks!

dcbaker:
> I was trying to avoid "loop" because it sounds like "main loop" to me..
Ah, okay. In that case I like your original idea better.

> More important is the possibility to switch this feature off. Maybe by setting this config setting to 0?
If you don't define a loop_hook then this won't get enabled. Is that enough, or would you like an additional switch?

pazz:
Quoting Dylan Baker (2017-01-26 18:51:22)
>      I was trying to avoid "loop" because it sounds like "main loop" to me..
>      Ah, okay. In that case I like your original idea better.

cool, your call.

>      More important is the possibility to switch this feature off. Maybe by
>      setting this config setting to 0?
>      If you don't define a loop_hook then this won't get enabled. Is that
>      enough, or would you like an additional switch?

Ah OK, so this defaults to `None`? I think, sinse a period of $0$ makes
little sense anyway, we could additionally interpret 0 to disable the
feature.

dcbaker:
> Up for discussion: Do you think that people will want to run several different periodic hooks or would one suffice?

One of the ideas I've had is to replace the hook module with a hook package, which would allow people to implement multiple functions called at the same time. The advantage of this would be that if you need to have cleanups you can implement a single module with the right hooks for a feature.

> Ah OK, so this defaults to `None`? I think, sinse a period of $0$ makes
little sense anyway, we could additionally interpret 0 to disable the
feature.

Right, it uses `settings.get_hook`, and then does 'if hook', so if there was no hook defined you get no loop. I don't know how useful it is to have a settings that says "I defined a loop_hook, but please don't run it."

alot/settings/manager.py
lucc: ok, sorry mixed it up with stuff like dict.get()

dcbaker: Yup. I had to look to see. It's strange that getattr and dict.get behave differently given that getattr is effecitvely a wrapper around dict.get.

PR #1008, opened 2017-01-26
Default to not handling mouse

dcbaker:
As discussed elsewhere, handling the mouse was not done before, and
breaks work flows that include using the mouse for other things
(middle-clock paste). This only changes the default option to False.


dcbaker:
Just noticed that, an update is on its way

PR #1009, opened 2017-01-26
Remove remaining uses of cmp

dcbaker:
These two patches replace our remaining uses of `cmp` with built in rich comparisons. This should both optimize building the tag widgets (the final implementation takes < 30%) as long as the cmp implementation in my tests, though it should be noted there will be some additional cost to calling a method. This is also necessary for python 3, which doesn't have cmp.

This is the test script I used to verify the speed of the optimizations, in case anyone is interested:

```python
import functools
import timeit


def cmp_(a, b):
    if min(len(a), len(b)) == 1 and max(len(a), len(b)) > 1:
        return cmp(len(a), len(b))
    return None


def orig(a, b):
    a = len(a)
    b = len(b)
    if min(a, b) == 1 and max(a, b) > 1:
        return a < b
    return None


def orig2(a, b):
    if min(len(a), len(b)) == 1 and max(len(a), len(b)) > 1:
        return len(a) < len(b)
    return None


def xor(a, b):
    a = len(a)
    b = len(b)
    if a == 1 ^ b == 1:
        return a < b
    return None


def xor2(a, b):
    if len(a) == 1 ^ len(b) == 1:
        return len(a) < len(b)
    return None


print(timeit.timeit(functools.partial(cmp_, 'a', 'aaaaa')))
print(timeit.timeit(functools.partial(orig2, 'a', 'aaaaa')))
print(timeit.timeit(functools.partial(orig, 'a', 'aaaaa')))
print(timeit.timeit(functools.partial(xor2, 'a', 'aaaaa')))
print(timeit.timeit(functools.partial(xor, 'a', 'aaaaa')))
```

And the results I think speak for themselves:
```
0.601747989655
0.534921169281
0.44206905365
0.177870988846
0.185997009277
```

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -498,17 +498,6 @@ def shell_quote(text):
     return "'%s'" % text.replace("'", """'"'"'""")
 
 
-def tag_cmp(a, b):
-    r'''
-    Sorting tags using this function puts all tags of length 1 at the
-    beginning. This groups all tags mapped to unicode characters.
-    '''
-    if min(len(a), len(b)) == 1 and max(len(a), len(b)) > 1:
-        return cmp(len(a), len(b))
-    else:
-        return cmp(a.lower(), b.lower())
-
-
 def humanize_size(size):
     """Create a nice human readable representation of the given number
     (understood as bytes) using the "K" and "M" suffixes to indicate kilo- and
--- a/alot/widgets/globals.py
+++ b/alot/widgets/globals.py
@@ -326,7 +326,8 @@ class TagWidget(urwid.AttrMap):
 
         self_len = len(self.translated)
         oth_len = len(other.translated)
-        if min(self_len, oth_len) == 1 and max(self_len, oth_len) > 1:
+
+        if (self_len == 1) is not (oth_len == 1):
             return self_len < oth_len
         return self.translated.lower() < other.translated.lower()
 
--- a/alot/widgets/globals.py
+++ b/alot/widgets/globals.py
@@ -7,6 +7,7 @@ This contains alot-specific :class:`urwid.Widget` used in more than one mode.
 """
 from __future__ import absolute_import
 
+import functools
 import re
 import operator
 import urwid
@@ -271,6 +272,7 @@ class HeadersList(urwid.WidgetWrap):
         return headerlines
 
 
+@functools.total_ordering
 class TagWidget(urwid.AttrMap):
     """
     text widget that renders a tagstring.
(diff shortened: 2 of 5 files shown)


dcbaker:
And I think "I should write tests", and this isn't quite right. Expect updates.

dcbaker:
Okay, and updated working version, including tests! whoo.

The original version didn't use parens properly, so the Xor was applied against the wrong values, not the boolean results of the comparison to 1. This means that it's also faster to cache the lenghts in the xor patch, so that is done now. The performance is a little less impressive, but still good.

Here's the updated script:
```python
import functools
import timeit


def cmp_(a, b):
    if min(len(a), len(b)) == 1 and max(len(a), len(b)) > 1:
        return cmp(len(a), len(b))
    return None


def orig(a, b):
    a = len(a)
    b = len(b)
    if min(a, b) == 1 and max(a, b) > 1:
        return a < b
    return None


def orig2(a, b):
    if min(len(a), len(b)) == 1 and max(len(a), len(b)) > 1:
        return len(a) < len(b)
    return None


def xor(a, b):
    a = len(a)
    b = len(b)
    if (a == 1) ^ (b == 1):
        return a < b
    return None


def xor2(a, b):
    if (len(a) == 1) ^ (len(b) == 1):
        return len(a) < len(b)
    return None


print(timeit.timeit(functools.partial(cmp_, 'a', 'aaaaa')))
print(timeit.timeit(functools.partial(orig2, 'a', 'aaaaa')))
print(timeit.timeit(functools.partial(orig, 'a', 'aaaaa')))
print(timeit.timeit(functools.partial(xor2, 'a', 'aaaaa')))
print(timeit.timeit(functools.partial(xor, 'a', 'aaaaa')))
```
output:
```
0.615078926086
0.546674013138
0.464797973633
0.240245103836
0.194459915161
```

pazz:
this looks all really nice! well done!

Question: It seems we use some strange ordering for these widgets, which is almost like the lexicographical ordering on strings. Can't we simply use some (hopefully already optimised) buildin for lexicographic ordering directly?

tests/widgets/globals.py
```diff
@@ -0,0 +1,39 @@
...
+    def test_sort(self):
+        """Test sorting."""
+        with mock.patch(
+                'alot.widgets.globals.settings.get_tagstring_representation',
+                lambda t, _, __: {'translated': t, 'normal': None,
+                                  'focussed': None}):
+            expected = ['a', 'z', 'aa', 'bar', 'foo']
+            base = [globals_.TagWidget(x) for x in expected]
+            actual = list(sorted(base))
+            self.assertListEqual([g.translated for g in actual], expected)
```
dcbaker: Sure, I'll make the change. The ordering in the current version is just an artifact of how the version that used cmp_tag was written.

dcbaker:
So, what this does is sort first any strings that are one character alphabetically, then using python's standard string sorting (which is alphabetical).

The reason for this (as explained in the commit message introducing this
function a222a41076), is to group unicode special characters before
alphabetical ones (like the mail symbol). Since under standard python rules this list [u'z', u'ɤ', u'zz', u'a'] would sort as [u'a', u'z', u'zz', u'ɤ'].

Maybe there's a better approach to sorting those correctly, but I doubt it's built in.

pazz:
Quoting Dylan Baker (2017-01-27 18:55:45)
>    So, what this does is sort first any strings that are one character
>    alphabetically, then using python's standard string sorting (which is
>    alphabetical).
> 
>    The reason for this (as explained in the commit message introducing this
>    function a222a41076), is to group unicode special characters before
>    alphabetical ones (like the mail symbol). Since under standard python
>    rules this list [u'z', u'ɤ', u'zz', u'a'] would sort as [u'a', u'z',
>    u'zz', u'ɤ'].

Ah yes, I remember now. Thanks for the clarification though.
 
>    Maybe there's a better approach to sorting those correctly, but I doubt
>    it's built in.

OK, let's use your current version then. Thanks.
I will add this to milestone .6 and release .5 without this,
and merge this PR just afterwards.
That way we have some more time to test and react.
Cheers!

dcbaker:
> OK, let's use your current version then. Thanks. I will add this to milestone .6 and release .5 without this, and merge this PR just afterwards. That way we have some more time to test and react.

That sounds like a good plan. It's probably worth having a freeze on merges until we make the 0.5 release.

pazz:
@lucc: can you merge this once you're happy with the change?

lucc:
ok

dcbaker:
@lucc anything else that needs to happen for this to land?

PR #1013, opened 2017-01-27
helper: Use kibi and mibibytes instead of kilo and megabytes

dcbaker:
Technically a kilobyte (and it's derivatives like megabytes and
gigabytes) are defined as powers of 1000, while a kibibyte (and it's
derivatives like mibibytes and gibibytes) are defined as powers of 1024.

This patch fixes incorrect language and formatting in the humanize_size
function which defined a kilobyte as 1024.

See this wikipedia article for more information:
https://en.wikipedia.org/wiki/Kibibyte

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -511,8 +511,10 @@ def tag_cmp(a, b):
 
 def humanize_size(size):
     """Create a nice human readable representation of the given number
-    (understood as bytes) using the "K" and "M" suffixes to indicate kilo- and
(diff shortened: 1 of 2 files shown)


pazz:
Quoting Lucas Hoffmann (2017-01-28 14:18:26)
>    @lucc approved this pull request.
> 
>    [1]@pazz I would not add an new option for something that trivial (if
>    nobody steps up and really argues that we need it)

OK, fine with me.

Issue #528, opened 2012-10-06
mimeparser


pazz:
a few ideas:

recursively call yourself, passing on mime-tree:
- if msg is encrypted, decrypt and continue for subtree from cleartext,
- if signed, verify in thread and set result as attribute at root node.
- if multipart, recur for all subparts
- if content-type can be interpreted: call handler in thread and store result as attribute at root

Moreover,
- Make use of twisteds deferToThread to speed up external handlers
  http://twistedmatrix.com/documents/current/core/howto/gendefer.html#auto5
- wrap this into a `Message.get_mimetree` that returns a Deferred.
  MessageWidget can then update itself via a callback.

PR #1015, opened 2017-01-28
[RFC] ANSI-escape codes for colourful HTML mails

pazz:
This is a rebase of an old branch "0.3.6-feature-color-wip", which I will retire now, and ultimately passes on a proof-of-concept implementation of TextWidgets that interprets [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors).

The aim is to let the html renderer interpret colours from html mails and translate them to ANSI escapes. For instance, elinks can do this:

```
text/html; elinks -force-html -dump -dump-color-mode 3 -dump-charset utf8 -eval 'set document.codepage.assume = "%{charset}"' %s; copiousoutput
```

This feature can be enabled simply by configuring your mailcap as above and should not have any effect
if the browser does not produce these characters.
The boolean config setting can be used to disable interpretation of a subset of escape codes,
those which change the background colour of text (on top of say, changing foreground or character weights).

This is intended as a RFC and is not at all ready to go live!

I'll retire the related issue #612 because that PR is outdated.

### DIFF
--- a/alot/widgets/ansi.py
+++ b/alot/widgets/ansi.py
@@ -1,0 +1,115 @@
+# Copyright (C) 2011-2017  Patrick Totzke <patricktotzke@gmail.com>
+# This file is released under the GNU GPL, version 3 or a later revision.
+# For further details see the COPYING file
+
+import urwid
+
+
+class ANSIText(urwid.WidgetWrap):
+
+    """Selectable Text widget that interprets ANSI color codes"""
+
+    def __init__(self, txt,
+                 default_attr=None,
+                 default_attr_focus=None,
+                 ansi_background=True, **kwds):
+        ct, focus_map = parse_escapes_to_urwid(txt, default_attr,
+                                               default_attr_focus,
+                                               ansi_background)
+        t = urwid.Text(ct, **kwds)
+        attr_map = {default_attr.background: ''}
+        w = urwid.AttrMap(t, attr_map, focus_map)
+        urwid.WidgetWrap.__init__(self, w)
+
+    def selectable(self):
+        return True
+
+    def keypress(self, size, key):
+        return key
+
+ECODES = {
+    '1': {'bold': True},
+    '4': {'underline': True},
+    '7': {'standout': True},
+    '30': {'fg': 'black'},
+    '31': {'fg': 'dark red'},
+    '32': {'fg': 'dark green'},
+    '33': {'fg': 'brown'},
+    '34': {'fg': 'dark blue'},
+    '35': {'fg': 'dark magenta'},
+    '36': {'fg': 'dark cyan'},
+    '37': {'fg': 'light gray'},
+    '40': {'bg': 'black'},
+    '41': {'bg': 'dark red'},
+    '42': {'bg': 'dark green'},
+    '43': {'bg': 'brown'},
+    '44': {'bg': 'dark blue'},
+    '45': {'bg': 'dark magenta'},
+    '46': {'bg': 'dark cyan'},
+    '47': {'bg': 'light gray'},
+}
+
+
+def parse_escapes_to_urwid(text, default_attr=None, default_attr_focus=None,
+                           parse_background=True):
+    """This function converts a text with ANSI escape for terminal
+    attributes and returns a list containing each part of text and its
+    corresponding Urwid Attributes object, it also returns a dictionary which
+    maps all attributes applied here to focused attribute.
+    """
+
+    text = text.split("\033[")
+    urwid_text = [text[0]]
+    urwid_focus = {None: default_attr_focus}
+
+    # Escapes are cumulative so we always keep previous values until it's
+    # changed by another escape.
+    attr = dict(fg=default_attr._foreground_color, bg=default_attr.background,
+                bold=default_attr.bold, underline=default_attr.underline,
+                standout=default_attr.underline)
+    for part in text[1:]:
+        esc_code, esc_substr = part.split('m', 1)
+        esc_code = esc_code.split(';')
+
+        if not esc_code:
+            attr.update(fg=default_attr._foreground_color,
+                        bg=default_attr.background, bold=default_attr.bold,
+                        underline=default_attr.underline,
+                        standout=default_attr.underline)
+        else:
+            i = 0
+            while i < len(esc_code):
+                code = esc_code[i]
+                if code is 0:
+                    attr.update({'bold': default_attr.bold,
+                                 'underline': default_attr.underline,
+                                 'standout': default_attr.standout})
+                if code in ECODES:
+                    attr.update(ECODES[code])
+                # 256 codes
+                elif code == '38':
+                    attr.update(fg='h' + esc_code[i+2])
+                    i += 2
+                elif code == '48':
+                    attr.update(bg='h'+esc_code[i+2])
+                    i += 2
+                i += 1
+
+        # If there is no string in esc_substr we skip it, the above
+        # attributes will accumulate to the next escapes.
+        if esc_substr:
+            # Construct Urwid attributes
+            urwid_fg = attr['fg']
+            urwid_bg = default_attr.background
+            if attr['bold']:
+                urwid_fg += ',bold'
+            if attr['underline']:
+                urwid_fg += ',underline'
+            if attr['standout']:
+                urwid_fg += ',standout'
+            if parse_background:
+                urwid_bg = attr['bg']
+            urwid_attr = urwid.AttrSpec(urwid_fg, urwid_bg)
+            urwid_focus[urwid_attr] = default_attr_focus
+            urwid_text.append((urwid_attr, esc_substr))
+    return urwid_text, urwid_focus
--- a/alot/widgets/ansi.py
+++ b/alot/widgets/ansi.py
@@ -17,7 +17,7 @@ class ANSIText(urwid.WidgetWrap):
                                                default_attr_focus,
                                                ansi_background)
         t = urwid.Text(ct, **kwds)
-        attr_map = {default_attr.background: ''}
+        attr_map = { default_attr.background: ''}
         w = urwid.AttrMap(t, attr_map, focus_map)
         urwid.WidgetWrap.__init__(self, w)
 
--- a/alot/widgets/ansi.py
+++ b/alot/widgets/ansi.py
@@ -17,7 +17,7 @@ class ANSIText(urwid.WidgetWrap):
                                                default_attr_focus,
                                                ansi_background)
         t = urwid.Text(ct, **kwds)
-        attr_map = { default_attr.background: ''}
+        attr_map = {default_attr.background: ''}
         w = urwid.AttrMap(t, attr_map, focus_map)
         urwid.WidgetWrap.__init__(self, w)
 
@@ -26,6 +26,7 @@ class ANSIText(urwid.WidgetWrap):
 
     def keypress(self, size, key):
         return key
+
 
 ECODES = {
     '1': {'bold': True},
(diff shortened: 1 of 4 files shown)


alot/widgets/ansi.py
```diff
@@ -0,0 +1,113 @@
...
+                                               ansi_background)
+        t = urwid.Text(ct, **kwds)
+        attr_map = { default_attr.background: ''}
+        w = urwid.AttrMap(t, attr_map, focus_map)
+        urwid.WidgetWrap.__init__(self, w)
+
+    def selectable(self):
+        return True
+
+    def keypress(self, size, key):
```
pazz: FTW: this is implementing an urwid interface.

alot/widgets/ansi.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def parse_escapes_to_urwid(text, default_attr=None, default_attr_focus=None,
+                           parse_background=True):
+    """This function converts a text with ANSI escape for terminal
+    attributes and returns a list containing each part of text and its
+    corresponding Urwid Attributes object, it also returns a dictionary which
+    maps all attributes applied here to focused attribute.
+    """
+    ECODES = {
```
dcbaker: For performance reasons I would not put this dict in the function, if we put it at the module level it will be initialized once, inside the function it will be re-initialized each time the function is called.

alot/widgets/ansi.py
```diff
@@ -0,0 +1,113 @@
...
+    # Escapes are cumulative so we always keep previous values until it's
+    # changed by another escape.
+    attr = dict(fg=default_attr._foreground_color, bg=default_attr.background,
+                bold=default_attr.bold, underline=default_attr.underline,
+                standout=default_attr.underline)
+    for part in text[1:]:
+        esc_code, esc_substr = part.split('m', 1)
+        esc_code = esc_code.split(';')
+
+        if len(esc_code) == 0:
```
dcbaker: `if not esc_code` avoids uneeded calls.

alot/widgets/ansi.py
```diff
@@ -0,0 +1,113 @@
...
+                        standout=default_attr.underline)
+        else:
+            i = 0
+            while i < len(esc_code):
+                code = esc_code[i]
+                if code in ECODES:
+                    attr.update(ECODES[code])
+                # 256 codes
+                elif code == '38':
+                    attr.update(fg='h'+esc_code[i+2])
```
dcbaker: let's be consistant and use spaces around the `+` operators.

alot/widgets/ansi.py
```diff
@@ -0,0 +1,113 @@
+# Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
```
dcbaker: you should probably update this to 2017

alot/widgets/ansi.py
```diff
@@ -0,0 +1,113 @@
...
+                    attr.update(fg='h'+esc_code[i+2])
+                    i += 2
+                elif code == '48':
+                    attr.update(bg='h'+esc_code[i+2])
+                    i += 2
+                i += 1
+
+        # If there is no string in esc_substr we skip it, the above
+        # attributes will accumulate to the next escapes.
+        if esc_substr != '':
```
dcbaker: just `if esc_substr:` is good here.

alot/widgets/ansi.py
```diff
@@ -0,0 +1,113 @@
...
+            # Construct Urwid attributes
+            urwid_fg = attr['fg']
+            urwid_bg = default_attr.background
+            if attr['bold']:
+                urwid_fg += ',bold'
```
dcbaker: This kind of str concatenation isn't recomended by the python devs for perforamance reasons. There's some clever hacks in CPython to make this fast, but it's not guaranteed to be. The usual idiom is to make a list and then return a `','.join(list_)`.

alot/widgets/ansi.py
pazz: yes, only that the value for '0' should be set to the function parameters.
I could define this dict at module level, then `.copy()` it and redefine the value for '0' in the function. But that's equally inefficient I suppose.

alot/widgets/ansi.py
pazz: ok

alot/widgets/ansi.py
pazz: really? even for "lists" of length 2?

alot/widgets/ansi.py
dcbaker: hmmm, that is an interesting problem. I guess we could special case '0' like we do a couple of other values?

alot/widgets/ansi.py
dcbaker: You may be right, for len <= 3 concatenation may be fine.

I know that i've run into cases in pypy where string concatenation is *significantly* slower than it is in cpython (granted with several hundred concatenation operations), and there is a comment in the documentation somewhere that says string concatenation isn't guaranteed to be efficient/fast.

alot/widgets/ansi.py
dcbaker: Whatever you decide to do here is fine. If it becomes a perforance bottleneck we can figure it out later.

pazz: ok, I hardcoded 0 now.

pazz:
I've just tested that this actually works :) it does, and e.g. github mails look kind of cool.
Now I just need a html renderer that enriches the plaintext and does not enforce a fixed line width.

I suppose one could also use some colouring scripts from `ranger`, to colorize text attachments like patches..

dcbaker:
I've never seen ranger, and it's remarkably hard to goolge for. Is it something like pygments?

pazz:
sorry, here you go: http://ranger.nongnu.org/
(It is a console file manager)
Maybe more related: http://www.andre-simon.de/doku/highlight/en/highlight.php#ch1

dcbaker:
What's the status of this branch?

pazz:
This seems to work as expected. But this being said, I'm not using it myself simply because it's difficult to find the right input:
In this branch, all ANSI codes are interpreted if they appear in what alot computes as "body text" of a mail. One application is html parts, which get piped through an interpreter that enriches the text with ANSI codes. I've only found one (elinks) with the mailcap entry below

```
text/html; elinks -force-html -dump -dump-color-mode 3 -dump-charset utf8 -eval 'set document.codepage.assume = "%{charset}"' %s; copiousoutput
```

... which does that, but elinks will hardcode linebreaks at a given line width. I find this ugly, so I don't use it.

Nevertheless, the widget could later be used nicely to inline display some text-based attachments, or other text-interpreted stuff (pdfs?) directly in alot. This  relates to #528 and #894, which are low prio unfortunately. Such a feature should also be via dedicated toggle commands and config settings for the default behaviour I think.

At the very least, ANSI code interpretation as is in this PR should get a disable-switch in the config.
So: This could be a rather cheap new feature merge. The question is just if it's useful to anyone..

lucc:
@pazz what do you use to not render html in fixed width?

I'm also not if this is an essential feature. *If* you are looking for arguments you can look at mutt. They have the ability to highlight text in the body via regex and if I remember correctly I saw a discussion about security once. It was about the gpg notification being intermixed and highlighted with the displayed body text so if the sender adds escape sequences to the mail they could fool you into beliving that it was signed/encrypted.

pazz:
Quoting Lucas Hoffmann (2017-06-21 09:13:00)
> @pazz what do you use to not render html in fixed width?

I use this in my ~/.mailcap

```
text/html; /usr/bin/w3m -I %{charset} -dump -T text/html %s;
copiousoutput; description=HTML Text; nametemplate=%s.html
```

lypanov:
Any further interest in this feature? I'd love to see alot support https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

Background: I have tried to move to an elinks+url_scan workflow for "clicking" on links in emails but I'm still having issues getting the numbering to be accurate between the two of them no matter how many hacky regexps I apply.

pazz:
Yes, I'm definitely still interested in this feature. Perhaps we should simply rebase and merge it...
There is also this branch, which may be related: https://github.com/pazz/alot/pull/1441

I'm not sure that your feature request is related at all @lypanov : 
This PR would introduce a new text widget that interprets escape codes for colourings, i.e., removes them from the text and wraps the text in urwid-styled subwidgets.
It would not interpret any other non-printable escape codes such as the url one. Therefore, the displayed text would simply still contain that code. If your terminal interprets these, it should work independently from alot. I'm surprized it doesn't for you.

lypanov:
Sorry for the delay in response. I've finally got gnome-terminal running locally to test this but (maybe due to confusion on my end) am unable to get the direct pass through working as expected.

Specifically, I have a .sh which wraps the printf test line (from https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda). I've verified that the .sh works from within the terminal, however, when used in alot via the the following mailcap entry:
  `text/html; /home/lypanov/.local/bin/urltest.sh; copiousoutput`

I see the following within alot when navigating to an email with HTML content:
  `?]8;;http://example.com?\This is a link?]8;;?\`

Is there something magical extra needed in the mailcap entry?

Thank you!

pazz:
I'm not sure how exactly these escape codes work, meaning where they are interpreted, but was under the impression that the terminal translates them whenever they occur on screen. So there should be no difference between the shell and urwid/alot.

Perhaps it's worth asking on the urwid list.

lypanov:
Dug a lot deeper on this this morning. URWIDs support for colors looks like it could be extended to support hyperlinks also as it's not using something like curses but rather a raw escape sequence based method for rendering. Supporting the OSC 8 hyperlink extension would just be a case of adding this support to URWID, and then extending your color patch.

I'll try this branch and look at the next steps as soon as I have a working terminal emulator (right now even that has a bunch of issues).

pazz:
@lypanov: maybe, but you'd also need to find a html2txt renderer that produces those sequences, unless they are already part of a plaintext email. I'm not sure if elinks as above already does that.

In any case, the patch still looks good to me and should be harmless if one doesn't use escape sequences.
@lucc: should we simply merge this into master?

lypanov:
> @lypanov: maybe, but you'd also need to find a html2txt renderer that produces those sequences, unless they are already part of a plaintext email. I'm not sure if elinks as above already does that.

Finally on a "coding vacation" and was able to make a postprocess hack in ruby for elinks to merge it's link numbering output back into the text in the hyperlink format needed. Works pretty much. Would like to look into the URWID / alot side of things next. Any progress on this ticket?

lypanov:
Got this working locally (in a very messy form for now). Have patches to urwid and alot (~20 lines) on top of this branch. I'll harden it for a few days and try to get the mess on github in some form.

lypanov:
Just in case it's useful here is the conflict resolved branch without my changes: https://github.com/lypanov/alot/tree/feature-ANSI-colour-interpretation

pazz:
@lypanov: I don't understand. Your last commit is a merge and is not easy to see what the diff to this branch is. Do you want to add to this branch? Then please rebase.
I'm tempted to merge this branch into master as is.

lypanov:
@pazz Sorry if the message wasn't clear. The branch above is just the merging of master into your branch + resolving of the conflicts. Still not 100% sure of a rebase workflow sorry. (Ah in the meantime see that you've anyway resolved them :) )

My changes aren't quite ready for a PR (and a PR is worthless until my urwid branch is merged anyway). See https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#gistcomment-3132951 for the various branches / scripts.

pazz:
ok cool. thanks for working on this :)

lypanov:
My understanding is that you're using w3m right @pazz? I might look into getting an option into w3m rather than using a postprocess mechanism. At least w3m looks relatively active. elinks / links are far from.

pazz:
I'm honestly not using this feature at all at the moment and only have tested it as described above.

PR #1019, opened 2017-01-31
more unittests

dcbaker:
This is a few more unittests (and a couple of cleanups mixed in). I don't think that there's anything too crazy in here.

### DIFF
--- a/tests/commands/envelope_tests.py
+++ b/tests/commands/envelope_tests.py
@@ -1,0 +1,107 @@
+# encoding=utf-8
+# Copyright © 2017 Dylan Baker
+
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Tests for the alot.commands.envelope module."""
+
+from __future__ import absolute_import
+import os
+import contextlib
+import shutil
+import tempfile
+import unittest
+
+import mock
+
+from alot.commands import envelope
+
+
+@contextlib.contextmanager
+def temporary_directory(suffix='', prefix='', dir=None):
+    """Python3 interface implementation.
+
+    Python3 provides a class that can be used as a context manager, which
+    creates a temporary directory and removes it when the context manager
+    exits. This function emulates enough of the interface of
+    TemporaryDirectory, for this module to use, and is designed as a drop in
+    replacement that can be replaced after the python3 port.
+
+    The only user visible difference is that this does not implement the
+    cleanup method that TemporaryDirectory does.
+    """
+    directory = tempfile.mkdtemp(suffix=suffix, prefix=prefix, dir=dir)
+    yield directory
+    shutil.rmtree(directory)
+
+
+class TestAttachCommand(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


tests/commands/envelope_tests.py
```diff
@@ -0,0 +1,103 @@
...
+            testfile = os.path.join(d, 'foo')
+            with open(testfile, 'w') as f:
+                f.write('foo')
+
+            cmd = envelope.AttachCommand(path=testfile)
+            cmd.apply(ui)
+        ui.current_buffer.envelope.attach.assert_called_with(testfile)
+
+    def test_user(self):
+        """A test for an existing single path prefaced with ~/."""
```
lucc: I don't understand where the path is that is prefixed with ~/ ? Shouldn't you do something like `AttachCommand(path="~/foo")` below?

tests/commands/envelope_tests.py
```diff
@@ -0,0 +1,103 @@
...
+import tempfile
+import unittest
+
+import mock
+
+from alot.commands import envelope
+
+
+@contextlib.contextmanager
+def temporary_directory(suffix='', prefix='', dir=None):
```
dcbaker: Yeah, i wasn't really sure about how to handle this. This code is basically emulation for `tempfile.TemporaryDirectory` which is avilable in python 3.2+, so it's hopefully going away pretty soon.

tests/commands/envelope_tests.py
dcbaker: I removed a mock patch from that test that I thought I needed when I started, then realized I didn't. I think I somehow backed out the change to make testfile = '~/foo' when I did that. I'll fix that.

lucc: thanks

tests/commands/envelope_tests.py
```diff
@@ -0,0 +1,107 @@
...
```
pazz: should I disable this codacity-bot target? It seems reasonable to not overide buildins, but if its a "standard" pattern or interface you are implementing, then these complaints are just annoying..

dcbaker: I'm not sure. Sometimes it does spot useful things, but I can't find any documentation on how to tell it "please ignore this", pylint and pyflakes both have a comment syntax that can be used to tell it "I know what I'm doing", I would think codacy does too?

pazz: yes, it can enable/disable various patterns.
I'm not sure if you as a collaborator can also edit those already:
https://www.codacy.com/app/patricktotzke/alot/patterns/list

dcbaker: nope, permission denied

pazz: OK fair enough. Anyways, let me know if i should supress anything in particular.

dcbaker: Is it possible to supress certain things on just the tests? unittest is going to generate a lot of "method could be function" errors by it's nature.

pazz: I think one can only ignore subdirectories completely.

pazz: No, actually: you can give it a `pylintrc` config file and it will pick it up. I haven't used pylint explicitly so far, but if you can cook up something quickly I'd instruct it to use that.
The only other tool that's currently enabled is  [Bandid](https://github.com/codacy/bandit), also configurable via config file.

dcbaker: does it honor pylint comment directives? It's pretty easy to slap `#pylint: disable=...` at the top of the modules.

pazz:
Quoting Dylan Baker (2017-02-02 17:23:54)
> does it honor pylint comment directives? It's pretty easy to slap #pylint:
> disable=... at the top of the modules.

We can cehck, I'm not sure as the docs don't say so explicitly.
But they do run pylint, so unless their configs (which we can change)
disable this feature, it should work.

The easiest would be to make a PR here that introduces such comments and
see if how the bot reviews..

dcbaker:
Okay, that seems easy enough :)

PR #1022, opened 2017-02-02
tests: Instruct pylint to ignore a large swath of warnings

dcbaker:
There are a number of things pylint warns on that absolutely make sense
to fix in production code, but for unittests they either don't matter
(like naming variables "foo"), can't be fixed (TestCase methods that
don't use self because they use a mock assert), or the descriptive names
violate PEP8. These are annoying and create noise, so tell pylint to
ignore them.

### DIFF
--- a/tests/commands/envelope_tests.py
+++ b/tests/commands/envelope_tests.py
@@ -27,9 +27,14 @@ import mock
 
 from alot.commands import envelope
 
+# When using an assert from a mock a TestCase method might not use self. That's
+# okay.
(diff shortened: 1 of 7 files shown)


dcbaker:
This is a test branch to see if codacy uses the pylint comment directives.

pazz:
i take it the absence of codacity-bot complaints means that it's working. Should we go ahead with this?

dcbaker:
Updated to merge cleanly with master.

PR #1023, opened 2017-02-02
alot/db/utils: Revert use of iterkeys on Message

dcbaker:
Message doesn't have an iterkeys method, just a keys method.

### DIFF
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -251,7 +251,7 @@ def extract_headers(mail, headers=None):
     """
     headertext = u''
     if headers is None:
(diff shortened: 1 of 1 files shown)


pazz:
yes, but why hasn't this blown up before? it looks as if this function is actually used quite a bit.

pazz:
ah! it is unly used in threadbuffers PipetoCommand and in `dbm.message.Message.get_headers_string`, which itself is never used.

dcbaker:
yup, I noticed it using PipetoCommand.

PR #1028, opened 2017-02-08
Convert gpg uid choices into a list

meskio:
reversed doesn't work with generators

### DIFF
--- a/alot/commands/utils.py
+++ b/alot/commands/utils.py
@@ -75,8 +75,8 @@ def _get_keys(ui, encrypt_keyids, block_error=False, signed_only=False):
                                  signed_only=signed_only)
         except GPGProblem as e:
             if e.code == GPGCode.AMBIGUOUS_NAME:
-                tmp_choices = (k.uids[0].uid for k in
-                               crypto.list_keys(hint=keyid))
+                tmp_choices = [k.uids[0].uid for k in
+                               crypto.list_keys(hint=keyid)]
                 choices = {str(i): t for i, t in
                            enumerate(reversed(tmp_choices), 1)}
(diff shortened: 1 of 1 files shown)


pazz:
We need more info here: Does this solve some known issue? Or a new one?

meskio:
I don't think this issue has being reported, AFAIK was introduced by #940, so is pretty recent.

When there is more than one key for a certain uid and you ask alot to encrypt the email the current code from master breaks with:
```
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1445, in unwindGenerator
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1299, in _inlineCallbacks
    result = g.send(result)
  File "/home/user/.local/lib/python2.7/site-packages/alot-0.5.1-py2.7.egg/alot/commands/utils.py", line 44, in set_encrypt
    signed_only=signed_only)
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1445, in unwindGenerator
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1299, in _inlineCallbacks
    result = g.send(result)
  File "/home/user/.local/lib/python2.7/site-packages/alot-0.5.1-py2.7.egg/alot/commands/utils.py", line 81, in _get_keys
    enumerate(reversed(tmp_choices), 1)}
exceptions.TypeError: argument to reversed() must be a sequence
```
After some research I found out that `reversed`, used just one line below my change, won't accept a generator, it needs an iterator. I'm converting the tuple comprehension into list comprehension to fix the issue.

Is it more clear now?

pazz:
> Is it more clear now?

very. Thanks for the bughunt and the fix!

PR #1031, opened 2017-02-09
MANIFEST.in: fix include of tests

dcbaker:
This fixes a couple of warnings when building.


dcbaker:
@zenhack, you seem to konw more about setuptools than the rest of use ;), does this look right to you?

zenhack:
Ack, I screwed up the other patch. You can actually just delete the last two lines of the file and it will do the right thing (with no warnings). the recursive-include actually isn't even needed since it will pull in *.py by default anyway.

PR #1032, opened 2017-02-09
Set version back to dev

dcbaker:
It appears after pushing the 0.5.1 release we forgot to set it back to dev.

### DIFF
--- a/alot/__init__.py
+++ b/alot/__init__.py
@@ -1,5 +1,5 @@
 __productname__ = 'alot'
-__version__ = '0.5.1'
+__version__ = '0.6.0dev'
 __copyright__ = "Copyright (C) 2012-17 Patrick Totzke"
 __author__ = "Patrick Totzke"
 __author_email__ = "patricktotzke@gmail.com"


lucc:
I understand that we are not on 0.5.1 any longer and want to communicate this and I also understand that we want to indicate which release we are currently working on.

But there are two things I did never understand about this "dev" suffix. Why are we on 0.6.0dev and not on 0.5.2dev or even 1.0.0dev?

And if I install or make a tarball from 0.6.0dev today and tomorrow we release 0.5.2 and then switch back to 0.6.0dev how do I still see which tarball or version is newer?  (In gerneral for any x and y, is 0.x.y < 0.(x+1).0dev or > ?).  

Or am I thinking about these things the wrong way and we are already developing in two branches: The 0.5.x maintainance branch, that will recive bug fixes and the 0.6.0dev branch that should get all the bugfixes and potentially additional stuff?

pazz:
Quoting Lucas Hoffmann (2017-02-09 21:30:38)
> I understand that we are not on 0.5.1 any longer and want to communicate this
> and I also understand that we want to indicate which release we are currently
> working on.
> 
> But there are two things I did never understand about this "dev" suffix. Why
> are we on 0.6.0dev and not on 0.5.2dev or even 1.0.0dev?


Actually, I was with you on that prior to 0.4:
I've always used the 'dev' suffix to mean "development after the
indicated release".

PEP 440 however is with dcbaker here:
https://www.python.org/dev/peps/pep-0440/#summary-of-permitted-suffixes-and-relative-ordering
 
> And if I install or make a tarball from 0.6.0dev today and tomorrow we release
> 0.5.2 and then switch back to 0.6.0dev how do I still see which tarball or
> version is newer? (In gerneral for any x and y, is 0.x.y < 0.(x+1).0dev or >
> ?).

Good question. I believe those branches are unrelated then..

Anyway, I'm OK with switching to the PEP semantics. We are planning to
make a .6 release soon anyway aren't we?

dcbaker:
I don't really care which way we do it, I've always worked on projects where `1.0-dev` or `1.0-devel` was the development version that would be 1.0, but really I just think it's useful to have a `dev` suffix on development snapshots.

PR #1034, opened 2017-02-13
alot.rc.spec: Fix typo in documentation

dcbaker:
It currently lists "terminal_command" in relation to "editor_spawn" but
the config option is actually "terminal_cmd".


pazz:
in fact, it should be nice to refer (in sphinx markup) to the other config option here..

dcbaker:
I can do that. Hold on, I'll send a v2

pazz:
thanks

PR #1039, opened 2017-02-25
Remove Java style getters

lucc:
There are some more in the code but I did not know which belong to some urwid interface that we implement and which could be removed.

If you agree I would say we should note which of our methods overwrite/implement some library interface in the docstring of the method. (Another job on the incremental TODO list.)

### DIFF
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -308,7 +308,7 @@ class SearchBuffer(Buffer):
             self.body.set_focus(0)
         elif self.result_count < 200 or self.sort_order not in self._REVERSE:
             self.consume_pipe()
-            num_lines = len(self.threadlist.get_lines())
+            num_lines = len(self.threadlist.lines)
             self.body.set_focus(num_lines - 1)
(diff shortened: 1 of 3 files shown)


pazz:
i like it a lot, this has been bothering my inner pedant for a while now! Is there a standard way to document all properties of a class? I suspect the latter two will not automatically be identified as properties by sphinx because you don't explicitly declare them as such with a decorator?

lucc:
I don't know of a standard way but we can use the class docstring.

lucc:
Should I add a commit to list the exposed variables in the class docstrings?

pazz:
I think that'd be helpful thanks.
Also, If you find other candidates, don't hesitate to mark them in comments as such. I will have a browse through it later regarding urwid interfaces..

pazz:
thanks.

lucc:
I am not sure if it is done. There are still some left but we have to check them against the urdiw interface (and mabe add a comment).

PR #1040, opened 2017-03-02
Revert parts of #1039

lucc:
I only noticed after the PR was merged that it seems to introduce a bug.  If I open any buffer I get the error message `No focus_position, ListBox is empty`.  If I revert these two commits the error is gone.  I assume that I deleted some of the interface of urwid that we need to implement and I was sloppy enough to not notice.

I am very sorry for the noise and inconvenience (at least it seems that nobody has used the code yet or at least nobody has reported it).  Please test the bug in master and the fix in this PR before you approve this PR.

### DIFF
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -282,7 +282,7 @@ class SearchBuffer(Buffer):
         returns curently focussed :class:`alot.widgets.ThreadlineWidget`
         from the result list.
         """
-        threadlinewidget, _ = self.threadlist.focus
+        threadlinewidget, _ = self.threadlist.get_focus()
         return threadlinewidget
 
     def get_selected_thread(self):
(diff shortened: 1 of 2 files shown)


pazz:
thanks for noticing and sending a pr! is this worth adding tests?

lucc:
It certainly is, I was even wondering if there is a way to check that we implement all necessary methods for the urwid interfaces we use but I guess the way to go here would be to use the abc metaclass upstream.

pazz:
strange: http://urwid.org/reference/widget.html#urwid.Widget.focus_position mentions focus_position as a *property*.. I think what happens is that de default implementation of its getter will call `.get_focus` internally. So a nic(er) solution would be to actually implement `focus_position`, not `.focus`.

pazz:
see also the discussion in #1043; thanks

PR #1050, opened 2017-05-29
Don't install test suite

aidecoe:
See
<http://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=find_packages#using-find-packages>
for example patterns.


pazz:
I'm confused as to what this patch actually does. As far as I understand, the tests directory is already excluded isn't it?

aidecoe:
Ah, indeed. Sorry. I was checking that against 0.5.1, but it's fixed in master, indeed.

Issue #332, opened 2012-02-16
alot and notmuch's exclude_tags


quite:
alot doesn't care about exclude_tags in .notmuch-config, right?

I guess this is intentional, and that the default search 
"tag:inbox AND NOT tag:killed" reflects a similar functionality?
(but only for inbox)

PR #1056, opened 2017-06-02
Exclude search.exclude_tags from searches

### LINKED ISSUE #707: searches match excluded tags
In my .notmuch-config I have these two lines
[search]
exclude_tags=deleted;spam;

I tagged some messages as "deleted", but a generic search (*) still shows those messages. They do not show when using the emacs frontend or `notmuch search \*`

kyrias:
Fixes #707 and #332.

Signed-off-by: Johannes Löthberg <johannes@kyriasis.com>

### DIFF
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -260,9 +260,14 @@ class SearchBuffer(Buffer):
         else:
             order = self.sort_order
 
+        exclude_tags = settings.get_notmuch_setting('search', 'exclude_tags')
+        if exclude_tags:
+            exclude_tags = [t for t in exclude_tags.split(';') if t]
+
         try:
             self.pipe, self.proc = self.dbman.get_threads(self.querystring,
-                                                          order)
+                                                          order,
+                                                          exclude_tags)
         except NotmuchError:
             self.ui.notify('malformed query string: %s' % self.querystring,
                            'error')
--- a/alot/db/manager.py
+++ b/alot/db/manager.py
@@ -368,7 +368,7 @@ class DBManager(object):
         sender.close()
         return receiver, process
 
-    def get_threads(self, querystring, sort='newest_first'):
+    def get_threads(self, querystring, sort='newest_first', exclude_tags=None):
         """
         asynchronously look up thread ids matching `querystring`.
 
@@ -377,6 +377,8 @@ class DBManager(object):
         :param sort: Sort order. one of ['oldest_first', 'newest_first',
                      'message_id', 'unsorted']
         :type query: str
+        :param exclude_tags: Tags to exclude by default unless included in the search
+        :type exclude_tags: list of str
         :returns: a pipe together with the process that asynchronously
                   writes to it.
         :rtype: (:class:`multiprocessing.Pipe`,
@@ -385,6 +387,9 @@ class DBManager(object):
         assert sort in self._sort_orders
         q = self.query(querystring)
         q.set_sort(self._sort_orders[sort])
+        if exclude_tags:
+            for tag in exclude_tags:
+                q.exclude_tag(tag)
         return self.async(q.search_threads, (lambda a: a.get_thread_id()))
 
     def query(self, querystring):
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -89,7 +89,8 @@ class ExitCommand(Command):
     (['query'], {'nargs': argparse.REMAINDER, 'help': 'search string'})])
 class SearchCommand(Command):
 
-    """open a new search buffer"""
+    """open a new search buffer. Search obeys the notmuch
+    :ref:`search.exclude_tags <search.exclude_tags>` setting."""
     repeatable = True
 
     def __init__(self, query, sort=None, **kwargs):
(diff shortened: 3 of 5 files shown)


kyrias:
I'm not sure whether it'd be better to do it this way, or have an alot config option for it instead, with a format that we can control.

kyrias:
hmr, just realized this doesn't work properly, causing `alot.db.errors.NonexistantObjectError` if a thread got excluded and it tries to get it anyway, hm.

kyrias:
Implemented differently now, and it seems to be working properly.

lucc:
It seems to work as intended.  But I think in order to get this merged you need to add docs and if possible tests.

pazz:
there​ is not much to be documented right? he does not introduce any new config options but  just reads the one for the notmuch cli (which is also used in Emacs afaik).
Tests could be useful though.

lucc:
But we should mention that we read this value from the notmuch config otherwise people might get confused why they don't see some mail.

pazz:
Quoting Lucas Hoffmann (2017-06-03 09:39:10)
> But we should mention that we read this value from the notmuch config otherwise
> people might get confused why they don't see some mail.

agreed.

kyrias:
Do you want it documented in the `search` command documentation section, or somewhere else?  And I'll look into writing some tests.

lucc:
Docs look good. If you want to be extra verbose you can add that the "notmuch options" are read from the notmuch config file (its path is already explained in the section 4.5, you could also link to that if you want).

Are there any tests we can add for this?

lucc:
@pazz do we need to add this to notmuc.rc.spec?

kyrias:
Problem with actually testing the functionality is that the test suite will need the ability to set up a notmuch database and fill it with data first, hm.  The only thing I think we could test right now would be if I split the getting of the setting into a new function and then write out a notmuch config and then test only that function...

lucc:
I thought about which test might be related to this and came up with #1061.

kyrias:
Actually, I wonder if I should actually add support to the settings manager for getting a semicolon separated value and automatically turning that into a list.  Feels like that would be a bit cleaner than doing it in the rest of the code.

alot/db/manager.py
```diff
@@ -368,7 +368,7 @@ def threaded_reader(prefix, fd):
         sender.close()
         return receiver, process
 
-    def get_threads(self, querystring, sort='newest_first'):
+    def get_threads(self, querystring, sort='newest_first', exclude_tags=None):
```
lucc: I just noticed that this needs a docstring update

alot/db/manager.py
```diff
@@ -385,6 +385,9 @@ def get_threads(self, querystring, sort='newest_first'):
         assert sort in self._sort_orders
         q = self.query(querystring)
         q.set_sort(self._sort_orders[sort])
+        if exclude_tags:
```
lucc: If you change the default arg to `exclude_tags=()` (not `[]`!!) then you could leave out the ` if` here.

lucc: Oh, but then we also need to use `fallback=()` in the `get_notmuch_settings` call. Not sure which is nicer ...

kyrias: Hmm, for the default it shouldn't make a difference whether it's an empty list or an empty tuple though?

But mmm... I generally opt for explicit checks, but I can do it either way.  @pazz any opinions on this?

alot/db/manager.py
kyrias: Fixed this in the following commit, can rebase it once it's ready for merging.

alot/db/manager.py
lucc: If you like the explicit if better lets leave it.

About the empty list, try this:
```python
def f(x=[]):
    x.append(1)
    return x
print(f())
print(f())
```
It is generally advised that one never uses mutable types for default params.

kyrias: Eh, I don't really see the point when you're not going to modify the argument, and I personally dislike using a type other than the expected type as the default one more, since that looks rather misleading.

dcbaker: I tend to agree. This is why the `x = x or []` pattern is idiomatic. I like the change the way it is actually.

Rant about lists and tuples ahead ;)
A tuple and a list are not the same type and they are for different things. Tuples are heterogeneous, that is, the behavior of the elements only matters by position. Something like `[(1, True), (0, False), (7.2, True)]` is good, but `[(1, True), ('x', 7)]` is not. Lists are homogeneous, it's type (or behavior since python is duck typed) that matters, not position. `[1, 1.5, 7]` is perfectly valid, but `['a', 1, False]` is not.

alot/db/manager.py
dcbaker: Please do squash the two patches together.

alot/buffers.py
```diff
@@ -261,8 +261,13 @@ def rebuild(self, reverse=False):
...
         try:
+            exclude_tags = settings.get_notmuch_setting('search', 'exclude_tags')
+            if exclude_tags:
+                exclude_tags = [t for t in exclude_tags.split(';') if t]
+
             self.pipe, self.proc = self.dbman.get_threads(self.querystring,
-                                                          order)
+                                                          order,
+                                                          exclude_tags)
         except NotmuchError:
```
dcbaker: Which command are we expecting to raise a `NotmuchError`? Does `settings.get_notmuch_setting` allow them to propagate up? If not then the `exclude_tags` bit out of the try/except.

kyrias:
Moved the `get_notmuch_setting` call out of the `try`, and squashed the commits.

PR #1061, opened 2017-06-07
add some first tests for settings.manager.SettingsManager

lucc:


lucc:
The string would be a filename so I would have to write out the file contents to that filename first which is a lot slower I think and storing the file on disk/in git permanently makes it a "remote dependency" to the test (or however you call that). Both are things I wanted to avoid.

pazz:
Quoting Lucas Hoffmann (2017-06-07 22:34:50)
> The string would be a filename so I would have to write out the file contents
> to that filename first which is a lot slower I think and storing the file on
> disk/in git permanently makes it a "remote dependency" to the test (or however
> you call that). Both are things I wanted to avoid.

Oh you're right! I assumed that the string contains the contents of the
file, not its path. In that case I fully agree with you.
cheers,
P

PR #1063, opened 2017-06-08
tests for addressbook classes

lucc:
(only some first tests)

### DIFF
--- a/alot/addressbook/external.py
+++ b/alot/addressbook/external.py
@@ -68,7 +68,7 @@ class ExternalAddressbook(AddressBook):
             m = re.match(self.regex, l, self.reflags)
             if m:
                 info = m.groupdict()
-                if 'email' and 'name' in info:
+                if 'email' in info and 'name' in info:
                     email = info['email'].strip()
                     name = info['name']
                     res.append((name, email))
--- a/tests/addressbook/abook_test.py
+++ b/tests/addressbook/abook_test.py
@@ -1,0 +1,38 @@
+# Copyright (C) 2017  Lucas Hoffmann
+# This file is released under the GNU GPL, version 3 or a later revision.
+# For further details see the COPYING file
+from __future__ import absolute_import
+
+import tempfile
+import unittest
+
+from alot.addressbook import abook
+from alot.settings.errors import ConfigError
+
+
+class TestAbookAddressBook(unittest.TestCase):
+
+    def test_abook_file_can_not_be_empty(self):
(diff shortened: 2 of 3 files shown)


tests/addressbook/abook_test.py
```diff
@@ -0,0 +1,38 @@
...
+        [1]
+        name = me
+        email = me@example.com
+        [2]
+        name = you
+        email = you@other.domain, you@example.com
+        """
+        with tempfile.NamedTemporaryFile(delete=False) as tmp:
+            tmp.write(data)
+            path = tmp.name
```
dcbaker: you should add: `self.addCleanup(os.unlink, path)` here to ensure that the temporary file is removed.

tests/addressbook/external_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+from alot.addressbook import external
+
+
+class TestExternalAddressbookGetContacts(unittest.TestCase):
+
+    """Some test cases for
+    alot.addressbook.external.ExternalAddressbook.get_contacts"""
+
+    regex = '(?P<name>.*)\t(?P<email>.*)'
```
dcbaker: this should be a a raw-string (`r'(?P...`)

alot/addressbook/external.py
```diff
@@ -68,7 +68,7 @@ def _call_and_parse(self, commandline):
             m = re.match(self.regex, l, self.reflags)
             if m:
                 info = m.groupdict()
-                if 'email' and 'name' in info:
+                if 'email' in info and 'name' in info:
```
dcbaker: Ohh, that's a subtle bug. Nice catch.

tests/addressbook/external_test.py
lucc: Why? Does it matter if that is a literal tab or the two byte sequence `\t`? And there are not many backslashes in that string so `r'...'` doesn't make it any more readable.

pazz: I'm starting to believe in tests after all :) nice one.

alot/addressbook/external.py
pazz: I'm starting to believe in tests after all :) nice one.

tests/addressbook/external_test.py
dcbaker: No, but it's being explict that this string is meant to be passed into one of the `re` functions. It's not the end of the world either way.

PR #1065, opened 2017-06-13
Add some tests

lucc:
I generated a coverage report with
```sh
coverage2 run setup.py test
coverage2 html --include=alot/'*'
xdg-open htmlcov/index.html
```
and picked some easy things that where not yet covered.

Now I have a question about the last test in the last commit.  It fails for me but I don't think that is consistent so I assume it is a bug.  What do you think about it?  And what is the intended behaviour for alot.commands.thread.determine_sender anyway?

### DIFF
--- a/tests/commands/thread_test.py
+++ b/tests/commands/thread_test.py
@@ -155,3 +155,11 @@ class TestDetermineSender(unittest.TestCase):
         account2 = _AccountTestClass(address='bar@example.com')
         expected = ('foo@example.com', account1)
         self._test(accounts=[account1, account2], expected=expected)
+
+    @unittest.expectedFailure
+    def test_matching_address_and_account_are_returned(self):
+        account1 = _AccountTestClass(address='foo@example.com')
+        account2 = _AccountTestClass(address='to@example.com')
+        account3 = _AccountTestClass(address='bar@example.com')
+        expected = ('to@example.com', account2)
+        self._test(accounts=[account1, account2, account3], expected=expected)
--- a/tests/commands/thread_test.py
+++ b/tests/commands/thread_test.py
@@ -6,9 +6,13 @@
 """Test suite for alot.commands.thread module."""
 from __future__ import absolute_import
 
+import email
 import unittest
 
+import mock
+
 from alot.commands import thread
+from alot.account import Account
 
 # Good descriptive test names often don't fit PEP8, which is meant to cover
 # functions meant to be called by humans.
@@ -59,8 +63,8 @@ class TestClearMyAddress(unittest.TestCase):
 
     def test_only_my_emails_result_in_empty_list(self):
         expected = []
-        actual = thread.ReplyCommand.clear_my_address(self.mine,
-                                                      self.mine+[self.me_named])
+        actual = thread.ReplyCommand.clear_my_address(
+            self.mine, self.mine+[self.me_named])
         self.assertListEqual(actual, expected)
 
     def test_other_emails_are_untouched(self):
@@ -87,3 +91,67 @@ class TestClearMyAddress(unittest.TestCase):
         mine = 'alot team'
         actual = thread.ReplyCommand.clear_my_address(mine, expected)
         self.assertListEqual(actual, expected)
+
+
+class _AccountTestClass(Account):
+    """Implements stubs for ABC methods."""
+
+    def send_mail(self, mail):
+        pass
+
+
+class TestDetermineSender(unittest.TestCase):
+
+    header_priority = ["From", "To", "Cc", "Envelope-To", "X-Envelope-To",
+                       "Delivered-To"]
+    mailstring = '\n'.join([
+        "From: from@example.com",
+        "To: to@example.com",
+        "Cc: cc@example.com",
+        "Envelope-To: envelope-to@example.com",
+        "X-Envelope-To: x-envelope-to@example.com",
+        "Delivered-To: delivered-to@example.com",
+        "Subject: Alot test",
+        "\n",
+        "Some content",
+        ])
+    mail = email.message_from_string(mailstring)
+
+    def _test(self, accounts=(), expected=(), mail=None, header_priority=None,
+              force_realname=False, force_address=False):
+        """This method collects most of the steps that need to be done for most
+        tests.  Especially a closure to mock settings.get and a mock for
+        settings.get_accounts are set up."""
+        mail = self.mail if mail is None else mail
+        header_priority = self.header_priority if header_priority is None \
+            else header_priority
+
+        def settings_get(arg):
+            """Mock function for setting.get()"""
+            if arg == "reply_account_header_priority":
+                return header_priority
+            elif arg.endswith('_force_realname'):
+                return force_realname
+            elif arg.endswith('_force_address'):
+                return force_address
+
+        with mock.patch('alot.commands.thread.settings.get_accounts',
+                        mock.Mock(return_value=accounts)):
+            with mock.patch('alot.commands.thread.settings.get', settings_get):
+                actual = thread.determine_sender(mail)
+        self.assertTupleEqual(actual, expected)
+
+    def test_assert_that_some_accounts_are_defined(self):
+        with mock.patch('alot.commands.thread.settings.get_accounts',
+                        mock.Mock(return_value=[])) as cm1:
+            with self.assertRaises(AssertionError) as cm2:
+                thread.determine_sender(None)
+        expected = ('no accounts set!',)
+        cm1.assert_called_once_with()
+        self.assertTupleEqual(cm2.exception.args, expected)
+
+    def test_default_account_is_used_if_no_match_is_found(self):
+        account1 = _AccountTestClass(address='foo@example.com')
+        account2 = _AccountTestClass(address='bar@example.com')
+        expected = ('foo@example.com', account1)
+        self._test(accounts=[account1, account2], expected=expected)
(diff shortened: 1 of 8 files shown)


tests/db/utils_test.py
```diff
@@ -5,6 +5,8 @@
 from __future__ import absolute_import
 
 import email
+import os
+import os.path
```
dcbaker: You don't need to import `os.path` if you've imported `os`.

tests/db/utils_test.py
```diff
@@ -189,3 +189,74 @@ def test_comma_in_names_are_allowed(self):
...
+    def _test(self, teststring, expected):
+        actual = utils.decode_header(teststring)
+        self.assertEqual(actual, expected)
+
+    def test_non_ascii_strings_are_returned_as_unicode_directly(self):
+        text = u'Nön ÄSCII string¡'
+        self._test(text, text)
+
+    def test_basic_utf_8_quoted(self):
+        text = '=?utf-8?Q?=C3=84=C3=96=C3=9C=C3=A4=C3=B6=C3=BC?='
```
dcbaker: These are difficult to read, is it possible to do something like: `'=?iso-8859-1?Q?' + expected.encode('iso-8859-1')`, or will that produce the wrong values?

tests/commands/thread_test.py
```diff
@@ -87,3 +91,78 @@ def test_real_name_is_never_considered(self):
...
+        self.assertTupleEqual(actual, expected)
+
+    def test_assert_that_some_accounts_are_defined(self):
+        with mock.patch('alot.commands.thread.settings.get_accounts',
+                        mock.Mock(return_value=[])) as cm1:
+            with self.assertRaises(AssertionError) as cm2:
+                thread.determine_sender(None)
+        expected = ('no accounts set!',)
+        cm1.assert_called_once_with()
+        self.assertTupleEqual(cm2.exception.args, expected)
```
dcbaker: does referencing `cm1` and `cm2` actually work after closing the context manager?

tests/db/utils_test.py
lucc: What kind of `encode` funcgtion is that? These are equal signs (`=`) and its all plain ascii. No funky encoding. See RFC 2047.

For the base64 stuff I could use the `base64` module to encode a byte sequence but for the `Q` stuff I don't know.

I could write two helper functions though ...

tests/commands/thread_test.py
lucc: It is documented for the stdlib: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises
and the test passes so I assume it works.

tests/db/utils_test.py
lucc: Is 4ec848a more readable?

lucc:
I want to repeat my question from the initial post: What do you think about the failure of `test_header_order_is_more_important_than_accounts_order:168` ([seen here](https://travis-ci.org/pazz/alot/jobs/242707808#L786)) even though the [line 166](https://github.com/lucc/alot/blob/4ec848a25e0098a63aa824733c22d61aec352e36/tests/commands/thread_test.py#L166) passes? Please have a look at that test case. I am not sure if the function I am testing here is behaving consistent as it looks like the selection of an account for replying to an email is dependent on the fact if a realname is presrent or not.

tests/db/utils_test.py
lucc: I think this is a special case for `os.path` (maybe `os` imports it by default) so I would not like to depend on it (unless it is documented?). In general submodules are not automatically available:
```python
>>> import email
>>> email.header
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'header'
>>> import email.header
>>> email.header
<module 'email.header' from '/usr/lib/python2.7/email/header.pyc'>
>>> import os
>>> os.path
<module 'posixpath' from '/usr/lib/python2.7/posixpath.pyc'>
>>> 
```

lucc:
I found some more seemingly unexpected behaviour of determin_sender with 49cf84b, please check the test failure. Either
1. I misunderstand the docstring and comments (and misdesigned the test) or
2. the docstring and comments don't match the code (but the code is what it should be) or 
3. the docstring and comments describe how it should be but the code is wrong

I will need some help to sort this out. `git-blame` is mostly on you @pazz and @MacGyverNL so maybe you can help me understand that function.

pazz:
@lucc I'm looking at the last test and am very confused by the inlince docs in https://github.com/pazz/alot/blob/master/alot/commands/thread.py#L65 .

The idea was that the config defines a priority among headers, and the account (associated address) that matches the highest-prio header wins. This way, you can reply from the address in the "to" header, even though "delivered-to" was different or vice versa.

If I understand your tests correctly, they do not extract any candidate headers in
https://github.com/pazz/alot/blob/master/alot/commands/thread.py#L72
since 4 lines before, the mock settings return empty?

MacGyverNL:
@pazz The comment looks like I wrote it for the initial version which didn't have a configurable order, simply a hardcoded one which made most sense to me at the time.

@lucc Afaict this goes off the rails at https://github.com/pazz/alot/blob/master/alot/commands/thread.py#L88 because you mock the settings.get()-function. If I understand correctly, any call to settings.get(), regardless of argument, will return your mocked header_priority, which is not an empty list. So, it forces the use of realname = account.realname, which defaults to None in this test, whereas the seen_name would be the empty string. Since we break the loop on ```realname is not None```, the loop is not broken on the correct account; that looks to me like a defect in your test.

However, it also shows that there might be a defect in the code if the _force_realname setting is used with accounts that do not have realnames set; they will simply be skipped rather than get an empty realname. @pazz, thoughts on that?

pazz:
Quoting Pol Van Aubel (2017-06-14 12:18:42)
> However, it also shows that there might be a defect in the code if the
> _force_realname setting is used with accounts that do not have realnames set;
> they will simply be skipped rather than get an empty realname. @pazz, thoughts
> on that?

I think the most reasonable default would then be to have an empty real name (which I believe would result in a mail with `FROM: <my@address.com>`).

MacGyverNL:
Not currently, if ```_force_realname``` is set; afaict ```Account.realname``` defaults to ```None```, not to ```''```, so the loop isn't broken.

It would make way more sense on lines 70, 82 and 98 to use ```address``` instead of ```realname```.

Otherwise, I think we either need to introduce a local variable in thread.py to break the loop on something other than the realname, check explicitly for ```account.realname is None``` and set ```realname = ''``` accordingly, or change the default for ```Account.realname``` to ```''```.

lucc:
I will try to fix my test as far as possible by using a better mock/drop in function for settings.get(), thanks for the hint.

I will also wait for what your further discussion for determine_sender brings up.

(But a warning I will be away till the beginning of next week)

tests/commands/thread_test.py
dcbaker: Wow, that's surprising to me, but it's in the standard docs so it's guaranteed to work.

lucc:
Jippie! I managed to resolve all the problems with determine sender that I have found so far. Also found two bugs in there. (Hoping for new reviews)

tests/db/utils_test.py
```diff
@@ -0,0 +1,59 @@
...
+        'From: me',
+        'To: you',
+        'Subject: header field capitalisation',
+        'Content-type: text/plain; charset=utf-8',
+        'X-Header: param=one; and=two; or=three',
+        "X-Quoted: param=utf-8''%C3%9Cmlaut; second=plain%C3%9C",
+        'X-UPPERCASE: PARAM1=ONE; PARAM2=TWO'
+        '\n',
+        'content'
+        ])
```
dcbaker: It's not a huge deal, but why using `str.join` instead of a  triple quote string?

tests/db/utils_test.py
dcbaker: It's the way that `os.path` is implemented. It has something to the effect of:
```python
if platform == 'Windows':
     import ntpath as path
elif platform == 'Posix':
     import posixpath as path
...
```

Since that has been the behavior for a **long** time, I very much doubt it will change before python 4.

tests/db/utils_test.py
dcbaker: Yes, much more readable, thank you.

tests/commands/thread_test.py
```diff
@@ -178,3 +178,50 @@ def test_doesnt_fail_with_force_realname_if_real_name_not_defined(self):
         expected = ('to@example.com', account2)
         self._test(accounts=[account1, account2, account3], expected=expected,
                    force_realname=True)
+
```
dcbaker: typo in commit message.

tests/db/utils_test.py
lucc: because I would have to dedent the string, no?

lucc:
Sorry about the type, I saw it before but forgot to fix it.

lucc:
@pazz, can you remember why there is a general except in `alot.addressbook.AddressBook.lookup()`?

lucc:
I added two more commits. One with some more tests and one with a small refactor of `alot.db.utils.encode_header` in order to fix some tests. Please have a look at the refactor and the tests for `alot.db.utils.encode_header` to verify that the function formats addresses the way we want it.

tests/db/utils_test.py
dcbaker: That's what `textwrap.dedent` is for. Not a big deal though, more a curiosity.

pazz:
@lucc I don't remember exactly but it looks like laziness on my part really.
Thanks for this. I'm also happy to push this.

lucc:
ok, merged, thanks for the reviews.

Issue #1064, opened 2017-06-09
Prompt unusable - Requests gpg key list each time, 1.5s each


julian-klode:
A new CryptoKeyCompleter instance is created each time you open the prompt, listing all keys in the keyring. This is slow: A single call to crypto.list_keys() takes about 1.5 seconds, probably due to one of the keys being on a smart card (and gpg checking if the smart card is there, I guess).

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        6    9.007    1.501    9.057    1.510 alot/completion.py:565(__init__)

dcbaker:
I actually ran into this recently, a corrupt key in my keyring caused gpg to take over 10 minutes to iterate the keys, so I have thought about this a bit. The problem I see with caching is we need a cheap, reliable way to know when we need to update the keys in the CryptoCompleter instance. I'm not familiar enough with GPG or GPGME to know if there is such a value or what to get it. Ideas very welcome.

meskio:
I use keys in smartcards and I don't notice this slowness.

@julian-klode is `gpg --list-keys` as slow as in alot? Do you have a huge keyring? is your keyring in the old gpg keyring format?

Since gnupg >= 2.1 there is a new keyring format (keybox), it's main improvement is to be fast for big keyrings. Are you using this keyring or the old gpg format? You can check if you have a `~/.gnupg/pubring.kbx` or a `~/.gnupg/pubring.gpg`. If you are using the old format will be nice if you can use the command `migrate-pubring-from-classic-gpg` to upgrade and check if it's still slow.

julian-klode:
It's a normally sized keyring I'd say (about 970 keys) - surprisingly the kbx was empty.  I migrated to kbx now and it's at 0.6 - but that's still too slow.

pazz:
@julian-klode  could you try if#1066 fixes this for you?

julian-klode:
@pazz That seems to solve the issue.

pazz:
OK great, thanks for confirming.

pazz:
I'm closing this issue then, as #1066 got merged. Thanks guys.

PR #1066, opened 2017-06-15
conservative lookup of sub-completers

pazz:
via cached properties.

This should help with #1064 and is generally a good idea because most of the time, only specific sub-completer will be in use and so there is no need to instantiate all of them upfront.

The decorator used comes from the [werkzeug](https://github.com/pallets/werkzeug/blob/10b4b8b6918a83712170fdaabd3ec61cf07f23ff/werkzeug/utils.py#L35) toolkit, which is avoided as a dependency here.

### DIFF
--- a/alot/utils/cached_property.py
+++ b/alot/utils/cached_property.py
@@ -1,6 +1,38 @@
-# werkzeug.utils.cached_property
+# verbatim from werkzeug.utils.cached_property
+#
+#Copyright (c) 2014 by the Werkzeug Team, see AUTHORS for more details.
+#
+#Redistribution and use in source and binary forms, with or without
+#modification, are permitted provided that the following conditions are
+#met:
+#
+#    * Redistributions of source code must retain the above copyright
+#      notice, this list of conditions and the following disclaimer.
+#
+#    * Redistributions in binary form must reproduce the above
+#      copyright notice, this list of conditions and the following
+#      disclaimer in the documentation and/or other materials provided
+#      with the distribution.
+#
+#    * The names of the contributors may not be used to endorse or
+#      promote products derived from this software without specific
+#      prior written permission.
+#
+#THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+#"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+#LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+#A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+#OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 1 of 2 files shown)


alot/utils/cached_property.py
```diff
@@ -0,0 +1,43 @@
...
+    work.
+    """
+
+    # implementation detail: this property is implemented as non-data
+    # descriptor.  non-data descriptors are only invoked if there is
+    # no entry with the same name in the instance's __dict__.
+    # this allows us to completely get rid of the access function call
+    # overhead.  If one choses to invoke __get__ by hand the property
+    # will still work as expected because the lookup logic is replicated
+    # in __get__ for manual invocation.
```
dcbaker: I don't think adding the comment about __get__ is useful, in fact, I think it would be better to replace that mechanism with an assertion that the value is unset.

pazz:
I don't think so: the `PromptCommand` will instantiate a new instance of `CommandLineCompleter` each time it is applied: https://github.com/pazz/alot/blob/master/alot/commands/globals.py#L147
So I'd expect to also get new instances of the gpg key completers each time you use one.

Actually, the caching might be overkill and we can simply get a new instance of the respective completers (inside `CommandLineCompleter.complete`) for each occasion.. but some may be used multiple times in one prompt session (like the `ArgparseOptionCompleter`)

alot/utils/cached_property.py
pazz: This is not my code of course and I assumed that they know what they are doing. If we chose to amend the code we should mention this at the start of the file.
I have no preferences either way.

dcbaker: I recognize the code (I've used it too), but I think disagree with the author's decision to be so liberal with what they accept. Calling dunder methods manually (except for inheritance cases) is ugly and not a good idea. I'm no dead set on this, but my experience is that supporting strange corners comes back to bite you.

dcbaker:
Cool, then I have no problem's merging this.

alot/utils/cached_property.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(self, func, name=None, doc=None):
+        self.__name__ = name or func.__name__
+        self.__module__ = func.__module__
+        self.__doc__ = doc or func.__doc__
+        self.func = func
+
+    def __get__(self, obj, type=None):
```
pazz: meh. go away.

lucc:
I am quite late for the party here I know but please bare with me ...

What is the big advantage of 
```python
@cached_property
def _completer(self):
    return SomeCompleter()
```
over
```py
def __init__(self):
    self._somecompleter_cache = None
@property
def _somecompleter(self):
    if self._somecompleter_cache is None:
        self._somecompleter_cache = SomeCompleter()
    return self._somecompleter_cache
```
We already use the latter aproach when building the widget tree for a message in thread view.

To me it looks like saving three lines of code (per completer) vs. having foreign code+licence in alot.

pazz:
Quoting Lucas Hoffmann (2017-06-21 01:46:16)
> To me it looks like saving three lines of code (per completer) vs. having
> foreign code+licence in alot.

You are right, I forgot that we already do that. Personally, I prefer
the first style because it improves readability IMO. We could also
simply write this decorator ourselves without verbatim quoting the
werkzeug people.. I have no strong feelings either way.

PR #1067, opened 2017-06-16
Add a reload command

dcbaker:
This adds a new command `:reload`, which reloads all configurations when called. There is a bit of refactoring in the first patch to get this all working (we don't want to pass the config path into the methods which reload each config, but make them instance attributes). This refactoring is all fairly straight forward and the end result is a relatively simple patch to add a reload command.


alot/commands/globals.py
```diff
@@ -955,3 +955,12 @@ def apply(self, ui):
         else:
             ui.notify('unknown movement: ' + self.movement,
                       priority='error')
+
+
+@registerCommand(MODE, 'reload', help='Reload all configuration')
```
pazz: should read "configurations"

dcbaker:
On my system the patches are ordered "move tests", "rework the settings manger", then the "add reload command". The test reworks need to be in the same commit as the settings manager reworks since the code they're testing changes in that commit.

pazz:
I'm confused. did you forget to push?

dcbaker:
I did push, but apparently the docs didn't update when I changed the description of the reload command and ran `make html`. It took me a while to get it to finally go through.

pazz:
OK thanks. Let us know when you think that this PR is ready then..
Cheers,
P

NEWS
```diff
@@ -1,3 +1,6 @@
+next:
+* Add command to reload configuration files in running sessin
```
lucc: typo: session

tests/settings/manager_test.py
```diff
@@ -19,8 +19,8 @@ def test_reading_synchronize_flags_from_notmuch_config(self):
             '[maildir]',
             'synchronize_flags = true',
         ]
-        manager = SettingsManager()
-        manager.read_notmuch_config(config)
+        manager = SettingsManager(notmuch_rc=config)
+        manager.read_notmuch_config()
```
lucc: isn't this called by `reload` in the constructor one line above?

dcbaker: yes, the `read_notmuch_config()` call can be removed

dcbaker:
Fixed a few more unit test naming things, added one unittest, enforced that paths be passed for alot and notmuch config, and fixed two unittest that were passing for some reason I don't understand.

Issue #870, opened 2016-05-24
Feature: Adding tags before sending.


lucc:
I was unable to find this feature so I propose it here. (Maybe I missed something?)

When I compose a mail I would like to add tags to it before sending it. Technically speaking the tags would only be applied after the message was send and a copy was saved to disk and registered with notmuch. But I would like to specify such tags in the envelope buffer before sending the mail. Otherwise I have to search for it after it was send and apply the tags. This (the search step) feels cumbersome to me.

pazz:
yes, I agree that this could be a nice feature.
We could replicate the tag commands from thread/search buffers also in envelope buffers,
keep the info in the envelope and tag command in the `afterwards` callback of
`envelope.SendCommand` (https://github.com/pazz/alot/blob/master/alot/commands/envelope.py#L254) .

lucc:
Cool, the constructor of `alot.db.envelope.Envelope` already has a `tags` argument. It is stored in the envelope and later (in the `afterwards` callback you mentioned) these tags are already applied to the message!

I just don't see any place where they are set in the constructor. But the new `:tag` command in an envelope buffer could store the tags there. 

What should the command interface look like? I think if I add a `:tag` command there should be a `:untag` at least. Maybe also a `:retag`?

lucc:
@pazz if we have a tag command in search buffers, in thread buffers and also in evelope buffers it might make sense to abstract some small base class out of these. Yes? Where should it do?

pazz:
nice idea. I'd simply introduce a new TagCommand for envelope buffers
and copy & paste the decorators. IMO the actual command code is
different enough from what we do in other buffers so I'm not sure
what is there to gain from making an abstraction in commands/globals
or so..

Quoting Lucas Hoffmann (2017-07-04 06:03:15)

lucc:
I started implementing it (and it seems to be rather easy). But there is already some documentation for all four commands in `docs/source/usage/modes/envelope.rst` . What is the deal with that? I can not find any corresponding code or functionallity in `master`.

pazz:
Quoting Lucas Hoffmann (2017-07-05 09:05:33)
> I started implementing it (and it seems to be rather easy). But there is
> already some documentation for all four commands in docs/source/usage/modes/
> envelope.rst . What is the deal with that? I can not find any corresponding
> code or functionallity in master.

I dint see it:
https://github.com/pazz/alot/blob/master/docs/source/usage/modes/envelope.rst

[...]

PR #1072, opened 2017-07-05
Implement :tag commands in envelope buffer

lucc:
Proposed in #870.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,6 @@
 next:
 * Add command to reload configuration files in running session
+* new command "tag" (and friends) in EnvelopeBuffer to add additional tags after sending
 
 0.5:
 * save command prompt, recipient and sender history across program restarts
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -625,3 +625,5 @@ class TagCommand(Command):
         elif self.action == 'toggle':
             new = old.symmetric_difference(tags)
         envelope.tags = sorted(new)
+        # reload buffer
+        ui.current_buffer.rebuild()
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -570,3 +570,58 @@ class EncryptCommand(Command):
             envelope.encrypt_keys = {}
         # reload buffer
         ui.current_buffer.rebuild()
+
+
+@registerCommand(
+    MODE, 'tag', forced={'action': 'add'},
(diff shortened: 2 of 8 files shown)


pazz:
Also: please squash  a6b185c and ee39bc7. Thanks!

lucc:
Squashed.

Please have a look at the additional commit. I do copy the tags and it works. But I decided to not copy all tags. A hardcoded list of special tags is deliberatly excluded from the copy. Does that list make sense or are these just the tags that I use?

NEWS
```diff
@@ -1,3 +1,6 @@
+next:
+* new command "tag" (and friends) in EnvelopeBuffer to add additional tags after sendout
```
dcbaker: perhaps 'sending' instead of 'sendout'?

alot/commands/envelope.py
```diff
@@ -141,7 +141,7 @@ def apply(self, ui):
             ui.notify(msg + ' to %s' % path)
             logging.debug('adding new mail to index')
             try:
-                ui.dbman.add_message(path, account.draft_tags)
+                ui.dbman.add_message(path, account.draft_tags+envelope.tags)
```
dcbaker: spaces around the `+` operator please

alot/commands/envelope.py
```diff
@@ -570,3 +570,59 @@ def apply(self, ui):
...
+)
+class TagCommand(Command):
+
+    """manipulate message tags"""
+    repeatable = True
+
+    def __init__(self, tags=u'', action='add', **kwargs):
+        """
+        :param tags: comma separated list of tagstrings to set
+        :type tags: str
```
dcbaker: The type is listed here as a `str` but the default argument is `unicode`. Let's either set this to `basestring` or `unicode` and use an `assert` to ensure it is a unicode object.

lucc:
applied comments and rebased

dcbaker:
There's some test failures now. Either we need to fix the tests or mark that as expected failure. I'll leave it up to you decide what the right thing is.

lucc:
Sorry, the fix was super easy (just add the `u` string prefix to the relevant test args).

dcbaker:
no worries. I still havn't figured out what's wrong with travis on our branches. I've been disabling things until I get it to work, but I'm not quite sure what's wrong yet.

Either way, I think this is ready to land.

PR #1075, opened 2017-07-09
Some misc tests

lucc:
These are some test cases I had lying around.  The test cases where constructed by looking at the code and executing it so find out what the results should be.

The last commit is a rather incomplete test as it only checks some error conditions.

38a2395b adds annotations to the source code in order to exclude some code from coverage reports by [coverage](http://nedbatchelder.com/code/coverage/).  They work analogous to the pylint comments we already have.  I hope it is acceptable that I just add something that might be specific to that tool even though we do not (yet?) officially use it.

### DIFF
--- a/tests/settings/utils_test.py
+++ b/tests/settings/utils_test.py
@@ -1,0 +1,76 @@
+# Copyright (C) 2017  Lucas Hoffmann
+# This file is released under the GNU GPL, version 3 or a later revision.
+# For further details see the COPYING file
+
+"""Tests for the alot.setting.utils module."""
+
+from __future__ import absolute_import
+
+import unittest
+
+import mock
+
+from alot.settings import utils
+
+
+class TestResolveAtt(unittest.TestCase):
+
+    __patchers = []
+    fallback = mock.Mock()
+    fallback.foreground = 'some fallback foreground value'
+    fallback.background = 'some fallback background value'
+
+    @classmethod
+    def setUpClass(cls):
+        cls.__patchers.append(mock.patch(
+            'alot.settings.utils.AttrSpec',
+            mock.Mock(side_effect=lambda *args: args)))
+        for p in cls.__patchers:
+            p.start()
+
+    @classmethod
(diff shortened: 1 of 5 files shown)


pazz:
I'm fine with the extra annotations but not really qualified to review tests :)

tests/settings/utils_test.py
```diff
@@ -0,0 +1,78 @@
...
+        m.background = background
+        return m
+
+    def test_passing_none_returns_fallback(self):
+        actual = utils.resolve_att(None, self.fallback)
+        self.assertEqual(actual, self.fallback)
+
+    def test_empty_string_in_background_picks_up_background_from_fallback(self):
+        attr = self._mock('valid foreground', '')
+        expected = attr.foreground, self.fallback.background
```
dcbaker: I know the parens around tuples in this case are optional, but could we add them for clarity?

tests/settings/utils_test.py
```diff
@@ -0,0 +1,78 @@
...
+        expected = self.fallback.foreground, attr.background
+        actual = utils.resolve_att(attr, self.fallback)
+        self.assertTupleEqual(actual, expected)
+
+    def test_other_values_are_used(self):
+        attr = self._mock('valid foreground', 'valid background')
+        expected = attr.foreground, attr.background
+        actual = utils.resolve_att(attr, self.fallback)
+        self.assertTupleEqual(actual, expected)
+        # FIXME: Do we really care if they are not the same object?
```
dcbaker: I don't think it really matters either way that much if they're the same object or not in this case, it really matters that they have the right behavior.

tests/settings/utils_test.py
lucc: done

tests/settings/utils_test.py
lucc: removed

lucc:
updated

dcbaker:
looks good, thanks

PR #1078, opened 2017-07-10
crypto: Fix handling of signed messages

dcbaker:
This fixes the signature testing code, and it turns out or signature
testing code is completely broken. We've been telling people that
signatures are verified, but they are in fact not verifying correctly.


pazz:
great catch, thanks! I guess that this will be a non-issue when we move to python-gpg but anyway it may make sense to add a test for this..

dcbaker:
we do actually need the second patch for python-gpg, otherwise everything fails to validate.

dcbaker:
But I agree that tests would be good. They're just going to be involved to write.

dcbaker:
I'm writing some tests for this series. I'll send them tomorrow hopefully.

tests/crypto_test.py
```diff
@@ -26,3 +33,65 @@ def test_returned_string_is_lower_case(self):
...
+            f.write(detached)
+            sig = f.name
+        self.addCleanup(os.unlink, f.name)
+
+        with tempfile.NamedTemporaryFile(delete=False) as f:
+            f.write(to_sign)
+            text = f.name
+        self.addCleanup(os.unlink, f.name)
+
+        res = subprocess.check_call(['gpg', '--verify', sig, text])
```
dcbaker: I don't know how to address this in a way that actually fixes it (not just shuts up the warning), on most linux systems `gpg` is installed to `/usr/bin`, on most BSD systems though it's in `/usr/local/bin`, and it could easily be in `/bin` on a non-symlinked system (Most linux distros using systemd symlink `/usr/bin` to `/bin`)

pazz: The only way I can think of is to use the python-gpg in the tests (after having tested its presence). But I'd be cool with just leaving it as is.

dcbaker: I guess I could just use gpgme to verify the data instead of writing it to a file and using gpg directly (or gpg after I get that work done), but if you're cool with it for now this is a valid approach. I can add a TODO.

dcbaker:
Is there anything that prevents us from merging this? This is actually a pretty serious bug for anyone who uses GPG to verify identity in a high risk environment, and at least having it fixed in master would be good, and the second patch is generally useful for the python-gpg work. I also have a branch with more tests for the crypto module that relies on the two `tests:` commits.

dcbaker:
I went ahead and merged this. I'll wait for reviews on the tests, and I'll continue to work on converting to python-gpg

PR #1081, opened 2017-07-12
Add crypto module tests

dcbaker:
This is based on my previous series to fix signed message handling, and provides a fairly comprehensive set of tests for the module.


tests/crypto_test.py
```diff
@@ -26,3 +56,286 @@ def test_returned_string_is_lower_case(self):
...
+                mock.patch('alot.crypto.list_keys',
+                           mock.Mock(return_value=[valid_key, invalid_key])):
+            key = crypto.get_key('placeholder')
+        self.assertIs(key, valid_key)
+
+    def test_ambiguous_two_valid(self):
+        with mock.patch('alot.crypto.gpgme.Context',
+                        mock.Mock(return_value=self._context_mock())), \
+                mock.patch('alot.crypto.list_keys',
+                           mock.Mock(return_value=[make_key() for _ in xrange(2)])):
```
lucc: Is the for loop more readable than two calls to make_key?

tests/crypto_test.py
```diff
@@ -26,3 +56,286 @@ def test_returned_string_is_lower_case(self):
...
+    def test_encrypt(self):
+        to_encrypt = "this is a string\nof data."
+        encrypted = crypto.encrypt(to_encrypt, keys=[crypto.get_key(self.FPR)])
+
+        with tempfile.NamedTemporaryFile(delete=False) as f:
+            f.write(encrypted)
+            enc_file = f.name
+        self.addCleanup(os.unlink, enc_file)
+
+        dec = subprocess.check_output(['gpg', '--decrypt', enc_file])
```
lucc: Again is there a reason you don't use [`gpgme.Context.decrypt`](http://pygpgme.readthedocs.io/en/latest/api.html#gpgme.Context.decrypt)?

tests/static/gpg-keys/amig1-pug.gpg
```diff
@@ -0,0 +1,30 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
lucc: should the file be called amig1-pu**b**.gpg?

tests/crypto_test.py
```diff
@@ -26,3 +56,286 @@ def test_returned_string_is_lower_case(self):
...
+            f.write(detached)
+            sig = f.name
+        self.addCleanup(os.unlink, f.name)
+
+        with tempfile.NamedTemporaryFile(delete=False) as f:
+            f.write(to_sign)
+            text = f.name
+        self.addCleanup(os.unlink, f.name)
+
+        res = subprocess.check_call(['gpg', '--verify', sig, text])
```
lucc: Otherwise you could hide the text output of the gpg command.

lucc:
I just noticed a bunch of lingering gpg-agents from running these tests. I asume they are started by the external gpg call. That is another reason to use the pygpgme bindings. Otherwise we should also find a way to kill them when the tests are done (or even better not start them in the first place, maybe some config file in the temporary gnupghome).

tests/crypto_test.py
dcbaker: Yes, the same reason as the other case, I want to use a different mechanism to prevent hiding bugs through misconfiguration. I can certainly send output to /dev/null

tests/crypto_test.py
```diff
@@ -26,3 +56,286 @@ def test_returned_string_is_lower_case(self):
...
+    def test_list_no_hints(self):
+        # This only tests that you get 3 keys back (the number in our test
+        # keyring), it might be worth adding tests to check more about the keys
+        # returned
+        values = crypto.list_keys()
+        self.assertEqual(len(list(values)), 3)
+
+    def test_list_hint(self):
+        values = crypto.list_keys(hint="ambig")
+        self.assertEqual(len(list(values)), 2)
```
dcbaker: It does not, I'll fix the docstring. I actually returns a `gpgme.KeyIter`, which is obviously an iterator.

tests/crypto_test.py
```diff
@@ -26,3 +56,286 @@ def test_returned_string_is_lower_case(self):
...
+        self.assertEqual(expected, actual)
+
+    def test_validate(self):
+        # Since we already test validation we're only going to test validate
+        # once.
+        expected = self.ctx.get_key(self.FPR).uids[0].uid
+        actual = crypto.get_key(self.FPR, validate=True, encrypt=True, sign=True).uids[0].uid
+        self.assertEqual(expected, actual)
+
+    def test_invalid_key(self):
```
dcbaker: probably a copy and paste error, I'll fix it.

tests/crypto_test.py
dcbaker: Eh, it's either way. I can change it if you think `[make_key(), make_key()]` is more readable.

tests/static/gpg-keys/amig1-pug.gpg
dcbaker: yup. I'll fix that too.

dcbaker:
no, it's not calling gpg directly, it's really that the builtin unittest framework doesn't have a good way to do module level fixtures, so each class that descends from `_TestKey` creates a new gpg keyring during the class fixture stage. I'm also very reluctant to use the python bindings for the first stage, because I want to use a different but equivalent method to test, that way if there's something wrong with the way we're setting up `pygpgme` or `gpg` we'll catch that. I'll see what I can do about either killing the agent or preventing it from starting.

I'll write some code for doing module level fixtures which should help with the performance.

tests/crypto_test.py
lucc: but doesn't `len()` work on iterators also?

teythoon:
> That is another reason to use the pygpgme bindings.

Just a reminder, that is not the binding you are looking for.  python-gpg is:

https://gnupg.org/blog/20160921-python-bindings-for-gpgme.html

lucc:
The man page for `gpg` lists `--no-use-agent` but sais that this has no effect for `gpg2`. There is also `--agent-program` but I am not sure if we can add a script as dummy agent (it might need to be able to speek the agent protocoll).

teythoon:
> The man page for gpg lists --no-use-agent but sais that this has no effect for gpg2. There is also --agent-program but I am not sure if we can add a script as dummy agent (it might need to be able to speek the agent protocoll).

It has to speak the protocol and bind to the sockets the agent binds to. Stubbing this out is not an option.  gpg-agent is not optional in modern versions of GnuPG.

dcbaker:
@teythoon I'm working on porting over to python-gpg, but I prefer to have tests for the functionality before I make changes so I know things still work (and like in a patch series that landed recently, discovering things are broken)

tests/crypto_test.py
dcbaker: no, how can an iterator have a length? An iterator is on object that yields objects in a sequence when it's `__iter__` method is called, and cannot be reiterated, it cannot have a length.

dcbaker:
Okay, I think this should address most of the problems. It gets the crypto tests runtime down from 6.6 seconds to 1.5, which seems pretty good, it also cleans up gpg-agent instances, and hopefully addresses the rest of your comments @lucc.

tests/crypto_test.py
```diff
@@ -95,3 +153,212 @@ def test_verify_signature_bad(self):
...
+    def test_list_hint(self):
+        values = crypto.list_keys(hint="ambig")
+        self.assertEqual(len(list(values)), 2)
+
+
+class TestGetKey(WithKey):
+
+    def test_plain(self):
+        # Test the uid of the only identity attached to the key we generated.
+        ctx = gpgme.Context()
```
lucc: Same question for all these Context instances again: Are they new ones or are they some kind of module level constant/singleton? If not, why is there a key in it?

lucc:
Can we somehow shut up this codacy spamer (for this PR only maybe?)

pazz:
@lucc : i put crypto.py and its tests on the codacy ignore list.. we can also change it completely if you guys prefer.

tests/crypto_test.py
```diff
@@ -12,10 +12,81 @@
...
+
+@MOD_CLEAN.wrap_setup
+def setUpModule():
+    home = tempfile.mkdtemp()
+    MOD_CLEAN.add_cleanup(shutil.rmtree, home)
+    mock_home = mock.patch.dict(os.environ, {'GNUPGHOME': home})
+    mock_home.start()
+    MOD_CLEAN.add_cleanup(mock_home.stop)
+
+    # create a single context to use class wide.
```
dcbaker: leftover from v1. I'll remove it

tests/crypto_test.py
```diff
@@ -12,10 +12,81 @@
...
+@MOD_CLEAN.wrap_teardown
+def tearDownModule():
+    # Kill any gpg-agent's that have been opened
+    lookfor = 'gpg-agent --homedir {}'.format(os.environ['GNUPGHOME'])
+
+    out = subprocess.check_output(['ps', 'xo', 'pid,cmd'], stderr=DEVNULL)
+    for each in out.strip().split('\n'):
+        pid, cmd = each.strip().split(' ', 1)
+        if cmd.startswith(lookfor):
+            subprocess.call(['kill', pid])
```
dcbaker: we already make a lot of assumptions about being on *nix, but os.kill works with python 2.7. I'll use it instead.

tests/crypto_test.py
dcbaker: They're each new ones. Originally I'd attached it to the class level fixture to save some typing, but there isn't a need to. A context isn't a new keyring, it's more of a view into gpg, and theoretically allows you to do some async stuff since Context is implemented in C and doesn't take the gil. I'm not sure that it really matters one way or another, but I've done away with the `WithKey` class, since it was basically unused and moved the fingerprint of the key to a module level variable.

lucc:
Great work, thanks for taking in my comments/suggestions!

lucc:
@pazz I was rather hoping for annotations like for pylint or coverage as that would be more fine grained.

pazz:
@lucc codacy uses the bandit static analysis tool and will look for it's config file within the repo but afaik it is not configurable using inline directives in comments..

PR #1082, opened 2017-07-14
Handle gpg signed and encrypted mail inside a multipart/mixed block

dcbaker:
While working on the python-gpg transition I noticed that there were a lot of threads that notmuch was tagging signed that there was no signature verification being done for. After a bit of looking I realized that these messages were inside a multipart/mixed block, since they came from a mailing list and the mailing list was adding a signature to the bottom that way.

This series adds multipart/mixed handling to the encryption and signature handling code. I have tested the signing code locally, but I don't have any tests for the encrypted. I will get around to writing tests for these, but it seemed pretty involved generating emails with signatures/encrypted bits.

### DIFF
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -29,6 +29,9 @@ charset.add_charset('utf-8', charset.QP, charset.QP, 'utf-8')
 X_SIGNATURE_VALID_HEADER = 'X-Alot-OpenPGP-Signature-Valid'
 X_SIGNATURE_MESSAGE_HEADER = 'X-Alot-OpenPGP-Signature-Message'
 
+_APP_PGP_SIG = 'application/pgp-signature'
+_APP_PGP_ENC = 'application/pgp-encrypted'
+
 
 def add_signature_headers(mail, sigs, error_msg):
     '''Add pseudo headers to the mail indicating whether the signature
@@ -90,6 +93,51 @@ def get_params(mail, failobj=None, header='content-type', unquote=True):
     return {k.lower(): v for k, v in mail.get_params(failobj, header, unquote)}
 
 
+def _handle_signatures(original, message, params):
+    """Shared code for handling message signatures.
+
+    RFC 3156 is quite strict:
+    * exactly two messages
+    * the second is of type 'application/pgp-signature'
+    * the second contains the detached signature
+
+    :param original: The original top-level mail. This is required to attache
+        special headers to
+    :type original: :class:`email.message.Message`
+    :param message: The multipart/signed payload to verify
+    :type message: :class:`email.message.Message`
+    :param params: the message parameters as returned by :func:`get_params`
+    :type params: dict[str, str]
+    """
+    malformed = False
+    if len(message.get_payload()) != 2:
+        malformed = u'expected exactly two messages, got {0}'.format(
+            len(message.get_payload()))
+    else:
+        ct = message.get_payload(1).get_content_type()
+        if ct != _APP_PGP_SIG:
+            malformed = u'expected Content-Type: {0}, got: {1}'.format(
+                _APP_PGP_SIG, ct)
+
+    # TODO: RFC 3156 says the alg has to be lower case, but I've seen a message
+    # with 'PGP-'. maybe we should be more permissive here, or maybe not, this
+    # is crypto stuff...
+    if not params.get('micalg', 'nothing').startswith('pgp-'):
+        malformed = u'expected micalg=pgp-..., got: {0}'.format(
+            params.get('micalg', 'nothing'))
+
+    sigs = []
+    if not malformed:
+        try:
+            sigs = crypto.verify_detached(
+                helper.email_as_string(message.get_payload(0)),
+                message.get_payload(1).get_payload())
+        except GPGProblem as e:
+            malformed = unicode(e)
+
+    add_signature_headers(original, sigs, malformed)
+
+
 def message_from_file(handle):
     '''Reads a mail from the given file-like object and returns an email
     object, very much like email.message_from_file. In addition to
@@ -108,50 +156,25 @@ def message_from_file(handle):
     del m[X_SIGNATURE_MESSAGE_HEADER]
 
     p = get_params(m)
-    app_pgp_sig = 'application/pgp-signature'
-    app_pgp_enc = 'application/pgp-encrypted'
 
     # handle OpenPGP signed data
     if (m.is_multipart() and
             m.get_content_subtype() == 'signed' and
-            p.get('protocol') == app_pgp_sig):
-        # RFC 3156 is quite strict:
-        # * exactly two messages
-        # * the second is of type 'application/pgp-signature'
-        # * the second contains the detached signature
+            p.get('protocol') == _APP_PGP_SIG):
+        _handle_signatures(m, m, p)
+    elif (m.is_multipart() and m.get_content_subtype() == 'mixed' and
+          m.get_payload(0).get_content_subtype() == 'signed'):
+        p = get_params(m.get_payload(0))
+        if p.get('protocol') == _APP_PGP_SIG:
+            _handle_signatures(m, m.get_payload(0), p)
 
-        malformed = False
-        if len(m.get_payload()) != 2:
-            malformed = u'expected exactly two messages, got {0}'.format(
-                len(m.get_payload()))
-        else:
-            ct = m.get_payload(1).get_content_type()
-            if ct != app_pgp_sig:
-                malformed = u'expected Content-Type: {0}, got: {1}'.format(
-                    app_pgp_sig, ct)
-
-        # TODO: RFC 3156 says the alg has to be lower case, but I've
-        # seen a message with 'PGP-'. maybe we should be more
-        # permissive here, or maybe not, this is crypto stuff...
-        if not p.get('micalg', 'nothing').startswith('pgp-'):
-            malformed = u'expected micalg=pgp-..., got: {0}'.format(
-                p.get('micalg', 'nothing'))
-
-        sigs = []
-        if not malformed:
-            try:
-                sigs = crypto.verify_detached(
-                    helper.email_as_string(m.get_payload(0)),
-                    m.get_payload(1).get_payload())
-            except GPGProblem as e:
-                malformed = unicode(e)
-
-        add_signature_headers(m, sigs, malformed)
-
+    # XXX: It is presumably possible to put an encrypted message in a
+    # multipart/mixed, but I have no such examples and haven't looked through
+    # RFC's thuroughly
     # handle OpenPGP encrypted data
     elif (m.is_multipart() and
           m.get_content_subtype() == 'encrypted' and
-          p.get('protocol') == app_pgp_enc and
+          p.get('protocol') == _APP_PGP_ENC and
           'Version: 1' in m.get_payload(0).get_payload()):
         # RFC 3156 is quite strict:
         # * exactly two messages
@@ -162,9 +185,9 @@ def message_from_file(handle):
         malformed = False
 
         ct = m.get_payload(0).get_content_type()
-        if ct != app_pgp_enc:
+        if ct != _APP_PGP_ENC:
             malformed = u'expected Content-Type: {0}, got: {1}'.format(
-                app_pgp_enc, ct)
+                _APP_PGP_ENC, ct)
 
         want = 'application/octet-stream'
         ct = m.get_payload(1).get_content_type()
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -203,11 +203,6 @@ def message_from_file(handle):
                         for k in (X_SIGNATURE_VALID_HEADER,
                                   X_SIGNATURE_MESSAGE_HEADER):
                             m[k] = n[k]
-                    else:
-                        # an encrypted message without signatures
-                        # should arouse some suspicion, better warn
-                        # the user
-                        add_signature_headers(m, [], 'no signature found')
                 else:
                     # 'Combined method', the signatures are returned
                     # by the decrypt_verify function.
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -1,4 +1,6 @@
+# encoding=utf-8
 # Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
+# Copyright © 2017 Dylan Baker <dylan@pnwbakers.com>
 # This file is released under the GNU GPL, version 3 or a later revision.
 # For further details see the COPYING file
 from __future__ import absolute_import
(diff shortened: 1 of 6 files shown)


dcbaker:
I've updated this with tests (which discovered all kinds of broken in my new code :))

@pazz is it possible to get @lucc and I the ability to mark codeclimate issues? In this case it's grumbling about complexity in functions that I split out of a larger, more complex function and I'd like to mark them as "wont fix"

pazz:
neither codacy nor codeclimate seem to allow for shared ownership or granting write access to non-owners. What I could do is kill those accounts (I've connected using my github ID) and set up new accounts using email, and then simply share the passwd with you..

BTW: on codeclimate I also did not find any way to change what it tests for, other than in codacy, where you basically web-configure this command-line tool they use.

dcbaker:
@pazz, could you mark the two codeclimate issues as "wontfix" I've reduced the complexity already, and I don't see how to reduce them further without splitting them in arbitrary ways.

pazz:
Quoting Dylan Baker (2017-08-05 01:41:48)
> @pazz, could you mark the two codeclimate issues as "wontfix" I've reduced the
> complexity already, and I don't see how to reduce them further without
> splitting them in arbitrary ways.

Just did this manually for this PR, and also added a PR wich adds a
codeclimate config for you guys to change. see #1106

dcbaker:
thanks!

dcbaker:
is anyone interested in looking at this?

lucc:
I am and will but I was intimidated because this is even bigger than the other gpg PR. Also I still have #894 lying around and was thinking of opening a subset PR for those changes first because it might help with message introspection by the user. That in turn is something I hope to be usefull when dealing with this PR here. 

Sorry for not posting earlier.

alot/db/utils.py
```diff
@@ -90,6 +95,106 @@ def get_params(mail, failobj=None, header='content-type', unquote=True):
...
 
 
+def _handle_signatures(original, message, params):
+    """Shared code for handling message signatures.
+
+    RFC 3156 is quite strict:
+    * exactly two messages
+    * the second is of type 'application/pgp-signature'
+    * the second contains the detached signature
+    """
```
lucc: Please also add the docs for the arguments.

alot/db/utils.py
```diff
@@ -90,6 +95,106 @@ def get_params(mail, failobj=None, header='content-type', unquote=True):
...
+def _handle_encrypted(original, message):
+    """Handle encrypted messages helper.
+
+    RFC 3156 is quite strict:
+    * exactly two messages
+    * the first is of type 'application/pgp-encrypted'
+    * the first contains 'Version: 1'
+    * the second is of type 'application/octet-stream'
+    * the second contains the encrypted and possibly signed data
+    """
```
lucc: docs for args.

alot/db/utils.py
```diff
@@ -107,122 +212,35 @@ def message_from_file(handle):
...
+              'Version: 1' in m.get_payload(0).get_payload()):
+            _handle_encrypted(m, m)
+
+        # It is also possible to put either of the abov into a multipart/mixed
+        # segment
+        elif m.get_content_subtype() == 'mixed':
+            sub = m.get_payload(0)
+            p = get_params(sub)
+
+            if (sub.get_content_subtype() == 'signed' and
```
lucc: This checks for content type `*/signed` but maybe it should check for `multipart/signed`? Same below for encrypted.

dcbaker: oops, that actually gets dropped in the patch that adds support for encryption, but is present in the one that only adds support for signing.

tests/db/utils_test.py
```diff
@@ -380,3 +387,235 @@ def test_unicode_as_bytes(self):
...
+    def _make_signed():
+        """Create a signed message that is multipart/signed."""
+        text = 'This is some text'
+        t = email.mime.text.MIMEText(text, 'plain', 'utf-8')
+        _, sig = crypto.detached_signature_for(helper.email_as_string(t))
+        s = email.mime.application.MIMEApplication(
+            sig, 'pgp-signature', email.encoders.encode_7or8bit)
+        m = email.mime.multipart.MIMEMultipart('signed', None, [t, s])
+        m.set_param('protocol', 'application/pgp-signature')
+        m.set_param('micalg', 'pgp-sha256')
```
lucc: How do we know it is (always?) sha256?

tests/db/utils_test.py
```diff
@@ -380,3 +387,235 @@ def test_unicode_as_bytes(self):
...
+        m['To'] = 'Nobody'
+        message = utils.message_from_file(io.BytesIO(m.as_string()))
+        self.assertEqual(message.get_payload(), 'This is some text')
+
+    @staticmethod
+    def _make_signed():
+        """Create a signed message that is multipart/signed."""
+        text = 'This is some text'
+        t = email.mime.text.MIMEText(text, 'plain', 'utf-8')
+        _, sig = crypto.detached_signature_for(helper.email_as_string(t))
```
lucc: How can we sign without a key? I did not see anything to explain this in `detached_signature_for()` either.

tests/db/utils_test.py
```diff
@@ -380,3 +387,235 @@ def test_unicode_as_bytes(self):
...
+        m = self._make_signed()
+        m = utils.message_from_file(io.BytesIO(m.as_string()))
+        self.assertEqual(m[utils.X_SIGNATURE_VALID_HEADER], 'True')
+
+    def test_signed_correct_from(self):
+        """Test that the signature is valid."""
+        m = self._make_signed()
+        m = utils.message_from_file(io.BytesIO(m.as_string()))
+        # Don't test for valid/invalid since that might change
+        self.assertIn('ambig <ambig@example.com>', m[utils.X_SIGNATURE_MESSAGE_HEADER])
```
lucc: Why is it this key? Where does it come from?

tests/db/utils_test.py
dcbaker: it doesn't matter, we just check that micalg is set to something that looks relatively valid, but don't actually use it for anything. That code was already there and I left it, but I'm not sure it's worth checking on incoming mail at all.

tests/db/utils_test.py
dcbaker: You know, we never treat the key as optional in alot, we should just make it a required positional argument.

dcbaker: I'll send some updated patches in a bit.

tests/db/utils_test.py
lucc: Ok I thought it had to match what was used in the signature.

tests/db/utils_test.py
```diff
@@ -380,3 +387,235 @@ def test_unicode_as_bytes(self):
...
+
+        m = email.mime.multipart.MIMEMultipart('encrypted', None, [f, e])
+        m.set_param('protocol', 'application/pgp-encrypted')
+
+        return m
+
+    def test_encrypted_length(self):
+        # It seems strang that we just attache the unsigned message to the end
+        # of the mail, rather than replacing the whole encrypted payload with
+        # it's unencrupted enquivalent
```
lucc: Typos: "strang" and "unencrupted".

tests/db/utils_test.py
```diff
@@ -380,3 +387,235 @@ def test_unicode_as_bytes(self):
...
+        m = self._make_encrypted()
+        m = utils.message_from_file(io.BytesIO(m.as_string()))
+        self.assertEqual(len(m.get_payload()), 3)
+
+    def test_encrypted_unsigned_is_decrypted(self):
+        m = self._make_encrypted()
+        m = utils.message_from_file(io.BytesIO(m.as_string()))
+        self.assertEqual(m.get_payload(2).get_payload(), 'This is some text')
```
lucc: Is there some more explicit way to check that the message is not signed? This relies on the knowlesge how a unsigned encrypted mime message has to be structured.

lucc: Ok seeing the next test we should probably not check for these headers here. Maybe check the content type?

tests/db/utils_test.py
```diff
@@ -380,3 +387,235 @@ def test_unicode_as_bytes(self):
...
+        """
+        m = self._make_encrypted()
+        m = utils.message_from_file(io.BytesIO(m.as_string()))
+        self.assertNotIn(utils.X_SIGNATURE_VALID_HEADER, m)
+        self.assertNotIn(utils.X_SIGNATURE_MESSAGE_HEADER, m)
+
+    def test_encrypted_signed_is_decrypted(self):
+        m = self._make_encrypted(True)
+        m = utils.message_from_file(io.BytesIO(m.as_string()))
+        self.assertEqual(m.get_payload(2).get_payload(0).get_payload(),
```
lucc: If you want to be reader friendly you can turn these things into something like this:
```py
multipart_signed = m.get_payload(2)
mime_text = multipart_signed.get_payload(0)
text = mime_text.get_payload()
```

tests/db/utils_test.py
```diff
@@ -380,3 +387,235 @@ def test_unicode_as_bytes(self):
...
+                         'This is some text')
+
+    def test_encrypted_signed_headers(self):
+        """Since the message isn't signed, it shouldn't have headers saying
```
lucc: It *is* and it *should* and that is what the test name says and what the test does.

tests/db/utils_test.py
```diff
@@ -380,3 +387,235 @@ def test_unicode_as_bytes(self):
...
+        multipart/mixed payload, verify that works.
+        """
+        s = self._make_signed()
+        m = email.mime.multipart.MIMEMultipart('mixed', None, [s])
+        m = utils.message_from_file(io.BytesIO(m.as_string()))
+        self.assertIn(utils.X_SIGNATURE_VALID_HEADER, m)
+        self.assertIn(utils.X_SIGNATURE_MESSAGE_HEADER, m)
+
+    def test_encrypted_unsigned_in_multipart_mixed(self):
+        """It is valid to encapsulate a multipart/signed payload inside a
```
lucc: typo "multipart/encrypted"

tests/db/utils_test.py
```diff
@@ -380,3 +387,235 @@ def test_unicode_as_bytes(self):
...
+        multipart/mixed payload, verify that works.
+        """
+        s = self._make_encrypted()
+        m = email.mime.multipart.MIMEMultipart('mixed', None, [s])
+        m = utils.message_from_file(io.BytesIO(m.as_string()))
+        self.assertEqual(m.get_payload(1).get_payload(), 'This is some text')
```
lucc: see two comments above about similar lines

tests/db/utils_test.py
```diff
@@ -380,3 +387,235 @@ def test_unicode_as_bytes(self):
...
+        self.assertNotIn(utils.X_SIGNATURE_VALID_HEADER, m)
+        self.assertNotIn(utils.X_SIGNATURE_MESSAGE_HEADER, m)
+
+    def test_encrypted_signed_in_multipart_mixed(self):
+        """It is valid to encapsulate a multipart/signed payload inside a
```
lucc: typo, wrong type

tests/db/utils_test.py
dcbaker: I thought so too, but we never pass it into gpg (though maybe we should somehow?)

lucc: At this point we have to retrive the algorithm from the generated signature. On master that is done in `alot/db/envelope.py:214`. I don't know why we go not have to pass this info back to gpg when verifying a signature. Maybe the info is also contained in the signature blob and gpg can find it there.

tests/db/utils_test.py
dcbaker: I really don't like attaching the decrypted text to the message; it's a huge layering violation. I'd really like to have another abstraction that can hold that data.

dcbaker:
I have not addressed any of the comments yet, just rebased on master with gpg instead of pygpgme

tests/db/utils_test.py
lucc: Sorry I don't really get your comment. Are you talking about the design of `utils.message_from_file` or are you directly replying to my comment?

dcbaker: Sorta both. I don't like that message_from_file attaches the decrypted/decoded playloads to the original message, which require this sort of ugly to test; but you're right, it's a fragile way to test that the correct behavior happens.

tests/db/utils_test.py
dcbaker: It's from the keys that are stored as static data with the tests. I'm not sure how to make this cleaner, but I'm opened to ideas.

lucc: I assume the key in the signature is set here: https://github.com/dcbaker/alot/blob/submit/gpg-multipart-mixed/tests/db/utils_test.py#L439

But I don't see why it is exactly this key and not any other of the static data keys.
Also I get this test failure locally and son't knwo if it is related:
```
FAIL: test_encrypted_signed_headers (tests.db.utils_test.TestMessageFromFile)
Since the message isn't signed, it shouldn't have headers saying
  File "/home/luc/vcs/alot/tests/db/utils_test.py", line 574, in test_encrypted_signed_headers
    self.assertIn('ambig <ambig@example.com>', m[utils.X_SIGNATURE_MESSAGE_HEADER])
AssertionError: 'ambig <ambig@example.com>' not found in u'Untrusted: ambigu <amigbu@example.com>'

FAIL: test_signed_correct_from (tests.db.utils_test.TestMessageFromFile)
Test that the signature is valid.
  File "/home/luc/vcs/alot/tests/db/utils_test.py", line 465, in test_signed_correct_from
    self.assertIn('ambig <ambig@example.com>', m[utils.X_SIGNATURE_MESSAGE_HEADER])
AssertionError: 'ambig <ambig@example.com>' not found in u'Untrusted: ambigu <amigbu@example.com>'

Ran 246 tests in 3.813s

FAILED (failures=2, expected failures=8)
Test failed: <unittest.runner.TextTestResult run=246 errors=0 failures=2>
error: Test failed: <unittest.runner.TextTestResult run=246 errors=0 failures=2>
```

tests/db/utils_test.py
dcbaker: Also, this isn't tested that the message is or isn't signed. it's testing that an unsigned message is properly decrypted.

tests/db/utils_test.py
dcbaker: I've switched this and the previous to us `m.walk` instead of `m.get_payload`, that way the order doesn't matter.

dcbaker:
Okay, I think I addressed all of the comments so far.

tests/db/utils_test.py
```diff
@@ -380,3 +387,241 @@ def test_unicode_as_bytes(self):
...
+        m['To'] = 'Nobody'
+        message = utils.message_from_file(io.BytesIO(m.as_string()))
+        self.assertEqual(message.get_payload(), 'This is some text')
+
+    def _make_signed(self):
+        """Create a signed message that is multipart/signed."""
+        text = 'This is some text'
+        t = email.mime.text.MIMEText(text, 'plain', 'utf-8')
+        _, sig = crypto.detached_signature_for(
+            helper.email_as_string(t), self.keys)
```
lucc: Thanks!

lucc:
This now LGTM and the test pass locally. Should we use it for some days or merge soon?

PR #1083, opened 2017-07-14
db/utils: Support unicode in signatures

dcbaker:
Currently if a signature name has a non-ascii unicode character in it,
the thread will fail to load because a UnicodeEncodeError. This patch
fixes that by converting the str into unicode.


dcbaker:
What I saw, and I'll restart it when I get home, is that notmuch's git
sever is down so the clone fails.

dcbaker:
since notmuch's git server seems to be back up so the tests pass, and it's reviewed I'll go ahead and merge this.

PR #1085, opened 2017-07-17
settings: Allow loading themes from system locations

dcbaker:
This adds additional system locations to search for theme files, which
are either defined as the environment variable XDG_DATA_DIRS, or the
generic fallback (per the xdg base directories spec) as
/usr/local/share and /usr/share. We then add alot/themees to those paths
to search for themes in.  This allows a package manager to install the
theme files system wide along with alot, and for alot to load them from
the system wide directories.

The Archlinux alot-git package is already installing the packages like
this.

### DIFF
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -4,10 +4,11 @@
 from __future__ import absolute_import
 
 import imp
+import itertools
+import logging
+import mailcap
 import os
 import re
-import mailcap
-import logging
 from configobj import ConfigObj, Section
 
 from ..account import SendmailAccount
@@ -23,6 +24,7 @@ from .theme import Theme
 
 
 DEFAULTSPATH = os.path.join(os.path.dirname(__file__), '..', 'defaults')
+DATA_DIRS = os.environ.get('XDG_DATA_DIRS', '/usr/local/share:/usr/share').split(':')
 
 
 class SettingsManager(object):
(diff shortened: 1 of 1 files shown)


alot/settings/manager.py
```diff
@@ -106,17 +108,26 @@ def read_config(self):
...
+            for dir_ in itertools.chain([themes_dir], data_dirs):
+                if not os.path.isdir(dir_):
+                    logging.warning(
+                        'cannot find theme %s: themes_dir %s is missing',
+                        themestring, dir_)
+                else:
+                    theme_path = os.path.join(dir_, themestring)
+                    try:
+                        self._theme = Theme(theme_path)
+                    except ConfigError as e:
```
pazz: why do you catch this here? Shouldn't we further raise this in so that the interface fails to load in this case?

dcbaker: Maybe this deserves a comment, since I'm using one of python's more unique features, the for/else. If the them fails to load then it will be logged and the loop will continue, if it succeeds the loop will break. If the loop doesn't break, then the else clause will be invoked, raising the ConfigError.

pazz: oh I see! thanks for the explanation. It is indeed a somewhat funky construct, but I see that this way you only need one 'raise' at the end. I'm OK with this.

dcbaker: exactly. for/else is one of my favorite weird python constructs for just that reason.

dcbaker:
I added a comment to the code and updated.

dcbaker:
thanks!

PR #1088, opened 2017-07-17
helper: Do not add a newline to the end of a signed section

dcbaker:
The spec is very clear here:

    (2)  An appropriate Content-Transfer-Encoding is then applied; see
         section 3.  In particular, line endings in the encoded data
         MUST use the canonical <CR><LF> sequence where appropriate
         (note that the canonical line ending may or may not be present
         on the last line of encoded data and MUST NOT be included in
         the signature if absent).

That very last MUST NOT is what's important here, we shouldn't be adding
newlines if they weren't there to begin with. Because we do it causes
signed blobs that have newlines to be invalid since we've changed the
substance of the signed code.

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -585,8 +585,7 @@ def RFC3156_canonicalize(text):
     This function works as follows (in that order):
 
     1. Convert all line endings to \\\\r\\\\n (DOS line endings).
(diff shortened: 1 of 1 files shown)


dcbaker:
thanks!

PR #1089, opened 2017-07-18
commands/envelope: Fall back to account for signing key

dcbaker:
If there isn't a key provided as an argument to sign or togglesign, fall
back to using the account of the sending address to determine the key,
otherwise the message will be marked as "to sign", but won't actually be
signed. This is the same type of logic used for sign_by_default.

### DIFF
--- a/alot/commands/utils.py
+++ b/alot/commands/utils.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 
 import re
 import logging
+
 from twisted.internet.defer import inlineCallbacks, returnValue
 
 from ..errors import GPGProblem, GPGCode
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -16,7 +16,7 @@ from twisted.internet.defer import inlineCallbacks
 
 from . import Command, registerCommand
 from . import globals
-from .utils import set_encrypt
+from . import utils
 from .. import buffers
 from .. import commands
 from .. import crypto
@@ -26,6 +26,7 @@ from ..errors import GPGProblem
 from ..helper import email_as_string
 from ..helper import string_decode
 from ..settings import settings
+from ..settings.errors import NoMatchingAccount
 from ..utils import argparse as cargparse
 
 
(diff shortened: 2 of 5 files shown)


pazz:
I like it. On a side node regarding the new `else` clause:
Maybe we should add a dedicated methon in `Envelope` or a related utility class that extracts a suitable accound from an envelope's From-header. The following code is already replicated in `alot.commands.envelope`:

```python
_, saddr = email.Utils.parseaddr(msg.get('From', ''))
account = settings.get_account_by_address(saddr)
if account is None:
    if not settings.get_accounts():
        ui.notify('no accounts set', priority='error')
        return
    else:
        account = settings.get_accounts()[0]
```

Since `db.envelope.Envelope` already refers to the settings manager internally, this could simply be a new Envelope method I believe.

dcbaker:
I think the helper is a good idea. I put it in utils instead of the Envelope class, I just couldn't justify why it should be in that class other than that it would be convenient, so I made it a utils helper. It's part of this series now.

dcbaker:
bump

alot/commands/utils.py
```diff
@@ -90,3 +96,27 @@ def _get_keys(ui, encrypt_keyids, block_error=False, signed_only=False):
                 continue
         keys[crypto.hash_key(key)] = key
     returnValue(keys)
+
+
+def get_account_by_address(addr, return_default=False):
+    """Get an account given and address.
```
pazz: type *an ?

alot/commands/utils.py
```diff
@@ -90,3 +96,27 @@ def _get_keys(ui, encrypt_keyids, block_error=False, signed_only=False):
...
+
+    :param str addr: The address to get an account for.
+    :param bool return_default: If true returns the first account if no account
+        can be found matching the addr.
+    :raises `~alot.errors.NoMatchingAccount`: If return_default if false and no
```
pazz: typo `s/if/is/`

dcbaker:
I've extended the Manager method with the return_default keyword and updated tests and documentation accordingly.

dcbaker:
thanks! merged

PR #1091, opened 2017-07-25
travis: Use github mirror of notmuch source

dcbaker:
Since the git repo on notmuchmail.org seems to be down.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ before_install: |
     export PATH=/usr/lib/ccache:$PATH
 
     # Clone the notmuch repository and move into it.
-    git clone git://notmuchmail.org/git/notmuch
+    git clone https://github.com/notmuch/notmuch.git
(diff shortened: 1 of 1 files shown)


pazz:
are we sure that this repo is regularly synced? In that case, sure.

dcbaker:
We're not. I'ts 7 days out of date right now, but since neither I nor the CI can clone from notmuchmail.org we need to do something, and this seemed like the most expedient fix. Maybe a better solution would be to skip using git at all and just uses the most recent tarball?

pazz:
Actually, I think you are right and we should use the latest stable release: alot master should work seamlessly with that and  this way we're less afflicted by a buggy development version 
(which I consider rather unlikely) or notmuchmail.org hiccups, which do happen from time to time..

dcbaker:
Okay, I'll do that instead.

PR #1093, opened 2017-07-26
Submit/update docs

dcbaker:
Update documentation for commits 0.5.0 that were missed in the original commits.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,9 @@
 next:
-* Add command to reload configuration files in running session
-* new command "tag" (and friends) in EnvelopeBuffer to add additional tags after sending
+* feature: Add command to reload configuration files in running session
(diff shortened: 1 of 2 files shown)


pazz:
Thanks for this. Regarding the NEWS, in recent releases *cough*years*, I've only listed stuff that changes the way a user interacts with alot. That is, config options and features but not bugfixes.
Shall we stick to this or do you consider the GPG fixes important enough to mention them here?

No objections to the docs patch, which lgtm.

dcbaker:
I think at least mentioning major bug fixes is worth while, since if you rely on gpg signing for cirtical work the fact that 0.5.0 doesn't actually validate said signatures is pretty significant. So maybe drop all of the "bug fix:" lines but that one?

pazz:
sure, that sounds good

PR #1095, opened 2017-07-27
Load default settings even if a user config doesn't exist

dcbaker:
There is a regression introduced by 86186a242c11fcbc501e7afc791a50187e0dd574 that causes default configurations to not be loaded if there is no notmuch config (for notmuch defaults) or not alot config (for alot defaults). This is because in the original patch the configs were not loaded in the constructor, but then were loaded outside the constructor. In the patch the logic that caused the constructor not to load the config was moved out of the constructor and prevented the calls in main from initializing to defaults.

### DIFF
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -67,15 +67,11 @@ class SettingsManager(object):
 
     def read_notmuch_config(self):
         """parse notmuch's config file from path"""
-        if self.notmuch_rc_path is not None:
-            spec = os.path.join(DEFAULTSPATH, 'notmuch.rc.spec')
-            self._notmuchconfig = read_config(self.notmuch_rc_path, spec)
+        spec = os.path.join(DEFAULTSPATH, 'notmuch.rc.spec')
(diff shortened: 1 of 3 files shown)


alot/settings/manager.py
```diff
@@ -68,15 +68,11 @@ def reload(self):
 
     def read_notmuch_config(self):
         """parse notmuch's config file from path"""
-        if self.notmuch_rc_path is not None:
```
pazz: Why is this safe? I mean, what if `self.notmuch_rc_path` is `None` because it is unset due to a missing alot config?
Then two lines below should raise an `ConfigError` which we don't catch here?

dcbaker: ahhh, I didn't look at the notmuch config path closely enough. We definately need to catch an exception there. I'll update and resend.

dcbaker:
I've added tests to demonstrate the implementation works.

@pazz, after writing a test I don't think that notmuch config is an issue. The way `get_notmuch_option` is implemented it will return `None` if it can't find any key/value, so allowing `read_config` to go with `None` is fine.

Issue #949, opened 2017-01-02
Encrypt for BCC addresses


meskio:
Right now alot doesn't encrypt for the addresses in BCC. As discussed in #878 we see two options for it:

- Send separate emails for BCC addresses. This will be a bit tricky to send the same email (same ID and other headers) and keep only one copy in the sent folder.
- Hide the key id with gnupg's `--throw-keyids`. This will still reveal the number of recipients and it's painful to use when you have many private keys, as gnupg needs to try to decrypt it for each private key.

I think the first option makes more sense. Any better ideas?

dkg:
gpg's `--throw-keyids` is universal, while `--hidden-recipient` lets you only hide certain key IDs.

the bcc'ed addresses should be given to gpg via `--hidden-recipient`, rather than using the blunter `--throw-keyids`.

This still leaks the number of bcc'ed recipients to the cc'ed folks, but that's probably OK compared to the cost of generating and sending multiple copies of the mail and changing the mail submission workflow entirely.

dcbaker:
I know I'm dragging this up from the dead but I'm doing it mainly for documentation since I've been beating the gpg code to death recently, but gpgme has no support for `--hidden-recipient`, or `--throw-keyids`, and seemed unwilling to add support for the former at all.

PR #1096, opened 2017-07-28
command/envelope: Prompt before sending an encrypted message with Bcc

dcbaker:
Currently alot cannot encrypt to Bcc recipients, and it isn't obvious
how to implement encrypted BCC without a metadata leak (the key ids of
the Bcc recpients would be visible to the to and cc recipients as well
as the other bcc recipients). As such it hasn't been implemented yet
although #949 is opened for such encryption.

In the mean time alot doesn't encrypt to bcc recipients at all, making
the message all but useless (it might be useful to send it to yourself
as a blind recipient, or to two email addresses of the same person).
Since most people don't know that alot has this limitation, we should
really warn them. This adds a prompt before constructing the message
for that case.

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -11,6 +11,7 @@ import logging
 import os
 import re
 import tempfile
+import textwrap
 
 from twisted.internet.defer import inlineCallbacks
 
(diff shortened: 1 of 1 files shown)


pazz:
Makes sense to me. I'm happy to push it

PR #1099, opened 2017-08-02
Sanitize message count in statusbar after tagging

### LINKED ISSUE #1098: Incorrect message count after running untag --all
I noticed that the message count of currently displayed items is incorrect after running (for example) `untag --all unread`. Please see attached screenshots and the numbering in the lower right.

Before:
img_0043

After:
img_0044

lucc:
Fixes #1098

@dcbaker do you see any easy way to add tests for this?  More general, are there any easy ways to add tests for our commands and the ui widgets?

### DIFF
--- a/alot/commands/search.py
+++ b/alot/commands/search.py
@@ -197,14 +197,6 @@ class TagCommand(Command):
         logging.debug('all? %s', self.allm)
         logging.debug('q: %s', testquery)
 
-        hitcount_before = ui.dbman.count_messages(testquery)
-
-        def remove_thread():
-            logging.debug('remove thread from result list: %s', thread)
-            if threadline_widget in searchbuffer.threadlist:
-                # remove this thread from result list
-                searchbuffer.threadlist.remove(threadline_widget)
-
         def refresh():
             # remove thread from resultset if it doesn't match the search query
             # any more and refresh selected threadline otherwise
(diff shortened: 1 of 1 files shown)


dcbaker:
@lucc, twisted has it's own unittest-like infrastructure, though when I tried to use it I found it somewhat awkward to use. I think that we could also write our own wrappers using `twisted.reactor` manually if we wanted, but I'm not sure that would actually be win.

I agree with pazz though, this feels like there's an off-by-one somewhere, and while this masks the symptom, there is a bug somewhere else that isn't being fixed.

dcbaker:
I'd put some logging information in to see what the hitcount_before and hitcount_after variables are. It might even be an off-by-one in notmuch or the notmuch python bindings.

lucc:
I did the logging thing and it seems that `searchbuffer.result_count == 0` before we hit the line that then updates it again.

lucc:
In `alot/commands/search.py:219` we call `searchbuffer.rebuild()` which will update the count. So thats why the count is already up to date when we modify it again. Fix comming up.

lucc:
I hope this is not cheating. I tried to outline my thoughts in the commit messages. Will squash when you approve.

pazz:
@dcbaker: regarding a test corpus: the guys over at notify already do this. Maybe we could reuse that?

dcbaker:
@pazz, could you point me to the repo in question, I'm not sure which notify to look at.

pazz:
@dcbaker 
Oh no! autocompletion on my tablet is to blame: i meant notmuch, not notify! https://github.com/notmuch/notmuch/tree/master/test

dcbaker:
ooooh, that makes sense.

Issue #1076, opened 2017-07-09
Tests failures on travis for forked repo


lucc:
The travis tests for lucc/alot [fail](https://travis-ci.org/lucc/alot/builds) recently. Normally it's an `AttributeError: 'module' object has no attribute 'abook_test'` when the tests are run. Does this happen for any body else? Apearently master in this repo builds just fine.

Also there is a supressed error for the command `pip install .[test]` in the install step. pip also doesn't understand that command locally on my end. @dcbaker you added it in 3e0e2d463. How should that work? Does it work for you?

pazz:
I cannot reproduce this:
https://gist.github.com/pazz/384fb847a047476d44a5c20269f49ea1

dcbaker:
I have the same bug.

dcbaker:
I'll look into it. thanks for bisecting @lucc

dcbaker:
It's due to updates to some dependent library. After updating my virtualenv I can reproduce this an similar errors on my own machine. I'll keep looking into it.

dcbaker:
Okay, I finally broke out git bisect, and this is broken as soon as the abook tests are added. I think it's actually tickling a real bug in the settings module (I've traced it back to something in `settings.utils.read_config`), but I'm still not exactly sure what it is or why it only manifests on travis.

PR #1101, opened 2017-08-03
Fix tests failing on other repos other than pazz's

dcbaker:
There are two problems here, the first is a circular import that I can suddenly reproduce locally. The second is that the gpgme tests don't work without sudo, since they need access to /dev/pts, which isn't available in the docker container.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,7 @@ sudo: True
 # Cache pip objects to speed up next build
 cache:
   pip: True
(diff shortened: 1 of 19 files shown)


dcbaker:
and perhaps more relavent: https://travis-ci.org/dcbaker/alot/builds/260812826

lucc:
Fixes #1076 
(let's see if this works)

PR #1103, opened 2017-08-04
Import single names in helpers.py

lucc:
Just an old commit I found.


dcbaker:
I'm really not crazy about massive numbers of imports without namespaces, they just don't feel very pythonic to me.

lucc:
ok

PR #1108, opened 2017-08-08
Handle servers that treat the user segment of an address as case insensitive

dcbaker:
Alot (like other email clients that aren't tightly coupled to one mail server) are in a difficult position, RFC 5321 section 2.3.11 (which is in the docstring of the "account: Add an Address class" patch), says that only the server for the receiving domain (for example, gmail.com) is allowed to alter or interpret the user name segment (everything before the @) in any way.

Why is this a problem for alot? Well, we need to apply the same semantics. If the server for "example.com" treats "foo@example.com" and "Foo@example.com" as equivalent (which it may), alot needs to do the same. If example.com treats them as different accounts, then alot does too.

There are simply too many servers out there to try to maintain a list of who does what (and it's a dangerous game to play, because some servers apply additional semantics, like gmail and fastmail that ignore anything after a "+" sign, so "foo@example.com" and "foo+spam@example.com" are the same to these services) and we don't want to manage that.

Instead this series takes the approach of adding a new class that acts like a `unicode` or `str` instance, including being comparable with rich comparison operators (at least == and !=, although the other's could easily be added) which can be flagged as either case sensitive or not, and adds a new option to the account section of the configuration to control this flag.

### DIFF
--- a/alot/account.py
+++ b/alot/account.py
@@ -1,4 +1,6 @@
+# encoding=utf-8
 # Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
+# Copyright © 2017 Dylan Baker
 # This file is released under the GNU GPL, version 3 or a later revision.
 # For further details see the COPYING file
 from __future__ import absolute_import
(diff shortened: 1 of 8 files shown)


pazz:
some thoughts on this:

* do we need case (in)sensitive string operations for anything else but email addresses?
* would it be convenient if Address not also contain realname parts and able to render as `real name <bla@blub.org>`?
* do we know of *any* service that does not adhere to the sane standard of case insensitive usernames? Does having a switch for this even makes sense? After all, it complicates our code.
* why do we need to render email addresses as unicode? (aren't they always ascii?) and how much does this interfere with the python3 port re strings?

lucc:
* I don't know of anything else that needs this kind of special string operations.
* maybe not see below
* As mentinoed before I do not know of any service that doesn't interpret the local part case insensitive. But I still think we need to support the official standard. We could default to insensitive so users would have to set the option to get RCF compliance. 
* The realname can be anything, hence unicode for that. The local part of the address is ascii.

dcbaker:
RFC 6531, Section 3.2:
```
An SMTP server that announces the SMTPUTF8 extension MUST be prepared
to accept a UTF-8 string [RFC3629] in any position in which RFC 5321
specifies that a <mailbox> can appear. 
```

So yes, both the user name and the domain may contain utf-8.

dkg suggested that there are some services that reject alternate capitalizations of words, although I haven't seen any. I think the important thing here is that an email service is allowed to treat different capitalizations differently, even if in practice no modern service does (the implication of RFC 5321 is that there were at one time services that did). I guess we could remove the case-sensitive handling though, since I haven't been able to find evidence of a (currently in use) SMTP server that actually treats Foo@example.com differently than foo@example.com.

The switch currently defaults to case insensitive, which I think is the sane default, I agree.

For python3 this should actually be pretty simple, all we should need to do to make it all work is drop the `bytes` handling and enforce that we always get a `str` (using py3 terminology).

Also, I just realized there is a bug in this series as is, I'm not comparing domain names correctly.

dcbaker:
I also removed a variable duplicated between the account and the address classes.

dcbaker:
I also added a `__hash__` method, since this implements an `__eq__` method.

pazz:
Thanks, this looks OK to me now.

Up for general discussion: What do you guys think of grouping the Attachment, Envelope, and Address classes together somehow?
Currently, our code layout has "database" related classes (wrapper around notmuchs classes) separate from "settings" classes such as Account. But the ones above somehow fit in neither category imo.

dcbaker:
Attachment and Envelope definitely belong together, but I think that Address belongs with Account. Account is the only consumer of Address atm, and I'm not really sure where else we'd use it to be honest. I do think there's some things in the db module that aren't really about the database, and envelope and attachment are among them.

Issue #986, opened 2017-01-18
command `send; bclose` will prompt to close message without sending, even when succesful


dcbaker:
This is an interesting bug, as the actual desired behavior is not obvious. If the send succeeds, then it will close the buffer, but if it fails it will not close the buffer. This means that the result of the bclose is dependent on the status of the SendCommand.apply, but we have no way to pass that result.

There are two possible resolutions to this. The first, and maybe best resolution is to stop the command stream if the command hits the Errback path. The alternative is to offer to close the buffer even if the command fails, so that the next command will be consistantly be applied to the next buffer, not the envelope.

pazz:
I'd opt for the first solution, with an additional error notification for now.
Another more complicated solution would be to introduce return values but then we're beginning to implement a scripting language and I don't want to open that can.

dcbaker:
I think the first solution is the best one too. It may not be optimal, but at least it's easy to be consistant.

Issue #985, opened 2017-01-18
validate list of commands when using ;


dcbaker:
When passing commands via the `:` prompt, the commands are unvalidated. If the first X commands are valid then the execution will stop when attempting to instantiate the invalid command.

As disscussed in #966 it is important to not instantiate the commands to validate them.

pazz:
for some context: We would like that a commandline say `bclose; dontexist; taglist` doesn't even start with the first command since the second one is syntactic garbage.
One caveat is that a command may be OK in one mode, but not in another.
So that for instance `taglist;compose;send;send` should really not be OK, as the second `send` would be instantiated out in a taglist buffer, where it makes no sense.

Issue #579, opened 2013-03-09
envelope header cmdline parsing is extremely fragile


josch:
in the current master branch it is not possible to edit the following to/cc/bcc header:

```
"foo bar" <foo@bar.com>, "blub bla" <blub@bla.com>
```

It will be converted to:

```
"foo bar <foo@bar.com>, blub bla" <blub@bla.com>
```

Because of the greedy regex around line 164 in `alot/commands/__init__.py`. Replacing the experession with a non-greedy one will fix this issue:

```
@@ -161,7 +161,7 @@ def commandfactory(cmdline, mode='global'):
     # allow to shellescape without a space after '!'
     if cmdline.startswith('!'):
         cmdline = 'shellescape \'%s\'' % cmdline[1:]
-    cmdline = re.sub(r'"(.*)"', r'"\\"\1\\""', cmdline)
+    cmdline = re.sub(r'"(.*?)"', r'"\\"\1\\""', cmdline)
     try:
         args = split_commandstring(cmdline)
     except ValueError, e:
```

But the parsing is still fragile at best! What if there is a (backslash escaped) quote inside the name like:

```
"foo \"bar" <foo@bar.com>
```

This could be caught by a lookbehind expression like:

```
r'"(.*?)(?<!\\)"'
```

But that one would not catch the case in which the backslash itself was escaped...

The python implementations to convert the result of shlex.split() back into a commandline can be found in pipes.quote and (since python 3.3) in shlex.quote. But both use the single quotes of the shell to get around the nasty escaping problem.

Is there a good way to fix this problem once and for all?

Maybe put the entire header into a single, single-quoted argument on the commandline?

PR #1109, opened 2017-08-09
RFC: How to treat quoting and escaping on the command line?

lucc:
I was trying to execute `:call "import alot;ui.notify(alot.__version__)"` and it did not work because the semicolon was used to split up the command line.  This lead to the observation that our command line parsing code needs some hardening regarding quoting and escaping of special chars (actually only ";").  (Note that I have since solved my original problem, see newest entries in the wiki.)

The tests I added show what *I* believe the command line parsing should do.  Do you agree or differ?  Please also post more `base = '...'; expected = ['...']` examples that should also be put in the tests.  I will gradually update the commits.  Once we have finished the initial discussion what the parser should do I will turn this into a proper PR (you probably should wait until then when you want to do full reviews).

The relevant part of the shlex docs is [here](https://docs.python.org/2.7/library/shlex.html#parsing-rules).  Some recent issues and PRs that also dealt with the command line parsing are #579, #985, #986 although none of them seems directly related.

The current solution for the first test is not complete.  It doesn't work with single quotes and backslashes in quotes and nested quotes are not handled.  In general this path (the current solution of doing some special `str.replace` calls before calling `shlex.shlex`) seems cumbersome in the long run and difficult to get right for all cases.  I had an idea for another way of solving the issue in general:  I could try to not set `shlex.shlex.whitespace` and `shlex.shlex.whitespace_split` to anything but the default.  This will split at normal whitespace and we can then later on "split" the list of strings at entries that are just one semicolon.  The lists for each individual command could then be joined with whitespace again.  This would potentially change the whitespace so the question for @pazz is:  Is this function used to process any strings where the amount of whitespace should not be changed (subject lines, email addresses, ...)?


tests/helper_test.py
```diff
@@ -129,6 +129,18 @@ def test_unicode(self):
...
 
+    @unittest.expectedFailure
+    def test_doublequoted_semicolons(self):
+        base = 'python -c "import alot; print(alot)"; echo "done"'
+        expected = ['python -c "import alot; print(alot)"', ' echo "done"']
+        self._test(base, expected)
+
+    @unittest.expectedFailure
+    def test_singlequoted_semicolons(self):
+        base = "python -c 'import alot; print(alot)'; echo 'done'"
```
dcbaker: These are broken on archlinux, `python` points to `python3`, and `alot` isn't available for python3. Can we either use a stdlib package like `os`, or explicitly use `python2`?

lucc: This is a string to test the splitting function. It is not to be executed. I just wipped something together which looks like valid shell code in order to use my intuition to find the desired result.

dcbaker: okay, consider my object withdrawn.

pazz:
Sorry to keep this hanging. @lucc I guess this is still relevant is it? I generally don't mind the changes and adding tests is certainly valuable. Could you rebase?

> Is this function used to process any strings where the amount of whitespace should not be changed (subject lines, email addresses, ...)?

Well yes: it is used by the ui at the point [where commandline input is interpreted](https://github.com/pazz/alot/blob/master/alot/ui.py#L259). So this means in particular if you use `set Subject foo bar` in envelope mode you probably want to preserve whitespaces.

lucc:
I rebased so we can better discuss this. There was some subtile issue with these changes but I have to think about this code thoroughly again to remember it.

PR #1112, opened 2017-08-10
Fix Issue 1110

dcbaker:
The bug in 1110 is that if sign_by_default is true, but no key is defined alot still tries to sign, even without a key. This results in undefined behavior, and some key is used, possibly changing the account. The right solution is to simply forbid sign_by_default without a gpg_key.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -857,8 +857,14 @@ class ComposeCommand(Command):
 
         # Figure out whether we should GPG sign messages by default
         # and look up key if so
-        self.envelope.sign = account.sign_by_default
-        self.envelope.sign_key = account.gpg_key
+        if account.sign_by_default:
+            if account.gpg_key:
+                self.envelope.sign = account.sign_by_default
+                self.envelope.sign_key = account.gpg_key
+            else:
+                msg = 'Cannot find gpg key for account {}'.format(account.address)
+                logging.warn(msg)
+                ui.notify(msg, priority='error')
 
         # get missing To header
         if 'To' not in self.envelope.headers:
--- a/tests/commands/global_test.py
+++ b/tests/commands/global_test.py
@@ -25,11 +25,6 @@ import mock
 from alot.commands import globals as g_commands
 
 
-def expected_failure(func):
-    func.todo = 'Expected Failure'
-    return func
-
-
 class Stop(Exception):
     """exception for stopping testing of giant unmanagable functions."""
     pass
@@ -78,7 +73,6 @@ class TestComposeCommand(unittest.TestCase):
         self.assertTrue(envelope.sign)
         self.assertIs(envelope.sign_key, mock.sentinel.gpg_key)
 
-    @expected_failure  # sign_key is set even though sign is False
     @inlineCallbacks
     def test_apply_sign_by_default_false_doesnt_set_key(self):
         envelope = self._make_envelope_mock()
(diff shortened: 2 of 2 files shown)


alot/commands/globals.py
```diff
@@ -857,8 +857,14 @@ def apply(self, ui):
...
-        self.envelope.sign = account.sign_by_default
-        self.envelope.sign_key = account.gpg_key
+        if account.sign_by_default:
+            if account.gpg_key:
+                self.envelope.sign = account.sign_by_default
+                self.envelope.sign_key = account.gpg_key
+            else:
+                msg = 'Cannot find gpg key for account {}'.format(account.address)
+                logging.warn(msg)
+                ui.notify(msg, priority='error')
```
lucc: This change looks good. It also makes `sign_by_default` work like `:sign` which will fail if no key is provided as argument or as the accounts `gpg_key` option.

I think I also found out why `sign_by_default` did previously work when `account.gpg_key = None`: The attribute `envelope.sign_key` is not touched at all until the `:send` command calls [`envelope.construct_mail()`](https://github.com/pazz/alot/blob/4f78529875c2687e58586c43512a71787f72fccb/alot/db/envelope.py#L169-L195) which just passes the key to [`detached_signature_for()`](https://github.com/pazz/alot/blob/4f78529875c2687e58586c43512a71787f72fccb/alot/crypto.py#L150-L170). Now if the key is `None` in there I assume that gpg will just pick up the default key from the gpg config file.

tests/commands/global_test.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class TestComposeCommand(unittest.TestCase):
+
+    """Tests for the compose command."""
+
+    @staticmethod
+    def _make_envelope_mock():
+        envelope = mock.Mock()
+        envelope.headers = {'From': 'foo foo@example.com'}
```
lucc: I think that is not a valid from header. It should be `foo <foo@example.com>`.

alot/commands/globals.py
```diff
@@ -857,8 +857,14 @@ def apply(self, ui):
...
         # Figure out whether we should GPG sign messages by default
         # and look up key if so
-        self.envelope.sign = account.sign_by_default
-        self.envelope.sign_key = account.gpg_key
+        if account.sign_by_default:
+            if account.gpg_key:
+                self.envelope.sign = account.sign_by_default
+                self.envelope.sign_key = account.gpg_key
+            else:
+                msg = 'Cannot find gpg key for account {}'.format(account.address)
```
pazz: alternatively, we can vakidate this in the config spec..

tests/commands/global_test.py
```diff
@@ -0,0 +1,118 @@
...
+        envelope.get = lambda x: envelope.headers[x]
```
lucc: Why not `envelope.get = envelope.headers.get`?

alot/commands/globals.py
dcbaker: That's my understanding as well, prompted by your question about another test I'd written where `detached_signature_for`'s key argument is unset. A nice side effect of this is that with this change we can force key to be set always (since it always will be)

alot/commands/globals.py
dcbaker: I could go either way.

dcbaker:
I've address @lucc's comments and removed the #Fixes 1110 tag, since the reporter says this doesn't fix his issue.

PR #1115, opened 2017-08-15
Fix issue with account selection

### LINKED ISSUE #1110: Mail sent by sign adress instead of from address
I use alot (msmtp) to send mails with to accounts. I use the same GPG key to sign both mails. Things used to work, but since commit 2504de648dd6413625387924baaca126eac68016 or 3995747c4efed39a74e75b5c11061574bc1e8cd8 of @dcbaker, all mails are sent through the account I sign with, not the one in the "from" header.

I'm mentioning the GPG part because of the contents of the mentioned commit. I can't image something else causing this commit to influence the sender.

pazz:
The problem is that `commands.envelope.SendCommand` simply passes the content
of the From-header to `settings.get_account_by_address`, without removing the realname part.
The problem was introduced in 084e37c, which silently drops the parsing of structured address strings .

So if the From-header is of the form "real name <email@address.org>", this does not match the account with
address "email@address.org" and falls back to the default account, which could be different.
I noticed this only by chance because my default provider (gmail) actually processes (and changes the sender) when given mails with foreign From-header values.

The first patch here moves this call into `get_account_by_address` itself, the second patch simplifies all calls to this method.

This might also fix #1110

### DIFF
--- a/alot/account.py
+++ b/alot/account.py
@@ -193,7 +193,7 @@ class SendmailAccount(Account):
         """Pipe the given mail to the configured sendmail command.  Display a
         short message on success or a notification on error.
         :param mail: the mail to send out
-        :type mail: str
+        :type mail: :class:`email.message.Message` or string
         :returns: the deferred that calls the sendmail command
         :rtype: `twisted.internet.defer.Deferred`
         """
@@ -213,6 +213,9 @@ class SendmailAccount(Account):
             logging.error(failure.getTraceback())
             logging.error(failure.value.stderr)
             raise SendingMailFailed(errmsg)
+
+        # make sure self.mail is a string
+        mail = str(mail)
 
         d = call_cmd_async(cmdlist, stdin=mail)
         d.addCallback(cb)
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -236,11 +236,6 @@ class SendCommand(Command):
             return
         logging.debug("ACCOUNT: \"%s\"" % account.address)
 
-        # make sure self.mail is a string
-        logging.debug(self.mail.__class__)
-        if isinstance(self.mail, email.message.Message):
-            self.mail = str(self.mail)
-
         # define callback
         def afterwards(_):
             initial_tags = []
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -118,7 +118,7 @@ class SaveCommand(Command):
         # determine account to use
         try:
             account = settings.get_account_by_address(
-                envelope.get('From'), return_default=True)
+                envelope['From'], return_default=True)
         except NoMatchingAccount:
             ui.notify('no accounts set.', priority='error')
             return
@@ -226,12 +226,15 @@ class SendCommand(Command):
         msg = self.mail
         if not isinstance(msg, email.message.Message):
             msg = email.message_from_string(self.mail)
+        address = msg.get('From', '')
+        logging.debug("FROM: \"%s\"" % address)
         try:
-            account = settings.get_account_by_address(
-                msg.get('From', ''), return_default=True)
+            account = settings.get_account_by_address(address,
+                                                      return_default=True)
         except NoMatchingAccount:
             ui.notify('no accounts set', priority='error')
             return
+        logging.debug("ACCOUNT: \"%s\"" % account.address)
 
         # make sure self.mail is a string
         logging.debug(self.mail.__class__)
@@ -505,8 +508,7 @@ class SignCommand(Command):
                     return
             else:
                 try:
-                    _, addr = email.utils.parseaddr(envelope.headers['From'][0])
-                    acc = settings.get_account_by_address(addr)
+                    acc = settings.get_account_by_address(envelope['From'])
                 except NoMatchingAccount:
                     envelope.sign = False
                     ui.notify('Unable to find a matching account',
--- a/tests/commands/envelope_test.py
+++ b/tests/commands/envelope_test.py
@@ -170,10 +170,10 @@ class TestSignCommand(unittest.TestCase):
     @staticmethod
     def _make_ui_mock():
         """Create a mock for the ui and envelope and return them."""
-        envelope = mock.Mock()
+        envelope = Envelope()
+        envelope['From'] = 'foo <foo@example.com>'
         envelope.sign = mock.sentinel.default
         envelope.sign_key = mock.sentinel.default
-        envelope.headers = {'From': ['foo <foo@example.com>']}
         ui = mock.Mock(current_buffer=mock.Mock(envelope=envelope))
         return envelope, ui
 
--- a/tests/commands/envelope_test.py
+++ b/tests/commands/envelope_test.py
@@ -23,7 +23,9 @@ import os
 import shutil
 import tempfile
 import textwrap
-import unittest
+
+from twisted.trial import unittest
+from twisted.internet.defer import inlineCallbacks
 
 import mock
 
@@ -370,18 +372,20 @@ class TestSendCommand(unittest.TestCase):
         Foo Bar Baz
         """)
 
+    @inlineCallbacks
     def test_get_account_by_address_with_str(self):
         cmd = envelope.SendCommand(mail=self.mail)
         account = mock.Mock()
         with mock.patch(
                 'alot.commands.envelope.settings.get_account_by_address',
                 mock.Mock(return_value=account)) as get_account_by_address:
-            cmd.apply(mock.Mock())
+            yield cmd.apply(mock.Mock())
         get_account_by_address.assert_called_once_with('foo@example.com',
                                                        return_default=True)
         # check that the apply did run through till the end.
         account.send_mail.assert_called_once_with(self.mail)
 
+    @inlineCallbacks
     def test_get_account_by_address_with_email_message(self):
         mail = email.message_from_string(self.mail)
         cmd = envelope.SendCommand(mail=mail)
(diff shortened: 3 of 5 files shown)


tests/commands/envelope_test.py
```diff
@@ -170,7 +170,7 @@ class TestSignCommand(unittest.TestCase):
     @staticmethod
     def _make_ui_mock():
         """Create a mock for the ui and envelope and return them."""
-        envelope = mock.Mock()
+        envelope = Envelope()
         envelope.sign = mock.sentinel.default
         envelope.sign_key = mock.sentinel.default
         envelope.headers = {'From': ['foo <foo@example.com>']}
```
dcbaker: If you're going to use an actual envelope you should probably set the headers rathaer than replacing them.

alot/commands/envelope.py
```diff
@@ -226,9 +226,12 @@ def apply(self, ui):
         msg = self.mail
         if not isinstance(msg, email.message.Message):
             msg = email.message_from_string(self.mail)
+            address = msg.get('From', '')
+            logging.debug("FROM: \"%s\"" % address)
```
lucc: Shouldn't these two lines be dedented one level?

lucc:
I uploaded one new test case for this PR to https://github.com/lucc/alot/tree/test/1115 . If you want you can add that commit to this PR. Obviously you can also modify it. The other places you touched are inside of apply methods of commands which I generally precieved harder to test. Maybe I can come up with a test for something there later. (Don't wait for me ;)

lucc:
I added two more tests to the mentioned branch.

alot/commands/envelope.py
pazz: yes, thanks.

tests/commands/envelope_test.py
pazz: done, via `envelope.__setitem__`..

pazz:
@lucc: the last test fails with

```
FAIL: test_get_account_by_address_with_email_message (tests.commands.envelope_test.TestSendCommand)
  File "/home/pazz/projects/alot/tests/commands/envelope_test.py", line 396, in test_get_account_by_address_with_email_message
    account.send_mail.assert_called_once_with(mail)
  File "/home/pazz/.local/lib/python2.7/site-packages/mock/mock.py", line 948, in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
  File "/home/pazz/.local/lib/python2.7/site-packages/mock/mock.py", line 937, in assert_called_with
    six.raise_from(AssertionError(_error_message(cause)), cause)
  File "/home/pazz/.local/lib/python2.7/site-packages/six.py", line 718, in raise_from
    raise value
AssertionError: Expected call: send_mail(<email.message.Message instance at 0x7ffb20294368>)
Actual call: send_mail('From nobody Wed Aug 16 09:14:31 2017\nFrom: foo@example.com\nTo: bar@example.com\nSubject: FooBar\n\nFoo Bar Baz\n')
```

I don't quite understand what exactly it does but as it tests the `get_account_by_address` method, shouldn't this go into a a test class for the settings manager?

pazz:
The last commit cleans up the signature of `SendmailAccount` slightly, simplifies the `SendCommand` and makes the tests pass as they are..

tests/commands/envelope_test.py
```diff
@@ -355,3 +356,41 @@ def test_apply_from_user_and_email(self):
...
+        """)
+
+    def test_get_account_by_address_with_str(self):
+        cmd = envelope.SendCommand(mail=self.mail)
+        account = mock.Mock()
+        with mock.patch(
+                'alot.commands.envelope.settings.get_account_by_address',
+                mock.Mock(return_value=account)) as get_account_by_address:
+            cmd.apply(mock.Mock())
+        get_account_by_address.assert_called_once_with('foo@example.com',
```
lucc: yes it is outside but the right side of the `with ... as` is even known after the with block. You can find this kind of mock.patch handling in the official docs I think.

tests/commands/envelope_test.py
```diff
@@ -355,3 +356,41 @@ def test_apply_from_user_and_email(self):
...
+        cmd = envelope.SendCommand(mail=mail)
+        account = mock.Mock()
+        with mock.patch(
+                'alot.commands.envelope.settings.get_account_by_address',
+                mock.Mock(return_value=account)) as get_account_by_address:
+            cmd.apply(mock.Mock())
+        get_account_by_address.assert_called_once_with('foo@example.com',
+                                                       return_default=True)
+        # check that the apply did run through till the end.
+        account.send_mail.assert_called_once_with(mail)
```
lucc: I would rather use `account.send_mail.assert_called_once()` instead of dropping it entirely. I added this assert line in the first place because I had the feeling that I want to assert that one of the last lines of `apply()` is executed. Maybe I was just to confused by all these if branches that can `yield` and `return` and all these asynchronous things.

Proceed as you will.

lucc:
tests.commands.envelope_test.TestSendCommand.test_get_account_by_address_with_email_message should *not* be moved to the settings tests as it tests if the send command calls get_account_by_address and not get_account_by_address itself.

pazz:
Quoting Lucas Hoffmann (2017-08-16 15:32:45)
> tests.commands.envelope_test.TestSendCommand.test_get_account_by_address_with_email_message
> should not be moved to the settings tests as it tests if the send command calls
> get_account_by_address and not get_account_by_address itself.


Alright, fine with me. thanks for the feedback!
P

tests/commands/envelope_test.py
```diff
@@ -355,3 +356,41 @@ def test_apply_from_user_and_email(self):
 
         self.assertTrue(env.sign)
         self.assertIs(env.sign_key, mock.sentinel.gpg_key)
+
+
+class TestSendCommand(unittest.TestCase):
```
dcbaker: This is a testing an asynchronous call, I don't think you can't use `unittest` for this. I think you need to use `twisted.trial.unittest` (I'm still slightly fuzzy on testing twisted, I use asyncio and python3 when I want to do asynchronous programming).

Have a look at the commands/global_test.py tests to see the differences. I think that using `@inlineCallbacks` is much easier than building a `deferred` to test with manually.

alot/account.py
```diff
@@ -214,6 +214,9 @@ def errb(failure):
             logging.error(failure.value.stderr)
             raise SendingMailFailed(errmsg)
 
+        # make sure self.mail is a string
+        mail = str(mail)
```
dcbaker: This is going to fail spectacularly if mail is a unicode object that contains a non-ascii encodable character.

tests/commands/envelope_test.py
pazz: you are right: I've now seen several examples like this for using mock.. It just feels wrong though :)

tests/commands/envelope_test.py
pazz: with the patch that fixes SendmailAccount's signature we can leave this as it is I believe.

tests/commands/envelope_test.py
lucc: But the idea of `with` is to handle the closing of stuff and not a naming scope. Note that even `for` does the same:
```py
for i in range(100):
    pass
print(i)
```

tests/commands/envelope_test.py
pazz: Oh I didn't think about that!
It seems that the tests you refer to actually just wrap the test-methods in a `@inlineCallbacks`, which makes them return deferreds themselves. I will add those here as well, this way we are at least consistent.

As a side-note: I wouldn't mind at all dropping twisted for asyncio, which is even part of the standard lib in python3 I think. But making such a big change should be topic for another time..

pazz:
Oh I see. makes sense.

pazz:
I don't understand why those newly decorated tests fail.

tests/commands/envelope_test.py
dcbaker: Agreed on dropping twisted. Note also that those classes use `twisted.trial.unittest` not `unittest`

dcbaker:
I think because `TagCommand.apply`, `AttachCommand.apply`, and `SignCommand.apply` aren't async methods, they're synchronous. I think you can use `trial.unittest` as the test class on them still, but I don't think you should decorate them with `@inlineCallbacks`

pazz:
Oh OK, thanks @dcbaker ! I also forgot to use `yield cmd.aaply(ui)` instead of just calling this method in the tests. This makes the test-method into a generator if I'm not mistaken, which is then compatible with the inlineCallback.

.oO(I wonder why the code after the yield gets executed during the tests at all. I must be because `twisted.unittest` will actually exhaust this generator...)

dcbaker:
Yeah, trial does something to ensure that each asynchronous test has exhausted before reporting the result. I don't really understand trial, just enough to write tests. It feels really wonky to me, I'm used to asyncio, where you can have multiple event loops and can restart them after stopping. And you have the `run_until_complete` method, which makes it easier for me to reason about how they work.

pazz:
I believe I've addressed your requests. Please have another look, this bug is quite annoying and I'd like to ship a fix in the upcoming release.

dcbaker:
@pazz I don't see an updated, did you forget to push?

pazz:
sorry, I overlooked some of your comments. will be back soon :)

tests/commands/envelope_test.py
pazz: OK, I use `twisted.trial.unittest` now.

alot/account.py
pazz: I'm not sure how this is worse than [what we did before](https://github.com/pazz/alot/pull/1115/commits/f67d3ac2bca0dd0dda6d5998ed1131d5b383f995#diff-feec6663d333aaa89ed596841b6acc19L242).
`mail` is, as assumed in the function docstring, either a `str` already, in which this does nothing. Or it is an `email.message`, in which case the email lib should do the right thing..

lucc:
My comments have been addressed. So I will approve this now. When the twisted unittest stuff (and whatnot?) is done dcbaker can unblock it.

alot/account.py
dcbaker: I'm doubtful that the email lib in python2 does the right thing. However, you are correct that this is equivalent to the code it's replacing, so it's not a regression. Handling unicode correctly in python2 is hard, so it's probably fine to leave it until python3 arrives.

PR #1126, opened 2017-08-19
Some small changes that are py2 and py3 compatible

lucc:
This just replaces some code that only works with py2 with a variant that works both with py2 and py3.  It should not break anything but at the same time will help us on the way to py3 (also if we drop py2 later).  With these changes 79 tests can already be run with py3 and 42 pass (many still fail because of string and unicode stuff or can not even be run because of import errors).

Triggered by #1116.

If you like this PR I also have some commits which add compatible imports with a `try: import py2stuff; except ImportError: import py3stuff` pattern.  Should I add them here. (They are different in that they have to be touched again if we drop py2. These changes here don't have to be touched again.)

### DIFF
--- a/alot/completion.py
+++ b/alot/completion.py
@@ -437,10 +437,10 @@ class CommandCompleter(Completer):
                                                                    localpos)
 
                         # prepend 'set ' + header and correct position
-                        def f((completed, pos)):
+                        def f(completed, pos):
                             return ('%s %s' % (header, completed),
                                     pos + len(header) + 1)
-                        res = [f(r) for r in res]
+                        res = [f(c, p) for c, p in res]
                         logging.debug(res)
 
                 elif self.mode == 'envelope' and cmd == 'unset':
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -863,7 +863,7 @@ class ComposeCommand(Command):
                 self.envelope.sign_key = account.gpg_key
             else:
                 msg = 'Cannot find gpg key for account {}'.format(account.address)
-                logging.warn(msg)
+                logging.warning(msg)
                 ui.notify(msg, priority='error')
 
         # get missing To header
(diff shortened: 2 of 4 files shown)


lucc:
PS: The other commits with import stuff I talked about are [here](https://github.com/lucc/alot/tree/py3/small-steps) and the test run for these is [here](https://travis-ci.org/lucc/alot/builds/266173558).

alot/completion.py
```diff
@@ -437,10 +437,10 @@ def complete(self, line, pos):
                                                                    localpos)
 
                         # prepend 'set ' + header and correct position
-                        def f((completed, pos)):
+                        def f(completed, pos):
                             return ('%s %s' % (header, completed),
                                     pos + len(header) + 1)
-                        res = [f(r) for r in res]
+                        res = [f(*r) for r in res]
```
dcbaker: how about `[f(c, p) for c, p in res]`?

dcbaker:
oh, and the logging changes are good :)

alot/completion.py
lucc: even better.

lucc:
I understand your comments about iter* and xrange. I was looking at it not from a semantical point of view but from a syntactical: Just for the fun of it I tried to run the tests with python3 and had a look what could be changed in a py2 compatible way to make as many of these tests pass. But I am also fine with keeping these commits of a little while longer.

Do you think commits with stuff like
```py
try:
    from urlparse import urlparse
except ImportError:
    from urllib.parse import urlparse
```
are a good thing?

lucc:
@dcbaker  I removed the commits you mentioned. I don't like the idea of switching to the view* variants as they are (even if semantically) also not syntactically compatible with python3. So how about we merge it like this and have a discussion about the other things in a new PR?

dcbaker:
@lucc, I thought the plan was to have a pure py3k code base after the transition rather than a hybrid one, at which point I don't think the try/except for imports are useful. 2to3 knows about all of those changes a does a good job fixing them.

PR #1128, opened 2017-08-19
polish the docs

pazz:
this prettifies the generated man-page and cli-usage docs.
It should be nice to have this (or similar) as part of the upcoming release.


docs/source/usage/cli_commands.rst
```diff
@@ -0,0 +1,14 @@
+search
+    start in a search buffer using the querystring provided as
+    parameter. See also the SEARCH SYNTAX section of notmuch(1)
```
lucc: On Arch Linux this is actually `notmuch-search-terms(7)`. Can any body verify this on another system? It also seems to be the case for the notmuch source distribution.

pazz: yes, on debian as well. thanks (done).

PR #1129, opened 2017-08-19
py3k small fixes

dcbaker:
Here's some additional py3k readying patches that do some more user/performance transparent changes in addition to those that lucc sent. These are mostly just syntactic or moved/renamed interface changes. The only thing that is significant, is the addition of the `__hash__` protocol to `TagWidget`. When implementing the `__eq__` protocol in python 2.x `__hash__` isn't required, but it is in python 3. After implementing it I noticed that the search buffer was much slower than it was before, so the hash is cached and not calculated on each walk, which makes it as fast as it was before. It's possible that we either need to make do some of the calculation asynchronously or cache them to make the interface a little faster.

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -146,8 +146,8 @@ class SaveCommand(Command):
                 ui.apply_command(globals.FlushCommand())
                 ui.apply_command(commands.globals.BufferCloseCommand())
             except DatabaseError as e:
-                logging.error(e.message)
-                ui.notify('could not index message:\n%s' % e.message,
+                logging.error(e)
+                ui.notify('could not index message:\n%s' % e,
                           priority='error',
                           block=True)
         else:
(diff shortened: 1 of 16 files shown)


alot/commands/envelope.py
```diff
@@ -146,8 +146,8 @@ def apply(self, ui):
                 ui.apply_command(globals.FlushCommand())
                 ui.apply_command(commands.globals.BufferCloseCommand())
             except DatabaseError as e:
-                logging.error(e.message)
-                ui.notify('could not index message:\n%s' % e.message,
+                logging.error(str(e))
+                ui.notify('could not index message:\n%s' % str(e),
```
lucc: I think logging automatically calls str on its argument. At least for string interpolation it is not neccessary.

alot/settings/manager.py
```diff
@@ -123,7 +123,7 @@ def read_config(self):
                     except ConfigError as e:
                         logging.warning(
                             'Theme file %s failed validation: %s',
-                            themestring, str(e.message))
+                            themestring, str(e))
```
lucc: string interpolation

alot/commands/envelope.py
dcbaker: If it's implemented the way I think it is then it will. I'll run a quick test to ensure that you're right, but I'm pretty sure you are. I'll update

PR #1133, opened 2017-08-21
Fix #1132 by reverting part of fa3dd1b04567.

### LINKED ISSUE #1132: Can not `:send` mail,  TypeError: 'unicode' does not have the buffer interface
Software Versions
-----------------

Python version: 2.7.13

Alot version: master+lucc/fix/hash (20dab95640592bd4f539a8a781f0c80c21cc5708)

Observed behavior
-----------------

I compose a mail, when I `:send` it I get a notification for this exception:

```
DEBUG:__init__:ARGS: [u'send']
DEBUG:__init__:cmd parms {}
DEBUG:envelope:signing plaintext: Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: quoted-printable

hans

  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 653, in _runCallbacks
  File "/home/luc/vcs/alot/alot/ui.py", line 679, in call_apply
  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 1532, in unwindGenerator
  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 1386, in _inlineCallbacks
    result = g.send(result)
  File "/home/luc/vcs/alot/alot/commands/envelope.py", line 217, in apply
    self.mail = email_as_string(self.mail)
  File "/home/luc/vcs/alot/alot/helper.py", line 611, in email_as_string
    g.flatten(mail)
  File "/usr/lib/python2.7/email/generator.py", line 83, in flatten
    self._write(msg)
  File "/usr/lib/python2.7/email/generator.py", line 115, in _write
    self._write_headers(msg)
  File "/usr/lib/python2.7/email/generator.py", line 142, in _write_headers
    print >> self._fp, '%s:' % h,
exceptions.TypeError: 'unicode' does not have the buffer interface
```

Expected behavior
-----------------

Mail should be send with the external command.

Further notes
-------

@dcbaker it looks like this is a bug in the recent changes about StringIO and BytesIO. I will try to look into it but if you know what is going on feel free to send a fix.

lucc:
The test in the first commit shows the bug.  We can also find another fix. But this was the fastest solution that came to my mind.  And I think we should try to unbreak master.

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -10,6 +10,7 @@ from datetime import timedelta
 from datetime import datetime
 from collections import deque
 from io import BytesIO
+from cStringIO import StringIO
 import logging
 import mimetypes
 import os
@@ -606,7 +607,7 @@ def email_as_string(mail):
     :param mail: email to convert to string
     :rtype: str
     """
-    fp = BytesIO()
+    fp = StringIO()
     g = Generator(fp, mangle_from_=False, maxheaderlen=78)
     g.flatten(mail)
     as_string = RFC3156_canonicalize(fp.getvalue())
--- a/tests/helper_test.py
+++ b/tests/helper_test.py
@@ -20,6 +20,7 @@
 from __future__ import absolute_import
 
 import datetime
+import email
 import errno
 import random
 import unittest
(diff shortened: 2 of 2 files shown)


pazz:
Thanks for this. I agree that we should un-break master as soon as possible and the test is a good idea!
For some reason i cannot add a review here..

lucc:
Maybe because I requested @dcbaker to do the review as he was the one how authored fa3dd1b04567?

alot/helper.py
```diff
@@ -10,6 +10,7 @@
 from datetime import datetime
 from collections import deque
 from io import BytesIO
+from cStringIO import StringIO
```
dcbaker: Sorry I didn't get to this sooner, I was taking a vacation day yesterday.

Could we use StringIO.StringIO instead of the cStringIO variant which can't represent unicode that isn't also ascii (utf-8 is a superset of base ascii)?

dcbaker:
actually, since this is a pretty serious bug, I went ahead and merged this. I'll have another go at using the `io` classes later (unless someone else beats me to it.)

PR #1134, opened 2017-08-22
additional tests for the helper module

dcbaker:
Mostly this is just more tests and a little bit of test framework (the addition of an expected_failure decorator for `twisted.trial.unittest based tests`). There is one small bug fix that stops `call_cmd_async` from polluting the global environment when called.

### DIFF
--- a/tests/helper_test.py
+++ b/tests/helper_test.py
@@ -18,16 +18,20 @@
 """Test suite for alot.helper module."""
 
 from __future__ import absolute_import
-
 import datetime
 import email
 import errno
+import os
 import random
-import unittest
 
 import mock
+from twisted.trial import unittest
+from twisted.internet.defer import inlineCallbacks
+from twisted.internet.error import ProcessTerminated
 
 from alot import helper
+
+from . import utilities
 
 # Descriptive names for tests often violate PEP8. That's not an issue, users
 # aren't meant to call these functions.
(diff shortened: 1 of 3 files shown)


dcbaker:
I really like working with the 3.5 asyncio "async def/await" keywords. They're like inlineCallbacks and much easier to understand IMHO. There is a twisted reactor that can be used as an asyncio event loop, which might make for a nice way to transition after we get to python 3 :)

tests/helper_test.py
```diff
@@ -420,3 +424,45 @@ def test_gt_maxlen(self):
...
+        with mock.patch.dict(os.environ, {}, clear=True):
+            yield helper.call_cmd_async(['echo', '-n', 'foo'],
+                                        env={'foo': 'bar'})
+            self.assertEqual(os.environ, {})
+
+    @inlineCallbacks
+    def test_command_fails(self):
+        with self.assertRaises(ProcessTerminated) as cm:
+            # NOTE: pyton is not a typo
+            yield helper.call_cmd_async(['pyton'])
```
lucc: Maybe we can use a more obvious "command_that_does_not_exist_(hopefully:-)".

dcbaker: Sure, pyton was just easy to type :)

dcbaker:
The end of that commit message made no sense, so I replaced it. I also made the command much more explicitly non-existant.

PR #1136, opened 2017-08-23
more  tests + cleanups

dcbaker:
This is even more cleanups and tests for the db/utils module. With these tests we have 98% coverage of the module. This also removes some dead code from the crypto handling (that code was for pygpgme, but isn't relevant for gpg), and a bug in the displaying of signatures if the key isn't in the users keyring.

### DIFF
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -45,6 +45,8 @@ def add_signature_headers(mail, sigs, error_msg):
                       string indicating no error
     '''
     sig_from = u''
+    sig_known = True
+    uid_trusted = False
 
     if isinstance(error_msg, str):
         error_msg = error_msg.decode('utf-8')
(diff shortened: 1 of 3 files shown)


alot/db/utils.py
```diff
@@ -298,16 +298,17 @@ def extract_headers(mail, headers=None):
...
 
 def extract_body(mail, types=None, field_key='copiousoutput'):
-    """
-    returns a body text string for given mail.
-    If types is `None`, `text/*` is used:
-    The exact preferred type is specified by the prefer_plaintext config option
-    which defaults to text/html.
+    """Returns a body text string for given mail.
+
+    If types is `None`, `text/*` is used the exact preferred type is specified
```
pazz: you've dropped the colon after 'used' here..
I must say that I'm also not a fan of the the original docstrings here anyways

dcbaker: I just went ahead and rewrote the docstring, so now it actually references the config option it checks and is hopefully a little clearer.

pazz: much better now, thanks.

PR #1138, opened 2017-08-23
commands/globals: Only set stdin to a pipe if there is data

### LINKED ISSUE #1137: Editor fails if stdin is None
In 3ad6ebf3b8d1e4835e4bb62be9dbc08fc48518ac I introduced a bug that causes (at least neovim, presumably other editors) to fail if stdin is None.

dcbaker:
Otherwise the editor will crash.

I can't figure out for the life of me to test this. I thought that I
would be able to do something like detect if stdin is a tty or something
else, but that doesn't seem to work.

Fixes #1137

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -268,7 +268,7 @@ class ExternalCommand(Command):
         def thread_code(*_):
             try:
                 proc = subprocess.Popen(self.cmdlist, shell=self.shell,
-                                        stdin=subprocess.PIPE,
+                                        stdin=subprocess.PIPE if stdin else None,
                                         stderr=subprocess.PIPE)
             except OSError as e:
                 return str(e)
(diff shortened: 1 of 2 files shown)


lucc:
In the shell you can check if stdin is a terminal with `test -t 0`. Does that help? Can you set the editor to something like `test -t 0`?

dcbaker:
that might work. I'd really like to test this at a lower level than the editor invocation level. I'll see about `test -t 0` for unit tests.

dcbaker:
Okay, tests added. It turns out that pytest was causing my problem, it mucks with stdin since it assumes that no one would want to test anything on stdin.

dcbaker:
@lucc I'm going to go ahead and push this, but if you think the tests need more work let me know and I'll rework them.

PR #1148, opened 2017-08-31
pep8 fixes

pazz:
This mostly shortens lines down to <=79 chars and fixes some other small
things I found using the pep8 tool.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -78,7 +78,8 @@ class ExitCommand(Command):
 
         if ui.db_was_locked:
             msg = 'Database locked. Exit without saving?'
-            if (yield ui.choice(msg, msg_position='left', cancel='no')) == 'no':
+            response = yield ui.choice(msg, msg_position='left', cancel='no')
+            if response == 'no':
                 return
             ui.exit()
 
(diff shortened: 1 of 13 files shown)


alot/widgets/search.py
```diff
@@ -115,7 +115,8 @@ def pad(string, shorten=None):
             if self.thread:
                 fallback_normal = struct[name]['normal']
                 fallback_focus = struct[name]['focus']
-                tag_widgets = sorted(TagWidget(t, fallback_normal, fallback_focus)
+                tag_widgets = sorted(TagWidget(t, fallback_normal,
+                                               fallback_focus)
```
dcbaker: This is really hard to read. What about
```python
tag_widgets = sorted(
    TagWidget(t, fallback_normal, fallback_focus)
    for t in self.thread.get_tags())
```

alot/commands/globals.py
```diff
@@ -78,7 +78,9 @@ def apply(self, ui):
 
         if ui.db_was_locked:
             msg = 'Database locked. Exit without saving?'
-            if (yield ui.choice(msg, msg_position='left', cancel='no')) == 'no':
+            if (yield ui.choice(msg,
+                                msg_position='left',
+                                cancel='no')) == 'no':
```
dcbaker: I've been thinking about these, they're hard to read, but what if we took it out of the if statement?
```python
response = yield ui.choice(msg, msg_position='left', cancel='no')
if responce == 'no':
    ...
```

alot/widgets/search.py
pazz: ok

alot/commands/globals.py
pazz: works for me

dcbaker:
This looks good to me at this point. I don't know what you want to do about the codeclimate warnings, I looked at them and they all look pretty silly to me.

pazz:
I'll fix the 'newline before bin op' issues and ignore the rest.

PR #1151, opened 2017-09-06
Submit/fix no from header

dcbaker:
I noticed today that I have a draft email which lacks a 'from' header. This is obviously pretty rare, since it's unlikely that you would receive an email without a 'from' header. For drafts we set the 'Sender' header, which can be used to the same effect.

This covers using the 'Sender' as a fallback for 'From', and failing that uses the default account as the from.

### DIFF
--- a/tests/db/message_test.py
+++ b/tests/db/message_test.py
@@ -78,7 +78,6 @@ class TestMessage(unittest.TestCase):
             MockNotmuchMessage({'From': '"User Name" <user@example.com>'}))
         self.assertEqual(msg.get_author(), ('User Name', 'user@example.com'))
 
-    @unittest.expectedFailure
     def test_get_author_sender(self):
         """Message._from is populated using the 'Sender' header when no 'From'
         header is present.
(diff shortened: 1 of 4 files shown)


tests/db/message_test.py
```diff
@@ -0,0 +1,100 @@
...
+        return self.mock_filename
+
+    def get_tags(self):
+        return self.mock_tags
+
+
+class TestMessage(unittest.TestCase):
+
+    def test_get_from_email_only(self):
+        """From is populated using the 'From' header when only an email address
```
pazz: shouldn't this docstring tell us what the method tests? same for the other docstrings in this file..

dcbaker:
I can drop the from me change. Alot does set the 'From' header, it appears outlook is the culprit for not setting it. I don't think we can assume that every draft is from alot. But I'll do some reworking of the series and send an update.

dcbaker:
Okay, I've changed it up a bit to address pazz's comments.
- The docstrings of the tests have been updated.
- The `Address` class has been updated to handle comparisons to an empty string.
- The default account is now only used if the 'draft' tag is also present, otherwise the from is set to 'unkown'

dcbaker:
I should add that alot does set the 'From' in drafts (which is about the only way I can imagine this happening), but not all clients do. Since alot my encounter drafts created by other MUAs we should try to be defensive about this.

pazz:
much better, thanks. So what exactly fails when you try to display draft messages without From-header on master? could you paste a trace? Do you think this approach here is better than making alot properly display those deafts?

dcbaker:
The traceback in #1152 is the same as I get here.

PR #1155, opened 2017-09-11
github: add notmuch version to issue template

dcbaker:
A recent bug made me think that maybe the notmuch was out of date. It would be good to include that information in a bug report.


NoctuaNivalis:
Yeah, I didn't think of including it myself.

PR #1158, opened 2017-09-22
crypto: Handle message that is encrypted, but the signature is invalid

### LINKED ISSUE #1157: Cannot read encrypted messages if it is signed and the signature is invalid
Software Versions
-----------------

Python version: 2.7.13

Notmuch version: 0.25

Alot version (either a number or sha for master): c6b8efc1f6784d757c601036e0ff982487fe0d29
Observed behavior
-----------------

Attempting to read a signed + encrypted message when I don't have the public key for the signer.

Opening the thread crashes with an unhandled 'BadSignatures' exception. even once that it handled it still simply render the message as "malformed", even though it's not. downloading the signing key fixes this

Expected behavior
-----------------

Not crashing, showing the message and 'invalid signature'.

dcbaker:
One case of this would be not having the public key of the signer. If
the verification of the signatures fails, then use the signatures from
the error, and try to redecrypt without verification.

I have no tests yet, and this probably deserves tests.

Fixes #1157

### DIFF
--- a/alot/crypto.py
+++ b/alot/crypto.py
@@ -202,12 +202,15 @@ def decrypt_verify(encrypted):
     """
     ctx = gpg.core.Context()
     try:
-        (plaintext, _, verify_result) = ctx.decrypt(encrypted, verify=True)
+        plaintext, _, verify_result = ctx.decrypt(encrypted, verify=True)
+        sigs = verify_result.signatures
     except gpg.errors.GPGMEError as e:
         raise GPGProblem(str(e), code=e.getcode())
-    # what if the signature is bad?
+    except gpg.errors.BadSignatures as e:
+        plaintext, _, _ = ctx.decrypt(encrypted, verify=False)
+        sigs = e.result.signatures
 
-    return verify_result.signatures, plaintext
+    return sigs, plaintext
 
 
 def validate_key(key, sign=False, encrypt=False):


alot/crypto.py
```diff
@@ -202,12 +202,15 @@ def decrypt_verify(encrypted):
...
     ctx = gpg.core.Context()
     try:
-        (plaintext, _, verify_result) = ctx.decrypt(encrypted, verify=True)
+        plaintext, _, verify_result = ctx.decrypt(encrypted, verify=True)
+        sigs = verify_result.signatures
     except gpg.errors.GPGMEError as e:
         raise GPGProblem(str(e), code=e.getcode())
-    # what if the signature is bad?
+    except gpg.errors.BadSignatures as e:
+        plaintext, _, _ = ctx.decrypt(encrypted, verify=False)
```
lucc: should we catch the same GPGMEError as in line 207 for this?

dcbaker: I thought about it, and I don't think we need to. I *think* that GPGMEError will be raised before BadSignatures if the content is invalid. But I'm not completely sure. That's one of the reasons I think tests are needed.

meskio:
I've tested the code (merged on top of master) and it solves the issue for me. Thanks.

varac:
@dcbaker This looks like a forgotten pearl behind a cupboard. Please merge, I'm running into the same issue.

dcbaker:
@pazz, @lucc, can one of you review this or merge it? I can't since it's not reviewed.

dcbaker:
Thanks!

PR #1160, opened 2017-09-25
Fix #1140

### LINKED ISSUE #1140: Regression: Encrypt to self not picked up from gpg.conf
Software Versions
-----------------

Python 2.7, Alot master (but I guess since the gpg backend switch) 

Observed behavior
-----------------

If I `:encrypt` a message to x@y.z (but not explicitly to me) it is only encrypted to x@y.z (I checked the resulting file in the maildir both after I did `:save` or `:send` it).

Expected behavior
-----------------

The message should be encrypted to x@y.z *and* also to the key given in my `$GNUPGHOME/gpg.conf` file as argument to `encrypt-to`. This was previously the case and is also noted in the docs, in the last "Note" [over here](http://alot.readthedocs.io/en/latest/usage/crypto.html).

possible solutions (in order of preference)
--------
1. make alots encyption backend pick up the value from the config file again
2. add a new option to alot to encrypt to an accounts gpg_key
3. change the docs + manually encrypt to self (temporary possible)

meskio:


meskio:
@dcbaker I think I followed your intention. How can I improve it?

alot/commands/utils.py
```diff
@@ -46,6 +48,19 @@ def set_encrypt(ui, envelope, block_error=False, signed_only=False):
...
+            try:
+                acc = settings.get_account_by_address(envelope['From'])
+                if acc.encrypt_to_self:
+                    if acc.gpg_key:
+                        logging.debug('encrypt to self: %s', acc.gpg_key.fpr)
+                        envelope.encrypt_keys[acc.gpg_key.fpr] = acc.gpg_key
+                    else:
+                        logging.debug('encrypt to self: no gpg_key in account')
+            except NoMatchingAccount:
+                logging.debug('encrypt to self: o account found')
```
dcbaker: type: o -> noe

alot/defaults/alot.rc.spec
```diff
@@ -351,6 +351,14 @@ thread_focus_linewise = boolean(default=True)
         #           1.0, please move to `all`, `none`, or `trusted`.
         encrypt_by_default = option('all', 'none', 'trusted', 'True', 'False', 'true', 'false', 'Yes', 'No', 'yes', 'no', '1', '0', default='none')
 
+        # If this is true when encrypting a message it will also be encrypted
+        # with the key defined for this account.
+        #
+        # .. warning::
+        #   
+        #    Before 0.6 this was controlled via gpg.conf.
```
pazz: shouldn't this be 0.7?

pazz:
looks all good to me as well

alot/defaults/alot.rc.spec
meskio: Since 0.6 gpg.conf doesn't work for this. It's true that I guess 0.7 will be the first version with `encrypt_to_self`.

pazz: ok i see. fine then :)

Issue #987, opened 2017-01-19
Missing warning or notice on missing mailcap command


lucc:
Put this in your mailcap file:
```mailcap
text/html; command_to_render_html_not_found %s; copiousoutput;
```
and then open alot and try to read a mail that is html. For me I get the header in the thread view but the body is empty (zero lines). 

I would prefer it if alot would display a warning or notification if the command can not be found. Maybe we should only display the error once when the thread buffer is opened as it might be a little intrusive on long threads with lots of html mail.

The mailcap entry is executed with `alot.commands.globals.ExternalCommand` and there already is a code path for failed commands in [line 256](https://github.com/pazz/alot/blob/master/alot/commands/globals.py#L256) but it seems to have no effect or is not reached in this case.

dcbaker:
There's some pretty strange handling of `subprocess.Popen` in there, since we're calling both `.wait` and `.communicate`, which is strange.

dcbaker:
This problem is in `alot.utils.extract_body` which has no access to the UI (since the HTML conversion isn't done by calling ExternalCommand. I looked over it a bit and there wasn't an immediately obvious fix.

pazz:
close due to #1187. Although the solution is not ideal, the original issue (here) is fixed.

PR #1163, opened 2017-09-29
Warn people of missing text/html mailcap entry.

pazz:


pazz:
This is a RFC and not ready yet.
Interestingly, it shows that alot will look up a mailcap entry also for mails with "multipart/signed"  parts...

pazz:
OK, this is good to go from my side now.

related: #987

tests/db/utils_test.py
```diff
@@ -723,8 +723,6 @@ def test_prefer_plaintext(self):
         self.assertEqual(actual, expected)
 
     # Mock the handler to cat, so that no transformations of the html are made
-    # making the result non-deterministic
```
dcbaker: depending on the handler it may or may not make transformations to the HTML, which would result in the test failing. I think that it's fine to remove the comment though.

pazz: oh i see. you mean the prefer-plaintext option. I suppose that it is safe to remove the mock here because we'd run test mostly in a sandbox which uses the default values. But to be sure ill leave it (I'll just reformulate the comment)

pazz:
I found a problem with this series: It complains about missing mailcap entries for `image/jpeg` in certain mails, where those appear (as part of the html I suppose) but are not attachments.. I'll investigate :)

pazz:
Here is an example part of a mail that causes problems.

```
--Apple-Mail=_8185E02D-6CAC-4405-AAE2-4886BC5C90DA
Content-Disposition: inline;
?filename="2017 leaflet.pdf"
Content-Type: application/pdf; x-unix-mode=0644; name="2017 leaflet.pdf"
Content-Transfer-Encoding: base64
...
--Apple-Mail=_8185E02D-6CAC-4405-AAE2-4886BC5C90DA
```


The `Content-Disposition: inline;` causes a mailcap lookup for `application/pdf` that for some reason is not matched by my entry `application/pdf; zathura %`.

This is the log:
```
DEBUG:utils:unquoted header: |2017 leaflet.pdf|
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 321, in addCallback
    callbackKeywords=kw)
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 310, in addCallbacks
    self._runCallbacks()
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 653, in _runCallbacks
  File "/home/pazz/projects/alot/alot/ui.py", line 679, in call_apply
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/home/pazz/projects/alot/alot/commands/search.py", line 40, in apply
    sb = buffers.ThreadBuffer(ui, self.thread)
  File "/home/pazz/projects/alot/alot/buffers.py", line 341, in __init__
    self.rebuild()
  File "/home/pazz/projects/alot/alot/buffers.py", line 369, in rebuild
    self._tree = ThreadTree(self.thread)
  File "/home/pazz/projects/alot/alot/widgets/thread.py", line 355, in __init__
    accumulate(msg)
  File "/home/pazz/projects/alot/alot/widgets/thread.py", line 334, in accumulate
    self._message[mid] = MessageTree(msg, odd)
  File "/home/pazz/projects/alot/alot/widgets/thread.py", line 175, in __init__
    self._maintree = SimpleTree(self._assemble_structure())
  File "/home/pazz/projects/alot/alot/widgets/thread.py", line 208, in _assemble_structure
    bodytree = self._get_body()
  File "/home/pazz/projects/alot/alot/widgets/thread.py", line 241, in _get_body
    bodytxt = extract_body(self._message.get_email())
  File "/home/pazz/projects/alot/alot/db/utils.py", line 355, in extract_body
    raise NoMailcapEntry(ctype)
alot.settings.errors.NoMailcapEntry: Could not render content of type "application/pdf" due to a missing mailcap entry.
```

I suppose we could hardcode a workaround for the apple mailer but I don't want to :)
Another possible solution would be to hat a sort of whitelist of mimetypes that will be turned into attachments if no mailcap entry was found, and then *not* put `text/html` on that list.

pazz:
closed in favor of #1187

PR #1170, opened 2017-10-20
RFE: implement folder tree buffer

mschaaf:
This PR is for bringing the code from PR #757 to the latest master code base.
Thanks to @TomasTomecek for creating the first PR. I like it so much that I want to start another try to bring it to the main code base.

# Contains

discover maildir structure
query notmuch for # of unread msgs in folders
display as tree
theming:
empty folders
not empty folders
focused folders
configuration
path to maildir
blacklisted folders
enable (un)folding
implement refresh

# Bugs
(these should be addressed before merging)

[] next unread folder jumps not to all unread folders

RFEs
(these may be addressed in subsequent PRs)

completion for folders when searching
new command: pre-complete searching for selected folder
figure out something better for status line of folders buffer (e.g. # of new messages in favorite folders)


pazz:
Hey and welcome!
It could be useful to add a minimal guide on how to test this. For instance, does it run as is from an empty config?

regarding the decoding utility: I'd be really surprized if this was nut properly solved in some imap library or so. We should not reinvent the wheel for fear of adding dependencies.

Needless to say this should not trigger many new issues on codeclimate nor failed tests.. So that would need to be fixed before any of this goes into the master branch.

Thanks for your work so far and happy hacking :)

mschaaf:
@pazz Thanks for your answer. I will see if I can find a small library to decode, what I found so far I didn't get working.

pazz:
I've quickly tested this for a while and here are some thoughts.
It looks nice generally. I noticed that it displays my folders in a slightly unexpected fashion:

* "gmail/INBOX" is displayed at level one of the tree, that is at the same depth as "gmail", whereas other subfolders are as expected at level 2.
* the root of the tree carries my first account's email address as label. I have more than one account. What's so special about that one?
* those folders that have subfolders are collapsible nodes. I'd expect that hitting "select" on them would toggle the collapse status. But it triggers a new search buffer to open, despite the fact that all such folders are not maildirs themselves, so these searches necessarily return zero result sets..


I think the `mark_as_read` command is really out of line with other buffer's commands. This should really be a `tag` and `untag` command and work just as the respective commands in thread (or search) buffers.

Finally, please squash down commits for easier review (for instance all "fix codeclimate issues", at least the ones addressing the same issue should go into one commit). It'd help  if you could split the initial big commit into small logical commits as well because otherwise it's really hard to review systematically.

mschaaf:
> those folders that have subfolders are collapsible nodes. I'd expect that hitting "select" on them would toggle the collapse status. But it triggers a new search buffer to open, despite the fact that all such folders are not maildirs themselves, so these searches necessarily return zero result sets..

Not in my case. In my setup such folders could contain and contain mail. They are maildir folders. So changing the behavior for these folders would be inconsistent in my opinion.

mschaaf:
> "gmail/INBOX" is displayed at level one of the tree, that is at the same depth as "gmail", whereas other subfolders are as expected at level 2.

This is the same in my setup. It reflects how mbsync syncs my folders from gmail. So it is what notmuch returns.

mschaaf:
> the root of the tree carries my first account's email address as label. I have more than one account. What's so special about that one?

I don't know. Maybe the code did not handle a multi account setup.
How is the setup in notmuch and alot for the mutli account, so that I can test it.

mschaaf:
> I think the mark_as_read command is really out of line with other buffer's commands. This should really be a tag and untag command and work just as the respective commands in thread (or search) buffers.

You mean implementing it as a "tag --all" command, right?

mschaaf:
I removed the command mark_as_read for me it is important and it is not necessary for the folder view feature. So it can be brought in with a subsequent PR.

Also, I found the issue with the inbox folder. I didn't recognized it because it was on the expected level, but saw in tests that any sub folder with that name was moved up. So I removed that code.

lucc:
I had to apply this patch to start testing this:
```diff
diff --git a/alot/db/folder.py b/alot/db/folder.py
index 73a76775..4e0556c8 100644
--- a/alot/db/folder.py
+++ b/alot/db/folder.py
@@ -24,7 +24,7 @@ class Folder(object):
         self.refresh()
 
     def __str__(self):
-        folder_name = imapUTF7Decode(str(self.folder_name))
+        folder_name = imap_utf7.imapUTF7Decode(str(self.folder_name))
         return " %s [%d/%d]" % (folder_name, self.unread_count, self.all_count)
 
     def __unicode__(self):

```

mschaaf:
@lucc Thanks for the hint. I applied your fix.

pazz:
Quoting Schaaf, Martin (2017-10-22 14:59:38)
>     those folders that have subfolders are collapsible nodes. I'd expect that
>     hitting "select" on them would toggle the collapse status. But it triggers
>     a new search buffer to open, despite the fact that all such folders are not
>     maildirs themselves, so these searches necessarily return zero result
>     sets..
> 
> Not in my case. In my setup such folders could contain and contain mail. They
> are maildir folders. So changing the behavior for these folders would be
> inconsistent in my opinion.


OK fair enough.

pazz:
Quoting Schaaf, Martin (2017-10-22 15:01:17)
>     "gmail/INBOX" is displayed at level one of the tree, that is at the same
>     depth as "gmail", whereas other subfolders are as expected at level 2.
> 
> This is the same in my setup. It reflects how mbsync syncs my folders from
> gmail. So it is what notmuch returns.

I'm not sure I understand. Here, "INBOX" is clearly a subfolder of "gmail" in your filesystem, i.e., after synchronization.
They should be in child/parent relation in the tree of folders you display, and not be siblings.

pazz:
Quoting Schaaf, Martin (2017-10-22 15:04:37)
>     the root of the tree carries my first account's email address as label. I
>     have more than one account. What's so special about that one?
> 
> I don't know. Maybe the code did not handle a multi account setup.
> How is the setup in notmuch and alot for the mutli account, so that I can test
> it.

notmuch has no notion of accounts as far as I know. Alot does, and you
can get them using ``settings.get_account()`:
https://alot.readthedocs.io/en/latest/api/settings.html#alot.settings.manager.SettingsManager.get_accounts

pazz:
Quoting Schaaf, Martin (2017-10-22 15:09:58)
>     I think the mark_as_read command is really out of line with other buffer's
>     commands. This should really be a tag and untag command and work just as
>     the respective commands in thread (or search) buffers.
> 
> You mean implementing it as a "tag --all" command, right?

I mean this new buffer should have commands called "tag", "untag" and "retag", similar to search and tread buffers.
The semantic of a "tag" command could be to tag all messages in the folder under the cursor.
Of course these commands would also eat a (list of) tag names to use, so that in particular "unread" is just one possible parameter and not hard-coded.

pazz:
Quoting Schaaf, Martin (2017-10-22 15:31:58)
> I removed the command mark_as_read for me it is important and it is not
> necessary for the folder view feature. So it can be brought in with a
> subsequent PR.
> 
> Also, I found the issue with the inbox folder. I didn't recognized it because
> it was on the expected level, but saw in tests that any sub folder with that
> name was moved up. So I removed that code.

OK great, I'm just now catching up with old messages :)

pazz:
So I have a problem with the initial commit that introduces all kinds of bulk that I think is not necessary. In particular this notion of a "database account", under `db.accounts` which in fact represents a directory. Why not simply ask notmuch what the root directory of its database is, and use the filesystem? Check out for example https://github.com/pazz/urwidtrees/blob/master/docs/examples/example4.filesystem.py how to build a tree (walker) for this. Obviously, you'd only be interested in directories, not files..

alot/db/folder.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+
+from alot.utils import imap_utf7
+
+
+class Folder(object):
```
pazz: I really doubt we need such a class and could instead do with strings representing paths relative to $NOTMUCH_ROOT. The difference to threads or messages, for which we do have such wrappers, is that those are concepts that directly appear in libnotmuch. Folders appear just as part of the query syntax for finding messages/threads. I think we should simply hand the `FolderWidget` that appears in the folder tree a reference to the database manager, so that it can query and display the result of `notmuch count folder:$MYPATH`. Keep it simple.

pazz:
Reading up on [my own old review of this](https://github.com/pazz/alot/pull/757#issuecomment-153469248), it seems simple enough to get the filesystem structure from notmuch (to build a custom DirectoryTreeWalker), using the API calls mentioned in my post. This does not mean that we need to explicitly represent folders as objects. Again, keeping them as strings and putting the query logic into the widget should suffice IMO.

Issue #1161, opened 2017-09-28
Malformed OpenPGP message: Invocation of gpgme_op_decrypt_verify


mschaaf:
Please ensure that the following information is filled before filing
--------------------------------------------------------------------
Issue exists in version 0.6
Installed with pip install alot
gpgme 1.9.0

Please run alot again with the options `-d debug -l log`.
```
INFO:search:open thread view for thread:000000000005c87c: encryption alot 2
DEBUG:utils:unquoted header: |name <mail>|
DEBUG:utils:unquoted header: |name <mail>|
DEBUG:utils:unquoted header: |name <mail>|
DEBUG:utils:unquoted header: |encryption alot 2|
DEBUG:buffers:Tbuffer: auto remove unread tag from msg?
DEBUG:buffers:Tbuffer: No, cursor on summary
DEBUG:ui:thread command string: " fold *"
DEBUG:__init__:mode:thread got commandline " fold *"
DEBUG:__init__:ARGS: [u'fold', u'*']
DEBUG:__init__:cmd parms {'query': [u'*'], 'visible': False}
DEBUG:thread:matching lines *...
DEBUG:buffers:setting focus to (u'150660789574.30954.11688721033365022327@stimpy', (0,)) 
DEBUG:thread:collapsed True
DEBUG:thread:display_source False
DEBUG:thread:display_all_headers False
DEBUG:thread:display_attachements True
DEBUG:thread:AHT None
DEBUG:thread:DHT <alot.widgets.thread.DictList object at 0x7f84b00dfcd0>
DEBUG:thread:MAINTREE [(<MessageSummaryWidget selectable box/flow widget>, [(<alot.widgets.thread.DictList object at 0x7f84b00dfcd0>, None), (<alot.widgets.thread.TextlinesList object at 0x7f84b00df2d0>, None)])]
DEBUG:ui:thread command string: " unfold tag:unread"
DEBUG:__init__:mode:thread got commandline " unfold tag:unread"
DEBUG:__init__:ARGS: [u'unfold', u'tag:unread']
DEBUG:__init__:cmd parms {'query': [u'tag:unread'], 'visible': True}
DEBUG:thread:matching lines tag:unread...
DEBUG:ui:thread command string: " move last"
DEBUG:__init__:mode:thread got commandline " move last"
DEBUG:__init__:ARGS: [u'move', u'last']
DEBUG:__init__:cmd parms {'movement': [u'last']}
DEBUG:thread:last
[...]

PR #1171, opened 2017-10-23
Don't use __str__ of  error BadSignature

mschaaf:
It seems not to be to handle error messages of different locale than en
this fixes issue #1161

### DIFF
--- a/alot/crypto.py
+++ b/alot/crypto.py
@@ -173,6 +173,22 @@ def encrypt(plaintext_str, keys):
     return out
 
 
+NO_ERROR = None
+
+
+def bad_signatures_to_str(error):
+    """
+    Convert a bad signature exception to a text message.
+    This is a workaround for gpg not handling non-ascii data correctly.
+
+    :param BadSignatures error: BadSignatures exception
+    """
+    return ", ".join("{}: {}".format(s.fpr,
+                                     "Bad signature for key(s)")
+                     for s in error.result.signatures
+                     if s.status != NO_ERROR)
+
+
 def verify_detached(message, signature):
     """Verifies whether the message is authentic by checking the signature.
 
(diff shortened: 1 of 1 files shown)


dcbaker:
does using unicode instead of str work? I'd really like to not reimplement the error message handling of gpg. If using unicode doesn't work, then we probably need to do some encode/decode magic.

mschaaf:
unicode didn't work. IMHO, doing some "encode/decode magic" in alot does not work because the exception is thrown in the gpgme code and not alot.

dcbaker:
more python2 bonghits. I'm sure this works correctly in python3.

Could you change the name of the function to follow PEP8 convention? (all lower case with _ separators) and add a comment to the function explaining that this is a workaround for gpg not handling non-ascii data correctly, and that it may not be necessary for python 3? I'd be willing to merge it with those changes.

mschaaf:
@dcbaker Thanks for your comments. I applied them.

pazz:
I'm late to the party so excuse my silly question here: It seems that you do decoding voodoo basically to properly pass on an error message for a very specific error.
Why not hardcode an appropriate message instead?

mschaaf:
It is now partly hardcoded and the other part is the fingerprint for the bad signature.

mschaaf:
@dcbaker and @pazz What is missing to get this merged?

pazz:
@dcbaker So if I understand correctly, this is a hotfix for python2 that will eventually be removed due to proper unicode handling in python3 yes?

@mschaaf thanks for the changes and sorry for the delay. If you could just squash these two commits into one I'd be happy to merge this. (I know I'm being pedantic but I'm trying to avoid commits that are introducing code that is imediatedly removed/superseded). Thanks!

dcbaker:
@pazz I think so. IIRC gpgme uses unicode strings in python 3, but bytestrings in python 2. Thus things like this.

pazz:
ok great. so I'll just push the squashed version of this then.

mschaaf:
@pazz Ready to merge.

pazz:
cool, thx

PR #1172, opened 2017-10-24
Fix for issue 1164 - Cannot select key for encryption by number keys

mschaaf:
You can ignore the first commit in the branch it is overwritten by the commits that followed this one.

### DIFF
--- a/alot/commands/utils.py
+++ b/alot/commands/utils.py
@@ -90,15 +90,16 @@ def _get_keys(ui, encrypt_keyids, block_error=False, signed_only=False):
                                  signed_only=signed_only)
         except GPGProblem as e:
             if e.code == GPGCode.AMBIGUOUS_NAME:
-                tmp_choices = [k.uids[0].uid for k in
+                tmp_choices = [k.uids[0].uid + " (" + k.fpr + ")" for k in
                                crypto.list_keys(hint=keyid)]
-                choices = {str(i): t for i, t in
-                           enumerate(reversed(tmp_choices), 1)}
-                keyid = yield ui.choice("ambiguous keyid! Which " +
-                                        "key do you want to use?",
-                                        choices, cancel=None)
-                if keyid:
-                    encrypt_keyids.append(keyid)
+                choices = {str(i): t for i, t in enumerate(tmp_choices, 1)}
+                keys_to_return = {str(i): t for i, t in enumerate([k for k in
+                                  crypto.list_keys(hint=keyid)], 1)}
+                choosen_key = yield ui.choice("ambiguous keyid! Which " +
+                                              "key do you want to use?",
+                                              choices, keys_to_return)
+                if choosen_key:
+                    keys[choosen_key.fpr] = choosen_key
                 continue
             else:
                 ui.notify(str(e), priority='error', block=block_error)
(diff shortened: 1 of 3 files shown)


pazz:
please use `git rebase -i` or similar to squash these commits for easier review.

mschaaf:
@pazz , I squashed the commits

mschaaf:
> please split commits so that each commit does exactly one thing:
> while it makes little sense to have lots of small commits that add, then later remove code etc,
> it does makes sense to have say one commit that changes the widget code, another that adds the
> docstring, and yet another that changes the code that uses this widget. Sorry to be pedantic about this :)

I am willing to apply your comments but I had them split in the first place by functional groups like you describe and you wanted me to squash them.

Please see your comment:

> please use git rebase -i or similar to squash these commits for easier review.

I am not sure where this will end ;-) but maybe you could be more clear in what you mean in the first place. Thanks.

pazz:
Sorry if I wasn't clear enough then. What I was referring to were the
several "fix codeclimate comments" commits that minimally amend code
that was added in a previous commit. I would prefer to have these
combined with their respective previous commits so that there is only
one patch to review for each new piece of code.

mschaaf:
@pazz Could you please have a chance to review now. I split the commits again by functional groups.

> can you elaborate why the generalized choice widget is necessary? Why can we not simply collect all possible values and give more options to begin with?

It is necessary because the algorithm runs in an endless loop. Because the choices are not unique and always lead to the same options to select.

pazz:
thanks for the explanation @mschaaf, and the splitting. Much easier to review :)

This leaves my second point from above. I'm not super happy with the API as it is because the naming of the parameters make it less obvious what they are supposed to do, and for my taste the logic begind the `choices_to_return` is slightly strange. But I'm biased of course, and since it's documented I only have aesthetic reservations. I'll leave it up to @lucc and @dcbaker to decide on the API change.

alot/commands/utils.py
```diff
@@ -90,15 +90,16 @@ def _get_keys(ui, encrypt_keyids, block_error=False, signed_only=False):
                                  signed_only=signed_only)
         except GPGProblem as e:
             if e.code == GPGCode.AMBIGUOUS_NAME:
-                tmp_choices = [k.uids[0].uid for k in
+                tmp_choices = [k.uids[0].uid + " (" + k.fpr + ")" for k in
```
dcbaker: I really don't like string concatenation (it's slow on some python implementations, and it's hard to read), I think that `'{}({})'.format(k.uids[0].uid, k.fpr) would be more readable here.

alot/ui.py
```diff
@@ -481,15 +481,18 @@ def clear_notify(self, messages):
             self._notificationbar = None
         self.update()
 
-    def choice(self, message, choices=None, select=None, cancel=None,
-               msg_position='above'):
+    def choice(self, message, choices=None, choices_to_return=None,
+               select=None, cancel=None, msg_position='above'):
```
dcbaker: I'd really prefer to see the new arguments added to the end of the function signature, there is no guarantee that callers aren't using the keyword arguments as positional, and adding the new arguments to the end ensures we don't accidentally break anything (and our test coverage is less than complete).

Issue #788, opened 2015-10-19
toggleable tree-indent in thread view


pazz:
fixed indentation in long threads sucks. add command to disable the ArrowTree decoration in threadbuffers

pazz:
done in #1173

PR #1173, opened 2017-11-05
thread mode message indentation

pazz:
This introduces a long discussed feature for the thread mode:
configurable indentation width of message/replies and accompanying commands.

The new config setting is called `thread_indent_replies`,
and the (thread mode) command is called `indent` and accepts integers, `+` and `-` as parameters.

related: #695 #788 #816


tests/utils/argparse_test.py
```diff
@@ -163,3 +163,18 @@ def test_fifo(self):
             path = os.path.join(d, 'fifo')
             os.mkfifo(path)
             cargparse.optional_file_like(path)
+
+
+class TestIntOrPlusOrMinus(unittest.TestCase):
+    """Tests for the is_int_or_pm validator."""
+
+    def test_int(self):
+        cargparse.is_int_or_pm('5')
```
dcbaker: I think this should be `self.assertTrue(cargparse.is_int_or_pm('5'))`, and for `test_pm` too.

alot/buffers.py
```diff
@@ -367,14 +368,29 @@ def rebuild(self):
...
+        if self._indent_width == 0:
+            A = self._tree
+        else:
+            # we want decoration.
+            bars_att = settings.get_theming_attribute('thread', 'arrow_bars')
+            # only add arrow heads if there is space (indent > 1).
+            heads_char = None
+            heads_att = None
+            if self._indent_width > 1:
+                heads_char = u'\u27a4'
```
dcbaker: feel free to take this or leave this, but if we added `#encoding = utf-8` to the top of the file we could insert the unicode character instead of the escape code here.

tests/utils/argparse_test.py
pazz: oh ok? I honestly just copied the test just above, check e.g. line 159 in that file.
So should we adjust those as well?

alot/buffers.py
pazz: yes, right. I'll do that.

pazz: Just tested this, and it triggers a weird error related to urwid not being able to determine the correct width..Or rather, demanding more than width 1 for 

```
  File "build/bdist.linux-x86_64/egg/urwidtrees/widgets.py", line 68, in _get
  File "build/bdist.linux-x86_64/egg/urwidtrees/lru_cache.py", line 102, in wrapper
  File "build/bdist.linux-x86_64/egg/urwidtrees/widgets.py", line 54, in __getitem__
  File "build/bdist.linux-x86_64/egg/urwidtrees/nested.py", line 88, in get_decorated
  File "build/bdist.linux-x86_64/egg/urwidtrees/nested.py", line 84, in _get_decorated_entry
  File "build/bdist.linux-x86_64/egg/urwidtrees/decoration.py", line 446, in decorate
  File "build/bdist.linux-x86_64/egg/urwidtrees/decoration.py", line 416, in _construct_first_indent
urwidtrees.decoration.NoSpaceError
```

```patch
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -1,3 +1,4 @@
+# encoding=utf-8
 # Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
 # This file is released under the GNU GPL, version 3 or a later revision.
@@ -380,7 +381,7 @@ class ThreadBuffer(Buffer):
             heads_char = None
             heads_att = None
             if self._indent_width > 1:
-                heads_char = u'\u27a4'
+                heads_char = '➤'
                 heads_att = settings.get_theming_attribute('thread',
                                                            'arrow_heads')
             A = ArrowTree(

```
```

pazz:
I actually found a bug. Your `ValidatedStoreAction` class does not behave as I expected in that it does not return the result of `validator(values)` but `values` directly. (so I still need to cast myself)..

alot/buffers.py
dcbaker: Then let's not worry about it for now. unicode support in the python2 world is often less than ideal, as we've seen lately.

pazz:
I actually found a bug. Your `ValidatedStoreAction` class does not behave as I expected in that it does not return the result of `validator(values)` but `values` directly. (so I still need to cast myself)..

pazz:
I actually found a bug. Your `ValidatedStoreAction` class does not behave as I expected in that it does not return the result of `validator(values)` but `values` directly. (so I still need to cast myself)..

tests/utils/argparse_test.py
dcbaker: optional_file_like doesn't have a return value, it either raises an exception or it doesn't. So I think those tests are correct.

pazz: ok. Right: so my validator has a return value but (as mentioned below) it will never be used. So maybe the name `is_..` is misleading and one could think that it would return a bool..

dcbaker:
LGTM. Thanks for making those changes.

pazz:
thanks @dcbaker !

Issue #1175, opened 2017-11-10
Packaging of alot-0.6 in Gentoo


GuillaumeSeren:
Hey there !

I am packaging alot-0.6 for Gentoo, and I have trouble with the GPG support.

I have successfully change the dependency from 'dev-python/pygpgme' to 'app-crypt/gpgme',
and I have a few issue, and I am not sure that it comes from config file or the package itself.

Here the PR of the alot-0.6 ebuild: https://github.com/gentoo/gentoo/pull/6153

* Signing require to give a key name of to use a gpg_key value in the config.
* Encrypt do not work with the following error:
```
DEBUG:globals:CMDLINE: encrypt        
DEBUG:ui:envelope command string: "encrypt "                                 
DEBUG:__init__:mode:envelope got commandline "encrypt "                      
DEBUG:__init__:ARGS: [u'encrypt']     
DEBUG:__init__:cmd parms {'action': 'encrypt', 'trusted': False, 'keyids': []}                                                                             
DEBUG:utils:encryption keys: [u'serendev@gmail.com']                         
ERROR:ui:Traceback (most recent call last):                                  
  File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 1237, in unwindGenerator                                                       
    return _inlineCallbacks(None, gen, Deferred())                           
  File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 1099, in _inlineCallbacks                                                      
    result = g.send(result)           
  File "/usr/lib64/python2.7/site-packages/alot/commands/utils.py", line 46, in set_encrypt                                                                
    signed_only=signed_only)          
  File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 1237, in unwindGenerator                                                       
    return _inlineCallbacks(None, gen, Deferred())                           
--- <exception caught here> ---       
  File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 1099, in _inlineCallbacks                                                      
    result = g.send(result)           
  File "/usr/lib64/python2.7/site-packages/alot/commands/utils.py", line 76, in _get_keys                                                                  
    signed_only=signed_only)          
  File "/usr/lib64/python2.7/site-packages/alot/crypto.py", line 76, in get_key                                                                            
    if e.getcode() == gpg.errors.AMBIGUOUS_NAME:                             
exceptions.AttributeError: 'module' object has no attribute 'AMBIGUOUS_NAME'
```

[...]

PR #1086, opened 2017-07-17
Use python-gpg instead of pygpgme

dcbaker:
This is a request for comment, I think this is largely ready to land, but there are a few things I'm not exactly sure about in this, which are marked with XXX or FIXME. Currently this seems to work as well as the pygpgme implementation did, which is to say mostly. There is an unrelated bug that causes some signatures to fail to validate, though I haven't been able to figure out what it is that's causing that.

It is somewhat pressing to get this finished and merged, as debian is moving to deprecated pygpgme and remove it from their repositories.

PR #1178, opened 2017-11-27
Upgrade dependencies to GPGME >= 1.9.0

GuillaumeSeren:
This version of GPGME is needed to support the ':sign' and ':encrypt'
commands, due to the migration of gpgme instead of pygpgme (#1086).

https://github.com/pazz/alot/issues/1175


GuillaumeSeren:
Yes sure @dcbaker

pazz:
I thought that we moved away from gpgme towards gpg?
Whats wrong with the tests on travis? is its package manager missing a suitable package?

dcbaker:
The official bindings for gpgme (which we now use) are called 'gpg' in python, probably because pygpgme (which was a canonical project) was called 'gpgme'. It's confusing.

GuillaumeSeren:
This is so confusing sorry,
I upgraded the files again tell me if there is anything wrong.
(Hope it will be ok)
But he namespace of packages is not the same between languages and distros.

pazz:
this looks harmless enough now. fine with me.

PR #1179, opened 2017-11-29
Bug fixes regarding XDG_* variable expansion + adds general env expansion

three-comrades:
~~Matches the [code](https://github.com/pazz/alot/blob/054d2a932e7cb8d18dae258750ed26fb86e650b9/alot/settings/manager.py#L104).~~

- [x] Fixes `:compose --template=foo` if default template_dir is used
- [x] Fixes `XDG_CONFIG_HOME='' alot` (does not find configuration files etc.)
- Expands environment variables:
 - [x] expand any environment variable
 - [x] write tests and check how it interacts with ConfigObj's string interpolation

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -756,8 +756,9 @@ class ComposeCommand(Command):
         if self.template is not None:
             # get location of tempsdir, containing msg templates
             tempdir = settings.get('template_dir')
-            tempdir = os.path.expanduser(tempdir)
-            if not tempdir:
+            if tempdir:
+                tempdir = os.path.expanduser(tempdir)
+            else:
                 xdgdir = os.environ.get('XDG_CONFIG_HOME',
                                         os.path.expanduser('~/.config'))
                 tempdir = os.path.join(xdgdir, 'alot', 'templates')
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -756,12 +756,6 @@ class ComposeCommand(Command):
         if self.template is not None:
             # get location of tempsdir, containing msg templates
             tempdir = settings.get('template_dir')
-            if tempdir:
-                tempdir = os.path.expanduser(tempdir)
-            else:
-                xdgdir = os.environ.get('XDG_CONFIG_HOME',
-                                        os.path.expanduser('~/.config'))
-                tempdir = os.path.join(xdgdir, 'alot', 'templates')
 
             path = os.path.expanduser(self.template)
             if not os.path.dirname(path):  # use tempsdir
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -625,3 +625,9 @@ def email_as_string(mail):
                            as_string, flags=re.MULTILINE)
 
     return as_string
+
+
+def get_env(env_name, fallback):
+    """ Gets environment variable and returns fallback if unset or empty """
+    env = os.environ.get(env_name)
+    return env if env else fallback
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -627,7 +627,7 @@ def email_as_string(mail):
     return as_string
 
 
-def get_env(env_name, fallback):
-    """ Gets environment variable and returns fallback if unset or empty """
+def get_xdg_env(env_name, fallback):
+    """ Used for XDG_* env variables to return fallback if unset *or* empty """
     env = os.environ.get(env_name)
     return env if env else fallback
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -98,12 +98,12 @@ class SettingsManager(object):
                 self._bindings.merge(newbindings)
 
         tempdir = self._process_xdg_default('template_dir', 'alot/templates')
-        logging.debug(tempdir)
+        logging.debug('template directory: `{}`'.format(tempdir))
 
         # themes
         themestring = newconfig['theme']
         themes_dir = self._process_xdg_default('themes_dir', 'alot/themes')
-        logging.debug(themes_dir)
+        logging.debug('themes directory: `{}`'.format(themes_dir))
 
         # if config contains theme string use that
         data_dirs = [os.path.join(d, 'alot/themes') for d in DATA_DIRS]
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -97,23 +97,12 @@ class SettingsManager(object):
             if isinstance(newbindings, Section):
                 self._bindings.merge(newbindings)
 
-        tempdir = self._config.get('template_dir')
-        if tempdir:
-            tempdir = os.path.expanduser(tempdir)
-        else:
-            xdgdir = os.environ.get('XDG_CONFIG_HOME',
-                                    os.path.expanduser('~/.config'))
-            tempdir = os.path.join(xdgdir, 'alot', 'templates')
+        tempdir = self._process_xdg_default('template_dir', 'alot/templates')
+        logging.debug(tempdir)
 
         # themes
         themestring = newconfig['theme']
-        themes_dir = self._config.get('themes_dir')
-        if themes_dir:
-            themes_dir = os.path.expanduser(themes_dir)
-        else:
-            configdir = os.environ.get('XDG_CONFIG_HOME',
-                                       os.path.expanduser('~/.config'))
-            themes_dir = os.path.join(configdir, 'alot', 'themes')
+        themes_dir = self._process_xdg_default('themes_dir', 'alot/themes')
         logging.debug(themes_dir)
 
         # if config contains theme string use that
@@ -147,6 +136,33 @@ class SettingsManager(object):
 
         self._accounts = self._parse_accounts(self._config)
         self._accountmap = self._account_table(self._accounts)
+
+    def _process_xdg_default(self, setting_name, fallback):
+        """
+        Processes setting that uses $XDG_CONFIG_HOME in the default value.
+
+        Setting should be set as None in alot.rc.default.
+
+        Expands home directory (~) and updates the path in _config object.
+
+        :param setting_name: name of setting to be processed
+        :type setting_name: str
+        :param fallback: fallback path relative to ~/.config if the setting
+        is not set by the user.
+        :type fallback: str
+        :returns: path
+        :rvalue: str
+        """
+        path = self._config.get(setting_name)
+        if path:
+            path = os.path.expanduser(path)
+        else:
+            xdgdir = os.environ.get('XDG_CONFIG_HOME',
+                                    os.path.expanduser('~/.config'))
+            path = os.path.join(xdgdir, fallback)
+
+        self._config[setting_name] = path
+        return path
 
     @staticmethod
     def _parse_accounts(config):
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -96,6 +96,15 @@ class SettingsManager(object):
             newbindings = newconfig['bindings']
             if isinstance(newbindings, Section):
                 self._bindings.merge(newbindings)
+
+        tempdir = self._config.get('template_dir')
+        if tempdir:
+            tempdir = os.path.expanduser(tempdir)
+        else:
+            xdgdir = os.environ.get('XDG_CONFIG_HOME',
+                                    os.path.expanduser('~/.config'))
+            tempdir = os.path.join(xdgdir, 'alot', 'templates')
+
         # themes
         themestring = newconfig['theme']
         themes_dir = self._config.get('themes_dir')
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -15,7 +15,7 @@ from configobj import ConfigObj, Section
 from ..account import SendmailAccount
 from ..addressbook.abook import AbookAddressBook
 from ..addressbook.external import ExternalAddressbook
-from ..helper import pretty_datetime, string_decode
+from ..helper import pretty_datetime, string_decode, get_env
 from ..utils import configobj as checks
 
 from .errors import ConfigError, NoMatchingAccount
@@ -25,8 +25,8 @@ from .theme import Theme
 
 
 DEFAULTSPATH = os.path.join(os.path.dirname(__file__), '..', 'defaults')
-DATA_DIRS = os.environ.get('XDG_DATA_DIRS',
-                           '/usr/local/share:/usr/share').split(':')
+DATA_DIRS = get_env('XDG_DATA_DIRS',
+                    '/usr/local/share:/usr/share').split(':')
 
 
 class SettingsManager(object):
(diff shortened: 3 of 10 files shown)


pazz:
I'd have to have a closer look but I believe that this does not fail graciously if this environment variable is unset (in which case we want to fall back to "~/.config"...

In any case, you'd also have to adjust the [failing test](https://travis-ci.org/pazz/alot/builds/308994091?utm_source=github_status&utm_medium=notification).

three-comrades:
Hm ... I was confused that under `Theming` in the documentation it said the default directory was `~/.config/alot/themes`, yet in the overview of configuration options it said 'None', when almost directly above `template_dir` had a default value of `$XDG_CONFIG_HOME/alot/templates`.

So, to 'fix' this I thought I could handle `themes_dir` the same way `template_dir` is handled. That seems to be wrong, but to be fair, isn't `template_dir` handled wrong as well? At least if I run `:compose --template foobar` I get `no templates directory: $XDG_CONFIG_HOME/alot/templates`. 

So, I pushed another commit on top that fixes that and updates the documentation by simply writing a comment. Tests pass locally (except for one that fails on master as well), let's see ... if this is the right fix, I'll squash.

alot/commands/globals.py
```diff
@@ -755,8 +755,9 @@ def apply(self, ui):
         if self.template is not None:
             # get location of tempsdir, containing msg templates
             tempdir = settings.get('template_dir')
-            tempdir = os.path.expanduser(tempdir)
-            if not tempdir:
+            if tempdir:
```
dcbaker: An alternative would be to do something like:
```python
tempdir = os.path.expandvars(tempdir)
if tempdir == '/alot/templates':
    tempdir = os.path.join(os.environ['HOME'], '.config', tempdir)
````
Then you could leave XDG_CONFIG_HOME as the default path.

Comment (three-comrades) [reply alot/commands/globals.py]: I think I'm not a fan of using `expandvars`, because that would try to expand any `$` in a path, if I understand correctly. If for some reason someone uses '~/my$bill` as a path, it would expand `$bill` if the variable exists. Admittedly that's very unlikely, but why take the risk - I mean XDG_CONFIG_HOME is still used as the default path if it is defined.

But of course, I'm happy to change it if you prefer it :)

Comment (three-comrades) [reply alot/commands/globals.py]: I withdraw my comment. How about this:

```python
    tempdir = settings.get('template_dir')
    if os.path.expandvars(tempdir) == '$XDG_CONFIG_HOME/alot/template':
        tempdir = '~/.config/alot/templates'
    
    tempdir = os.path.expanduser(tempdir)
```

But I don't know, if that file path check works on platforms where the separator is different.

dcbaker: I can go either way, although someone would have to go well out of their way to make `my$bill as a file or path, since shell interpolation will happen before passing the value to the file, and most gui file managers don't let you use `$` as a path, and I'm not a fan of bending over backwards to support people doing stupid things. In this case it's not a big deal either way, you could also do something like:

```python
if 'XDG_CONFIG_HOME' in os.environ:
    tempdir = os.path.expanduser(tempdir)
else:
    tempdir = os.path.join(os.environ['HOME'], '.config', 'alot', 'templates')
```

But it was just a suggestion.

dcbaker:
```python
    tempdir = settings.get('template_dir')
    if os.path.expandvars(tempdir) == '$XDG_CONFIG_HOME/alot/template' and 'XDG_CONFIG_HOME not in os.environ:
        tempdir = '~/.config/alot/templates'
    
    tempdir = os.path.expanduser(tempdir)
```

three-comrades: My last try, inspired  by yours that adds the case where XDG_CONFIG_HOME is actually defined:

```python
    tempdir = settings.get('template_dir')
    if ('XDG_CONFIG_HOME' in os.environ
            and os.environ.get('XDG_CONFIG_HOME')):
        tempdir = tempdir.replace('$XDG_CONFIG_HOME',
                                                  os.environ.get('XDG_CONFIG_HOME'))
    elif tempdir == '$XDG_CONFIG_HOME/alot/templates':
        tempdir = '~/.config/alot/templates'
    tempdir = os.path.expanduser(tempdir)
``

Comment (dcbaker) [reply alot/commands/globals.py]: That looks fairly reasonable, although you could simplify it
```python
if os.environ.get('XDG_CONFIG_HOME', False):
     tempdir = tempdir.replace('$XDG_CONFIG_HOME', os.environ['XDG_CONFIG_HOME'])
...
```
os.environ.get allows you to avoid the initial membership test, and avoiding using after you know that it is a memeber should be slightly faster, plus if someone does something bad like remove XDG_CONFIG_HOME from the environment between the test and the assignment we get an exception instead of a race.

three-comrades: That's indeed nicer. Okay, I will update the PR tomorrow.

three-comrades:
So, ultimately there were two points:

- the bug with `:compose --template foo`
- my initial motivation for this PR: the (to me) confusing documentation.

So, I thought it would make sense to handle `template_dir` and `themes_dir` the same and it seemed to me that the `template_dir` logic should be moved. Well, I hope it makes sense, it works for me. If you think that's a silly idea, I'll just fix `template_dir` without the other shenanigans :)

three-comrades:
I've pushed a simplified version. To summarize this:

- This fixes a bug: `:compose --template=foo`, if `template_dir` is not set in the user configuration file, results in `no templates directory: $XDG_CONFIG_HOME/alot/templates`. The problem is, that in this case this (check)[https://github.com/pazz/alot/blob/c81582a36aeaacbe762fa24208bc778fd4dbf275/alot/commands/globals.py#L759] is always true, because `tempdir` is never empty (it is set to the default value).
- This PR is also a cleanup, because it moves the processing of the `template_dir` setting to `alot/settings/manager.py`, where I think it belongs.
- As a side effect, this also removes a small confusion in the docs, because now `themes_dir` can be documented in the spec file with the config value that is really used, instead of `None`.

lucc:
Of topic: can we configure the codacy bot to remove its own obsolete comments (the summaries)? Or do you mind if I delete them by hand if they start to bother me somewhere?

pazz:
@lucc codacity has 3 toggles for github PRs: summary, comments and status. I've now unticked the status one. Let's see what happens.

three-comrades:
- Split the changes into two commits.
- Split the tests into individual functions (in their own class)

alot/settings/manager.py
```diff
@@ -181,28 +179,26 @@ def _parse_accounts(config):
...
         """
         path = self._config.get(setting_name)
         if path.startswith('$XDG_CONFIG_HOME'):
             if os.environ.get('XDG_CONFIG_HOME', False):
-                path = path.replace('$XDG_CONFIG_HOME',
-                                    os.environ['XDG_CONFIG_HOME'])
+                xdg_expanded = os.environ['XDG_CONFIG_HOME']
             else:
-                path = fallback
+                xdg_expanded = '~/.config'
```
lucc: Instead of the if else stuff you can simply do `xdg_expanded = os.environ.get('XDG_CONFIG_HOME', '~/.config')`.

three-comrades: Turns out that unfortunately that doesn't work, because if `XDG_CONFIG_HOME` is empty, it replaces `$XDG_CONFIG_HOME` with an empty string. The tests caught this 👍

three-comrades:
Thanks a lot for the reviews. Regarding the three commits, I'll leave this then as-is for now, but if anyone wants me to squash the third commit, I'll gladly do so.

pazz:
Regadring the template-dir bug: whats wrong with *just* unsetting the default value as you did and leave the rest? I mean what is so special about this 'XDG_CONFIG_HOME' variable that we will only expand that and no other variables that may potentially be part of the config value string?

three-comrades:
Somehow Github displays the first two commits in the wrong order. It is displayed correctly in [my clone](https://github.com/three-comrades/alot/commits/themes_dir).

I split this up to make it easier to show and pick individual steps. The story here is now:
- fix the bug by unsetting the variable
- move the code, de-duplicate code with a helper function, find another bug with the tests (empty `XDG_*` enviroment variables should be set to the default value). 
- Fix all occurences with a helper function.
- Document default values
- Store the settings in `alot.rc.spec`, instead of unsetting it. @pazz: in my opinition this makes sense, because:
 - it self-documents the correct defaults. 
 - it allows users to use XDG_CONFIG_HOME, which they may expect, because its use is already documented with the default values.
 - it currently only expands XDG_CONFIG_HOME, because that's exactly what the code already does. I think it would be overkill to expand arbitrary environment variables, but maybe it would make sense to allow users to use some other `XDG_*` variables?

If you really just want the minimum change take the first commit and maybe I should open another PR for the 'empty XDG_* variables' case?

pazz:
I am mostly ok with this. Thanks for all the tests:)
I  still don't see why we cannot simply slap
`os.path.expandvars` and `os.path.expanduser` around every config value string inside `settings.get`..

three-comrades:
> I still don't see why we cannot simply slap os.path.expandvars and os.path.expanduser around every config value string inside settings.get..

- I think you'll always need special handling for  `XDG_*` variables, because they require a fallback value if they are empty or not set. 
- I guess it would be nicer to simply apply the environment expansion magic to all configuration values.  
- `ConfigObj` offers the `walk` method to apply a function to all configuration values. If that works, wouldn't it be nicer to process the data once and store it (by using `walk` somewhere in `read_config`), instead of processing the data the same way on each retrieval with `settings.get`?

pazz:
Quoting Julian Mehne (2018-01-22 11:33:50)
>     I still don't see why we cannot simply slap os.path.expandvars and
>     os.path.expanduser around every config value string inside settings.get..
> 
>   • I think you'll always need special handling for XDG_* variables, because
>     they require a fallback value if they are empty or not set.

Good, point, I overlooked that. thanks.

>   • I guess it would be nicer to simply apply the environment expansion magic
>     to all configuration values.

agreed.

>   • ConfigObj offers the walk method to apply a function to all configuration
>     values. If that works, wouldn't it be nicer to process the data once and
>     store it (by using walk somewhere in read_config), instead of processing
>     the data the same way on each retrieval with settings.get?

I'm happy with whatever ends up being least intrusive.

alot/helper.py
```diff
@@ -625,3 +625,9 @@ def email_as_string(mail):
                            as_string, flags=re.MULTILINE)
 
     return as_string
+
+
+def get_env(env_name, fallback):
+    """ Gets environment variable and returns fallback if unset or empty """
+    env = os.environ.get(env_name)
+    return env if env else fallback
```
lucc: How is this function any better than a direct call like `os.environ.get(env_name, fallback)`?

tests/helper_test.py
```diff
@@ -464,3 +464,29 @@ def test_command_fails(self):
...
+
+
+class TestGetEnv(unittest.TestCase):
+    env_name = 'XDG_CONFIG_HOME'
+    default = '~/.config'
+
+    def test_env_not_set(self):
+        with mock.patch.dict('os.environ'):
+            if self.env_name in os.environ:
+                del os.environ[self.env_name]
```
lucc: `mock.patch.dict` can clear the dict it mocks: `...dict('os.environ', clear=True)`.

tests/helper_test.py
```diff
@@ -464,3 +464,29 @@ def test_command_fails(self):
...
+        with mock.patch.dict('os.environ'):
+            os.environ[self.env_name] = ''
+            self.assertEqual(helper.get_env(self.env_name, self.default),
+                             self.default)
+
+    def test_env_not_empty(self):
+        custom_path = '/my/personal/config/home'
+
+        with mock.patch.dict('os.environ'):
+            os.environ[self.env_name] = custom_path
```
lucc: Here and above: `mock.patch.dict` also accepts kwargs to populate the mocking dict: `...dict('os.environ', self.env_name=custom_path)`.

tests/settings/manager_test.py
```diff
@@ -136,6 +136,56 @@ def test_dont_choke_on_regex_special_chars_in_tagstring(self):
         manager.get_tagstring_representation(tag)
 
 
+class TestSettingsManagerProcessXDG(unittest.TestCase):
```
lucc: For some shorter idioms with `mock.patch.dict` see above or check the docs.

alot/helper.py
lucc: This could also be replaced with a simple `get_env = os.environ.get` and the tests should still pass. But then we could just replace it altogether and ditch the tests for get_env.

three-comrades: I should name the function better and add a comment - maybe `get_xdg_env`. The point is that `os.environ.get('env, fallback)` returns `''`, if the variable is set, but empty. But in the case of an empty `XDG_*` variable, it requires the fallback value, otherwise in all the cases where I use `get_env` the paths are wrong.

lucc: Ok I see. I will now approve, you can decide on your own if you want to take the comments in the tests into account or not.

three-comrades: Cool - I'll try to simplify the tests with the help of your comments.

three-comrades:
I'm building on top of the old commits for now, so that new changes are clear.

@pazz: The second-to-last commit implements an expansion function. Regarding `XDG` variables, it  expands `XDG_CONFIG_HOME` and `XDG_CACHE_HOME` - these were the two I found in the current code base.

The last commit just moves a function around. I'm not sure if that is better or not.

tests/helper_test.py
three-comrades: I decided against clearing the whole dictionary after all, because I feel a completely empty environment would make the test conditions a lot different from all the other tests in the test suite.

alot/settings/manager.py
```diff
@@ -137,30 +139,59 @@ def read_config(self):
...
-            else:
-                xdg_expanded = '~/.config'
-            path = path.replace('$XDG_CONFIG_HOME', xdg_expanded)
+        xdg_vars = {'XDG_CONFIG_HOME': '~/.config',
+                    'XDG_CACHE_HOME': '~/.cache'}
+
+        for xdg_name, fallback in xdg_vars.items():
+            if xdg_name in value:
+                xdg_value = get_xdg_env(xdg_name, fallback)
+                value = value.replace('$%s' % xdg_name, xdg_value)\
```
three-comrades: Chaining `replace` calls feels a bit weird, but it is so much more readable and simpler than other variantes I tried (regular expression; some form of a for loop)

alot/settings/manager.py
```diff
@@ -139,6 +138,60 @@ def read_config(self):
         self._accounts = self._parse_accounts(self._config)
         self._accountmap = self._account_table(self._accounts)
 
+
```
lucc: blank line

pazz:
thanks guys

PR #1181, opened 2017-12-02
Give better feedback on invalid user input.

three-comrades:
As an inexperienced user, I do a lot of silly things that leave me wonder why something doesn't work. I'm really impressed at how detailed alot can log, but at the same time I find it sometimes annoying that I have to check the logs to find out why something didn't work - which means, worst case, I have to restart alot and redo the same operation, because per default alot doesn't log anything.

So, in this particular case, I used an email address for the `from:` field that wasn't defined in the config.

In general, there are more cases where I think more information would help:
 - [x] Sending an email. I use `msmtpd` and alot just prints `failed to send: msmtp --account=posteo -t failed with code 78:
A process has ended with a probable error condition: process ended with exit code 78.`, but alot has the exact message in the logs.
 - [x] don't just silently accept a syntactically correct config file that has unknown settings - this made me pull my hair out, when I misspelled `theme_dir` instead of `themes_dir` (which is an easy mistake, because it's `template_dir` and not `templates_dir`) (Edit: or maybe it was the other way around ...)
 - [x] check if the same should be done for theme files.
 - [x] say which config file couldn't be parsed, when there's an error (true for alot's config, notmuch config and abook's address book)

So, tl;dr: if you agree these things are worth fixing, I could take a look and either push to this PR, or open new PRs.

### DIFF
--- a/alot/settings/utils.py
+++ b/alot/settings/utils.py
@@ -5,14 +5,16 @@ from __future__ import absolute_import
 
 import logging
 
-from configobj import ConfigObj, ConfigObjError, flatten_errors
+from configobj import (ConfigObj, ConfigObjError, flatten_errors,
+                       get_extra_values)
 from validate import Validator
 from urwid import AttrSpec
 
 from .errors import ConfigError
 
 
-def read_config(configpath=None, specpath=None, checks=None):
+def read_config(configpath=None, specpath=None, checks=None,
+                report_extra=False):
     """
     get a (validated) config object for given config file path.
 
@@ -23,6 +25,8 @@ def read_config(configpath=None, specpath=None, checks=None):
     :param checks: custom checks to use for validator.
         see `validate docs <http://www.voidspace.org.uk/python/validate.html>`_
     :type checks: dict str->callable,
+    :param report_extra: log if a setting is not present in the spec file
+    :type report_extra: boolean
     :raises: :class:`~alot.settings.errors.ConfigError`
     :rtype: `configobj.ConfigObj`
     """
@@ -65,6 +69,18 @@ def read_config(configpath=None, specpath=None, checks=None):
                     msg = 'section "%s" is missing' % '.'.join(section_list)
                 error_msg += msg + '\n'
             raise ConfigError(error_msg)
+
+        extra_values = get_extra_values(config) if report_extra else None
+        if extra_values:
+            msg = ['Unknown values were found in `%s`. Please check for '
+                   'typos if a specified setting does not seem to work:'
+                   % configpath]
+            for sections, val in extra_values:
+                if sections:
+                    msg.append('%s: %s' % ('->'.join(sections), val))
+                else:
+                    msg.append(str(val))
+            logging.info('\n'.join(msg))
     return config
 
 
--- a/alot/settings/utils.py
+++ b/alot/settings/utils.py
@@ -3,6 +3,8 @@
 # For further details see the COPYING file
 from __future__ import absolute_import
 
+import logging
+
 from configobj import ConfigObj, ConfigObjError, flatten_errors
 from validate import Validator
 from urwid import AttrSpec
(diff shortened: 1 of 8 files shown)


pazz:
Quoting Dylan Baker (2017-12-05 17:54:25)
> For your second point we should see if configobj can check for unkown options,
> since we already use it to validate the config anyway.


I believe that we could use 

  http://www.voidspace.org.uk/python/configobj.html#extra-values

for this: just checking if the return value is empty and otherwise
printing/logging its content.

three-comrades:
>  http://www.voidspace.org.uk/python/configobj.html#extra-values

Great, thanks for digging this up, I missed that when skimming the docs.

So, I did a quick implementation that raises an error and exits the program if there are extra values in alot's config. Looking at the specs file, we should maybe do the same for themes. Will take a look later.

three-comrades:
Okay, this is 'done' from my side. The last commit is optional, but I find it reports errors more clearly - see the commit message.

three-comrades:
Oh, and just to be clear: this PR makes alot error out if there are extra lines in the config or theme file. Given that there were two (I believe unintentional) typos in the tests, I fully expect there to be typos in user configuration files in the wild. So in that regard, it's currently implemented as a breaking change. I'd argue that this is the right thing to do for new users and that for existing users this one time breakage should be easy to fix, because the error message prints the path to the config file and the offending name. But if you think this is a bad trade-off, I'll change it to log only.

lucc:
I would be against the hard error for additional config file fields. I have something like this in my config file: 
```
[accounts]
  # Some defaults for all accounts
  lucas = Lucas Hoffmann
  msmtp = msmtp --file /home/luc/.config/msmtp/msmtprc --read-recipients --read-envelope-from
  sent = maildir:///home/luc/mail/inbox
  draft = maildir:///home/luc/mail/drafts
  # ...
  [[web]]
    address = a@b.c
    # some defaults
    realname = %(lucas)s
    sendmail_command = %(msmtp)s
    sent_box = %(sent)s
    draft_box = %(draft)s
    # ...
  [[lists]]
    address = x@y.z
    # some defaults
    realname = %(lucas)s
    sendmail_command = %(msmtp)s
    sent_box = %(sent)s
    draft_box = %(draft)s
    # ...
```
I am very happy with this generalisation. With you branch I get errors for these general fields under `[accounts]`. 

Descriptive error messages are a good thing.

three-comrades:
> I am very happy with this generalisation.

Oops, I didn't know you could use string interpolation in config files. Okay, so I guess I'll just log extra values as `info` and while I'm at it I could add a test that uses string interpolation to the test suite.

pazz:
I think I'm, with @lucc regarding the extra settings: some users may want to use string interpolation (do we mention this in the docs?) and I would prefer to avoid extra validation code that has to be maintained.

Apart from 9cb75aa and 9d816de the remaining commits look very nice and worth pushing.

three-comrades:
I pushed a new version. Commits bea1848  and 29ca684 were split out from the old validation commits - even if you object to the validation code, these should be fine I think.

The last commit simply logs unknown settings now. I'd like to try one last time to convince you by summarizing:

- I had `templates_dir` and `themes_dir` in my config file - I don't think it is very obvious that one of the settings is misspelled and I found it very frustrating that alot gave no feedback at all (or at least I couldn't find it). With this commit, there's at least a hint in the log.
- Regarding the maintenance burden: the recognition of extra values kind of comes for free with ConfigObj and its `get_extra_values` function. The only code that's really added is for creating a useful help message and a test (that is a bit clumsy ...).

If that still doesn't convince you, I'll drop it of course :-)

pazz:
thanks. I'm happy to push this with extra options detected but onky logged.

alot/settings/utils.py
```diff
@@ -61,6 +69,18 @@ def read_config(configpath=None, specpath=None, checks=None):
...
+        extra_values = get_extra_values(config) if report_extra else None
+        if extra_values:
+            msg = ['Unknown values were found in `%s`. Please check for '
+                   'typos if a specified setting does not seem to work:'
+                   % configpath]
+            for sections, val in extra_values:
+                if sections:
+                    msg.append('%s: %s' % ('->'.join(sections), val))
+                else:
+                    msg.append('%s' % val)
```
lucc: Why not `str(val)`?

three-comrades:
Fixed @lucc's comment and expanded the test a bit (added an expected setting and an unknown section to the test case's config).

PR #1182, opened 2017-12-04
Add intelligent content summary to thread summary widget

ASzc:
When config thread_content_summary is True, some body text is appended to the
displayed email summary strings. The body text is selected through regex
filters, which exclude things like quoted text. This filtering attempts to be
intelligent, but it probably won't be correct 100% of the time. A statistical
approach would result in better quality, but would take much longer to
implement.

I tried to have a minimal impact on the code. However, some extra things that might be beneficial are:

- Extract the filtering logic to a common module, and use it for the search buffer's `content` element too.
- Split the thread buffer's widget layout so that different themeing can be applied to the summary text and the original Name + Date.

### DIFF
--- a/alot/widgets/thread.py
+++ b/alot/widgets/thread.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
 # This file is released under the GNU GPL, version 3 or a later revision.
 # For further details see the COPYING file
@@ -7,6 +8,7 @@ Widgets specific to thread mode
 from __future__ import absolute_import
 
 import logging
+import re
 import urwid
 from urwidtrees import Tree, SimpleTree, CollapsibleTree
 
@@ -21,6 +23,13 @@ class MessageSummaryWidget(urwid.WidgetWrap):
     """
     one line summary of a :class:`~alot.db.message.Message`.
     """
+
+    pat_blank_line = re.compile(r'^\s*$')
+    pat_quote_header_line = re.compile(
+        r'^(?:On .*, .* wrote:|----- Original Message -----)'
+    )
+    pat_quote_line = re.compile(r'^\s*>')
+    body_filters = (pat_blank_line, pat_quote_header_line, pat_quote_line)
 
     def __init__(self, message, even=True):
         """
@@ -40,7 +49,7 @@ class MessageSummaryWidget(urwid.WidgetWrap):
         cols = []
 
         sumstr = self.__str__()
-        txt = urwid.Text(sumstr)
+        txt = urwid.Text(sumstr, wrap='clip')
         cols.append(txt)
 
         if settings.get('msg_summary_hides_threadwide_tags'):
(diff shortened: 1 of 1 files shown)


ASzc:
Pushed a change to try and appease codeclimate

alot/widgets/thread.py
```diff
@@ -22,6 +24,13 @@ class MessageSummaryWidget(urwid.WidgetWrap):
     one line summary of a :class:`~alot.db.message.Message`.
     """
 
+    pat_blank_line = re.compile(r'^\s*$')
+    pat_quote_header_line = re.compile(
+        r'^(?:On .*, .* wrote:|----- Original Message -----)'
```
dcbaker: This is going to be problematic. It first assumes that the message will be in english (which, I get plenty where these sort of metadata tags are not), second it assumes the form of the tag, which may be okay, but I'm not sure.

ASzc: Fair point. I'm thinking that a lot of the parameters would need to be configurable: the exclusion patterns, threshold values, etc. This would allow the user to tweaking things to fit their typical email.

I'm unsure how alot's config parser would react to regex patterns, hopefully it's doable.

dcbaker: alot uses [configobj](http://www.voidspace.org.uk/python/configobj.html) for configuration. I don't think it will have a problem with regex, but we probably need to `re.escape()` the values when they go into alot's config object.

lucc:
The summary for this feature in your first post is very terse. I am not sure I understand it correctly. What does it do (in more verbose and simple terms) and what is the motivation to have this in alot?

Also the new config options needs docs.

ASzc:
@lucc I find the thread view to be visually overwhelming. When a message is expanded, quoted and unquoted text, plus the To:, Subject:, etc. is all mixed together, so it's very hard to see the point at a glance. This is true even if you're only looking at just one message, so quickly getting the point of the whole thread is impossible.

The mail client I was using before had a nice way of displaying one line from each message in a thread, but that line was selected carefully so that it avoided quoted text and useless stuff like `On 5 December 2017 at 19:14, Asd Example <asd@example.com> wrote:` This is what I'm trying to emulate in alot.

Here's a screenshot: 
screenshot from 2017-12-06 07-22-37
I'm still experimenting with the ordering in the line, probably I will attempt to make that configurable too. Here I have the content first and the original Name+Date at the end.

When I get some more time to work on this, I'll work on fixing your suggestions.

teto:
Looks cool/useful ! I would rather have the mail author on the right, similiar to gmail but maybe this is colorscheme-specific.

ASzc:
@teto Yeah, I intend to make it configurable. Still haven't got back to this PR unfortunately.

pazz:
I believe that the least invasive variant to add this feature would be to introduce a new hook, `Message-->str` that is used to produce a text summary of the message. Then you can let a user go wild on personal filters..
In any case, a summary method for messages should perhaps be part of the Message class?

pazz:
closing due to inactivity..

PR #1183, opened 2017-12-16
allow selection of contenttype

### LINKED ISSUE #405: toggle plaintext alternative vs. rendered html
introduce a command in thread mode that allows to toggle the display between
externally rendered html and alternative plaintext

also, make the default choice a config option

ninewise:
Fixes #405.
  
- [ ] Tests
- [x] Rename `:nextctype` to `:nextpart` and reflect this in the code
- [x] Add a "previous" to the "next"
- [x] Fix "prefer_plaintext"
- [x] Quote the selected part on reply/forward
- [x] Make it pretty
- [x] rebase and cleanup

Issues:
- [ ] Alot crashes when I swap to a content tab with less content, because we try to select the same line in every content tab (the same line might not exist).
- [x] A message with multipart/mixed, but only a single part (the attachment), fails to open (example for self: thread:0000000000006739)
- [x] Body of neither text nor html is shown (example for self: thread:0000000000007fc1)

### DIFF
--- a/alot/widgets/thread.py
+++ b/alot/widgets/thread.py
@@ -280,16 +280,22 @@ class MessageTree(CollapsibleTree):
             self._current_part += offset
             self._current_part %= len(mps)
 
+    def get_current_text(self):
+        return self._get_current(1)
+
     def _get_current_part(self):
+        return self._get_current(2)
+
+    def _get_current(self, index):
         mps = self._get_multiparts()
         if mps:
-            return mps[self._current_part][1]
+            return mps[self._current_part][index]
 
     def _get_multipart_selector(self):
         mps = self._get_multiparts()
         if mps:
             return MultipartWidget(
-                [ctype for ctype, _ in self._get_multiparts()],
+                [ctype for ctype, _, _ in self._get_multiparts()],
                 self._current_part)
 
     def _get_multiparts(self):
@@ -300,6 +306,7 @@ class MessageTree(CollapsibleTree):
 
             self._multiparts = [
                 (ctype,
+                 text,
                  TextlinesList(text, att, att_focus))
                 for ctype, text
                 in self._message.get_body_parts()]
--- a/alot/widgets/thread.py
+++ b/alot/widgets/thread.py
@@ -304,7 +304,11 @@ class MessageTree(CollapsibleTree):
                 for ctype, text
                 in self._message.get_body_parts()]
 
-            self._current_part = 0
+            preferred = 'text/plain' if settings.get(
+                'prefer_plaintext') else 'text/html'
+            prefs = [i for i, e in enumerate(self._multiparts)
+                     if e[0] == preferred]
+            self._current_part = prefs[0] if prefs else 0
 
         return self._multiparts
 
--- a/alot/widgets/thread.py
+++ b/alot/widgets/thread.py
@@ -182,7 +182,7 @@ class MessageTree(CollapsibleTree):
         self._default_headers_tree = None
         self.display_attachments = True
         self._attachments = None
-        self._current_part = 0
+        self._current_part = None
         self._multiparts = None
         self._maintree = SimpleTree(self._assemble_structure(True))
         CollapsibleTree.__init__(self, self._maintree)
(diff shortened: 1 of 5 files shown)


NoctuaNivalis:
I fixed the failing tests.

The 2 codeclimate issues are actually issues already there before I started modifying these functions, should I try to fix them? (What is cognitive complexity anyway?)

I'm not sure what's wrong with the first travis job, it doesn't show any errors but still fails.

alot/widgets/thread.py
```diff
@@ -236,15 +255,31 @@ def _get_source(self):
             self._sourcetree = TextlinesList(sourcetxt, att, att_focus)
         return self._sourcetree
 
+    def _get_content_parts(self):
```
pazz: I think this logic should not be part of the FocusableText widget tbh.

NoctuaNivalis: It isn't? :confused: It's in MessageTree.

pazz: Oh sorry, my bad: Github folding tricked me into believing the change is in FocussableText. I take it back :)

lucc:
I like this feature very much. I tried to implement a different way to navigate the mime tree some time back (#894) but never finished it as I was stuck with some urwidtree issues. So I hope we can merge this feature but here are some more points for consideration/discussion/todo:

* Tests for new code would be nice. Alot aims to improve its testsuite so a new feature and new code is always a good chance to add some more tests.
* I have `prefer_plaintext = True` in my config but it is not respected any more. I always get the first mime part initially, no matter its type (I admit it is not very often that you get a mail that has a html part before the plain text part, I only found one in my mail store after a short search)
* We might want a `:previousctype` command for symmetry but that can also wait for a later PR (also do we want to jump to one ctype directly or allow count arguments for these commands?)
* It could maybe be named `:nextpart` as there might be several parts that do not represent the same information in different types but actually different info. (eg. a mailing list digest from the notmuch mailing list)
* I am unsure about the visual representation. For me the ctype indication merges with the body but it should feel more like a part of the header information. Or even something like a tabline below the headers. Also mails might have attachements and be signed or encrypted. So we should spend some thought on different cases and how to represent them.

BTW I tried the PR and it works :)

NoctuaNivalis:
- Tests should indeed be written. I'll postpone them because I'm lazy and we might still rename some functions though.
- My PR actually brakes/changes the `prefer_plaintext` option. AFAIK, it now only influences which text is quoted when replying to an email. I'll select the preferred part by default (but not change the order of the parts) and ensure the quoted text is the currently visible one, not the preferred ctype.
- A "previous" variant could be trivially added, so I'll add it to here. Only reason I didn't already is, I couldn't think of a good keybinding for it.
- `:nextpart` sounds better. I'll rename this later unless someone disagrees.
- Visual representation is hard. I tried something with trees first, but I got stuck with cycles in the MessageTree :-/. I'd appreciate some thoughts on this.

And yeah, it works ;-)

lucc:
Your suggestion in the second point sounds good: select preferred part when opening a mail and use current part when replying. (I have a test mail ready to try it ;)

The cycles in trees are exactly what stopped me in my tracks back in #894.

pazz:
what do you mean by cycles in the tree?

NoctuaNivalis:
So I had to do something soft of ugly to quote the selected part, but I'm not sure it can be improved without overhauling a lot of command-related-code. Is it even possible to reply to some message that's not the `ui.current_buffer.get_selected_message()`?

NoctuaNivalis:
About the cycles, it's probably because I did something wrong with the parent pointers, and at some point while holding the down arrow, all parents are suddenly erased and the parents are gone. Holding the up arrows goes through an infinite cycle of messages.

lucc:
preferring plaintext works again.

lucc:
maybe also something for the todo list: rebase and clean up history with nice logical steps in the commit log and descriptive messages.

ninewise:
Rebased this feature onto the new master.

ninewise:
I need some help to finish this:

* Are there some examples of graphical tests anywhere? I can write a test for the `quote` function, but all other tests are graphical and I'm not seeing any in tests/widgets, yet.

* Is there a way to set the position of the cursor? I'd use this to resolve the first issue in the top post here, but I'm not finding how in the urwidtrees documentation.

ninewise:
I believe this was superseded by #1480 so I'm closing this.

PR #1184, opened 2017-12-20
Add option to hide tags that are used in the current search query of a search buffer.

three-comrades:
If this option is true, tags that are part of the current search query are not displayed as part of the threadline of each mail in the search buffer. E.g., if the current query is `tag:inbox and tag:gmail`, the `inbox` and `gmail` tags are not displayed for each email. Because these tags are common for each mail in the search buffer, they don't contain any new information and just make it harder to visually process the relevant tags, i.e., the ones that are not common for all mails in the buffer.

The changes are not very invasive, but I didn't really know how to test this, so that's why I have the static method ...


pazz:
nice feature!
Btw: as far as I remember, the notmuch query syntax allows the keyword 'is' equivalently to 'tag'..

three-comrades:
>  Btw: as far as I remember, the notmuch query syntax allows the keyword 'is' equivalently to 'tag'..

Thanks, you're right! I pushed a new version.

I just realized that the title of this PR is correct, but then I go on and talk about how it hides tags that are common to all mails - which this doesn't do. This PR is dumb and hides all tags that are explicity used in the query (it ignores wildcards and regular expressions, but I guess this could be fixed without being super invasive), i.e. `tag:foo or tag:bar` hides `foo` **and** `bar`. I think this is still useful, even as a general option, but I can imagine there are cases where I would want to toggle this behavior for a search buffer. Hm ...

three-comrades:
I've added a `togglequerytags` command in a second commit that toggles the behavior in a search buffer.

pazz:
Would you mind rebasing this to current master? We've refactored the threadline widgets a bit since you proposed this first..

PR #1187, opened 2018-01-02
mention unrendered HTML parts directly in message body text.

pazz:
This fixes issue #987 and is intended as a replacement for #1163.

### DIFF
--- a/alot/db/message.py
+++ b/alot/db/message.py
@@ -18,6 +18,11 @@ from ..settings.const import settings
 
 charset.add_charset('utf-8', charset.QP, charset.QP, 'utf-8')
 
+MISSING_HTML_MSG = ("This message contains a text/html part that was not "
+                    "rendered due to a missing mailcap entry. "
+                    "Please refer to item 5 in our FAQ: "
+                    "http://alot.rtfd.io/en/latest/faq.html")
+
 
 @functools.total_ordering
 class Message(object):
@@ -256,7 +261,14 @@ class Message(object):
         returns bodystring extracted from this mail
         """
         # TODO: allow toggle commands to decide which part is considered body
-        return extract_body(self.get_email())
+        eml = self.get_email()
+        bodytext = extract_body(eml)
+
+        # check if extracted body is empty but msg contains html parts
+        if (not bodytext and
+           'text/html' in (part.get_content_type() for part in eml.walk())):
+            return MISSING_HTML_MSG
+        return bodytext
 
     def get_text_content(self):
         return extract_body(self.get_email(), types=['text/plain'])
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -325,6 +325,10 @@ def extract_body(mail, types=None, field_key='copiousoutput'):
 
     body_parts = []
     for part in mail.walk():
+        # skip non-leaf nodes in the mail tree
+        if part.is_multipart():
+            continue
+
         ctype = part.get_content_type()
 
         if types is not None:
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -346,10 +346,11 @@ def extract_body(mail, types=None, field_key='copiousoutput'):
         else:
             # get mime handler
             _, entry = settings.mailcap_find_match(ctype, key=field_key)
-            tempfile_name = None
-            stdin = None
-
-            if entry:
+            if entry is None:
+                part.add_header('Content-Disposition', 'attachment; ' + cd)
+            else:
+                tempfile_name = None
+                stdin = None
                 handler_raw_commandstring = entry['view']
                 # in case the mailcap defined command contains no '%s',
                 # we pipe the files content to the handling command via stdin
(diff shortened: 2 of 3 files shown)


alot/db/message.py
```diff
@@ -18,6 +18,10 @@
 
 charset.add_charset('utf-8', charset.QP, charset.QP, 'utf-8')
 
+MISSING_HTML_MSG = """
+This message contains a text/html part that could not be rendered due to a missing mailcap entry.
+Please refer to item 5 in out FAQ.
```
lucc: our

pazz: yep, thanks

lucc:
After a longer look at this I must say that I don't like to have warning or error messages in the body text. Back when I used mutt there was a discussion somewhere (or a comment in the docs) about how intermixing program generated text that should be displayed (pgp info stuff, error messages) with the actual message. The problem is that an "attacker" could try to send you a message that might look similar to one of these intermixed messages. We are already dealing with this kind of problem in one place: the pgp signature pseudo header.

Back to the warning message in the body text. I would prefer some kind of notification. That is what notifications are for. I understand that on a long thread it is very annoying to see many of these notifications pop up. So the question is if we can find another solution.

1. We could delay rendering the html in some way until a message is unfolded or actually shown on screen. But I have not seen code of that kind anywhere in alot so it might be hard to do.
2. We might introduce a new kind of body text widget. It could be shown inside the thread tree *instead* of the bodytext. This could be used for all kinds of error messages of this kind and could be styled different. (my current preference)

pazz:
Quoting Lucas Hoffmann (2018-01-09 08:46:06)
> After a longer look at this I must say that I don't like to have warning or
> error messages in the body text. Back when I used mutt there was a discussion
> somewhere (or a comment in the docs) about how intermixing program generated
> text that should be displayed (pgp info stuff, error messages) with the actual
> message. The problem is that an "attacker" could try to send you a message that
> might look similar to one of these intermixed messages. We are already dealing
> with this kind of problem in one place: the pgp signature pseudo header.

In principle I agree but for me this is not a priority and the proposed
hotfix is better than nothing and having to refer people to the FAQ all
the time.
 
> Back to the warning message in the body text. I would prefer some kind of
> notification. That is what notifications are for. I understand that on a long
> thread it is very annoying to see many of these notifications pop up. So the
> question is if we can find another solution.

Again, In principle yes, but I can't think of easy to implement
alternatives atm.

We could just raise an exception in `utils*.*accumulate_body()`
and catch it in `ui.apply_command`. This would however stop the
threadopen command (not open the thread buffer). OK with me.

 
>  1. We could delay rendering the html in some way until a message is unfolded
>     or actually shown on screen. But I have not seen code of that kind anywhere
>     in alot so it might be hard to do.

Possible, yea: In the command that opens the threadbuffer iterate over
message widgets and notify if some predicate `message.unrendered_html`
is true in one of them.

>  2. We might introduce a new kind of body text widget. It could be shown inside
>     the thread tree instead of the bodytext. This could be used for all kinds
>     of error messages of this kind and could be styled different. (my current
>     preference)

Sounds like a lot of effort to me.

alot/db/message.py
```diff
@@ -256,7 +261,17 @@ def accumulate_body(self):
...
+        bodytext = extract_body(eml)
+
+        # check if extracted body is empty but msg contains html parts
+        if not bodytext:
+            types = []
+            for part in eml.walk():
+                types.append(part.get_content_type())
+            if 'text/html' in types:
+                bodytext = MISSING_HTML_MSG
+        return bodytext
```
lucc: Can this be simplified to this:
```py
if not bodytext:
    for part in eml.walk():
        if part.get_content_type() == 'text/html':
            return MISSING_HTML_MSG
return bodytext
```

lucc: Or even
```py
if not bodytext and 'text/html' in (part.get_content_type() for part in eml.walk()):
    return MISSING_HTML_MSG
return bodytext
```

alot/db/message.py
```diff
@@ -18,6 +18,11 @@
 
 charset.add_charset('utf-8', charset.QP, charset.QP, 'utf-8')
 
+MISSING_HTML_MSG = ("This message contains a text/html part that was not"
```
lucc: missing space after "not".

alot/db/message.py
```diff
@@ -18,6 +18,11 @@
...
+                    "rendered due to a missing mailcap entry."
```
lucc: missing space

pazz:
Quoting Lucas Hoffmann (2018-01-09 10:27:55)
> @lucc commented on this pull request.
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In alot/db/message.py:
> 
> > @@ -18,6 +18,11 @@
> 
>  charset.add_charset('utf-8', charset.QP, charset.QP, 'utf-8')
> 
> +MISSING_HTML_MSG = ("This message contains a text/html part that was not"
> +                    "rendered due to a missing mailcap entry."
> 
> missing space

Thank you

alot/db/message.py
pazz: sure, sounds good. I've take the latter and indented the long conditional.

ids1024:
This message seems to also display if the mailcap command fails, which can be confusing. It would be helpful if it said "command not found" or "failed with exit status..." in those cases.

PR #1188, opened 2018-01-08
Fix test output

lucc:
This fixes some things that I would suggest on top of the last minute fixes in #1181.  Most importantly the commit about print().  The other two commits are less important.  The thing with any() is subject to tast (what do you think?).

### DIFF
--- a/tests/settings/manager_test.py
+++ b/tests/settings/manager_test.py
@@ -102,12 +102,8 @@ class TestSettingsManager(unittest.TestCase):
 
         with mock.patch('alot.settings.utils.logging') as mock_logger:
             SettingsManager(alot_rc=f.name)
-        success = False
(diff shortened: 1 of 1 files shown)


three-comrades:
Sorry for implementing last minute fixes, I thought they were simple enough - I didn't mean to invalidate your review. Good to know how to properly report messages in case of failure - still learning ...

I think the `any` line with two list comprehensions is now:
- a bit harder to parse
- nicer because it got rid of the `success` + `break` construct
- less nice because the meaning of the cryptic `call_args[0][0]` is not explicitly stated any more.

Personally, I like it when test code is very explicit - it makes especially debugging a lot easier, in my opinion (I can easily print  `call_args` in the explicit version, for example, or just put an explicit debugger call somewhere in the loop). But then again, it is still not that complex, so either version should be fine?

Issue #1051, opened 2017-05-30
compose html mails


pazz:
I often find myself writing plaintext mails in a format similar to markdown, mentioning links like this [0].
It should be great to be able to actually write markdown, and pipe this through a md2html process before sending to have inline links as well.

One alternative would be to actually compose html in the editor, and let alot recognize that when constructing the MIME-structure.
[0]: https://example.com

dcbaker:
I actually very much like the idea of using markdown or restructured text or similar, and being able to pipe it through a tool to convert to HTML before sending.

pazz:
You can of course already do that using post-edit hooks. The new thing
here would be to actually construct html MiME parts when building the
mail to send..

Quoting Dylan Baker (2017-05-30 18:14:03)

dkg:
see also [the discussion on LWN about sending and rendering text/markdown directly](https://lwn.net/Articles/730277/).

PR #1192, opened 2018-01-28
Call a text-to-html converter to build html alternative body parts

pazz:
as discussed in #1051.

This will use a new config string "txt2html_converter" to translate plaintext body message to html and change the created mime structure accordingly.
See https://stackoverflow.com/questions/10631856/mime-type-to-satisfy-html-email-images-and-plain-text

It works nicely for me when I compose simple markdown and use "pandoc -t html" as converter.
But I expect it to get messy with reply quotes (untested for now).

### DIFF
--- a/alot/buffers/envelope.py
+++ b/alot/buffers/envelope.py
@@ -89,7 +89,7 @@ class EnvelopeBuffer(Buffer):
             self.attachment_wgt = urwid.Pile(lines)
             displayed_widgets.append(self.attachment_wgt)
 
-        self.body_wgt = urwid.Text(string_sanitize(self.envelope.body))
+        self.body_wgt = urwid.Text(string_sanitize(self.envelope.body_txt))
         displayed_widgets.append(self.body_wgt)
         self.body = urwid.ListBox(displayed_widgets)
 
--- a/alot/buffers/envelope.py
+++ b/alot/buffers/envelope.py
@@ -10,6 +10,8 @@ from ..widgets.globals import HeadersList
 from ..widgets.globals import AttachmentWidget
 from ..helper import shorten_author_string
 from ..helper import string_sanitize
+from ..db.utils import render_part
+from email.mime.text import MIMEText
 
 
 class EnvelopeBuffer(Buffer):
@@ -21,6 +23,7 @@ class EnvelopeBuffer(Buffer):
         self.ui = ui
         self.envelope = envelope
         self.all_headers = False
+        self.displaypart = "plaintext"
         self.rebuild()
         Buffer.__init__(self, ui, self.body)
 
@@ -31,6 +34,7 @@ class EnvelopeBuffer(Buffer):
     def get_info(self):
         info = {}
         info['to'] = self.envelope.get('To', fallback='unset')
+        info['displaypart'] = self.displaypart
         return info
 
     def cleanup(self):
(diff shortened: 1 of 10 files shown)


lucc:
For replies: https://www.w3schools.com/TAGS/tag_blockquote.asp

pazz:
Quoting Lucas Hoffmann (2018-01-28 17:17:41)
> For replies: https://www.w3schools.com/TAGS/tag_blockquote.asp


nice idea! But what's with inline replies?
Also, I would expect a reply prefic "> " even in html mail..

lucc:
I just checked some html I mail in my store (notmuch syntax to the rescue: `mimetype:html`) and they really use `<blockquote>...</blockquote>`.

pazz:
thanks. It seems that one can also nest these. That leaves the problem that alot
doesn't yet allow to edit html mail bodies: atm the workflow is:

- (html) mail comes in, maybe containing blockquotes,
- your textbrowser converts it to txt
- you edit it, possibly adding stuff in between the blockquotes
- txt2html converter creates html.

I have to investigate what happens to the blockquotes in step 3.
One possibility would be to put every line starting with a reply prefix into a
blockquote before turning the whole thing into html.
At least pandoc will not escape this when translating markdown to html afaik..

Quoting Lucas Hoffmann (2018-01-28 22:14:49)

lucc:
with pandoc I get this:
```
pandoc -t html <<EOF
first

> inside
> foo

after
EOF
<p>first</p>
<blockquote>
<p>inside foo</p>
</blockquote>
<p>after</p>
```

pazz:
OK that looks reasonable so it *should* work out of the box.
It's just that I've seen a scrambled mail in my gmail webthingie when I tested
it briefly.. will investigate.

Quoting Lucas Hoffmann (2018-01-29 09:40:54)

CaptainQuirk:
Hi !

Any news on this ?

I'm really looking forward to it !

pazz:
sorry i haven't had time to follow up on this POC. 
as is it is still very raw: the is no nice mechanism to test, remove or edit the generated html before sending.
I suspect we'd want a few extra extra envelope commands, such as edit-html, plain-to-html, remove-html...
I'd happily consider PRs on this!

CaptainQuirk:
I'm afraid that my very limited knowledge of python only bests my ignorance of the inner workings of alot (and emails) !
I could help in outlining the problems to solve ... Write some rough specification. What do you make of it ?

Thanks !

pazz:
I've been thinking about this further: A nice solution UI-wise would be to 
- have `envelope`s store separately a plaintext "original" and the generated HTML.
- Similar to the `togglesource` command in thread mode, one can switch between three views: plaintext, html code, rendered html.
- the `edit` command will load the current variant into the editor (this way you can still edit generated html). Optionally, html is re-generated after editing plaintext
- new command to manually trigger txt2html converter.

pazz:
Please test and comment. I am using the following in my config:

```
envelope_txt2html = "pandoc -f markdown -t html"
envelope_html2txt = "pandoc -t markdown -f html"
```

I think we should add 

[ ] an easier "toggle" command in tnvelope mode to iterate through the three display modes (plaintext, html source, rendered html).
[ ] default bindings for the above, and the new command `txt2html` and `html2txt`
[ ] unit tests *cough*

pazz:
The new codeclimate issue will not be fixed btw, because it is a silly "you shall not have more than 4 parameters in constructors" convention

pazz:
Just played around with this a bit and what works really well is to bind the enter (edit) key in envelope to

`edit --part=plaintext; txt2html; display html`

This way, you can always edit the plaintext version as markdown and after closing the editor, that would be converted to html via the command set in `envelope_txt2html`, and then rendered again and displayed as usual, via the mailcap entry.

CaptainQuirk:
Can't wait to try it out ! Thanks

patricktotzke:
Bug with forwarding (but not replies?!?) on this branch: the template sent to the editor does not contain q quoted version of the previous mail any more.

pazz:
Fixed the fwd issue (was a typo)

Issue #673, opened 2013-11-15
UnicodeEncodeError when starting alot


chmduquesne:
Hi,

I tried to start alot (current git version, installed from AUR with archlinux). I cannot get it to work: running 'alot' gets me the following stacktrace:

```
$ alot 

Traceback (most recent call last):
  File "/usr/bin/alot", line 20, in <module>
    main()
  File "/usr/lib/python2.7/site-packages/alot/init.py", line 187, in main
    UI(dbman, cmd)
  File "/usr/lib/python2.7/site-packages/alot/ui.py", line 87, in __init__
    self.mainloop.run()
  File "/usr/lib/python2.7/site-packages/urwid/main_loop.py", line 272, in run
    self.screen.run_wrapper(self._run)
  File "/usr/lib/python2.7/site-packages/urwid/raw_display.py", line 242, in run_wrapper
    return fn()
  File "/usr/lib/python2.7/site-packages/urwid/main_loop.py", line 312, in _run
    self.draw_screen()
  File "/usr/lib/python2.7/site-packages/urwid/main_loop.py", line 563, in draw_screen
    canvas = self._topmost_widget.render(self.screen_size, focus=True)
  File "/usr/lib/python2.7/site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/lib/python2.7/site-packages/urwid/decoration.py", line 225, in render
    canv = self._original_widget.render(size, focus=focus)
  File "/usr/lib/python2.7/site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/lib/python2.7/site-packages/urwid/container.py", line 1058, in render
    focus and self.focus_part == 'body')
  File "/usr/lib/python2.7/site-packages/alot/buffers.py", line 37, in render
    return self.body.render(size, focus)
  File "/usr/lib/python2.7/site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/lib/python2.7/site-packages/urwid/listbox.py", line 487, in render
    canvas = widget.render((maxcol,))
  File "/usr/lib/python2.7/site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/lib/python2.7/site-packages/alot/widgets/search.py", line 168, in render
    return urwid.AttrMap.render(self, size, focus)
[...]

PR #1193, opened 2018-01-29
Fix `LANG=C alot` with custom bindings.

three-comrades:
This fixes two crashes I encountered when playing around with `LANG=C alot` in #673, but @pazz couldn't reproduce it. As requested, I post the one-line change here, but if no one can reproduce it I'm not sure it's worth it :-)

Reproduction steps:
- Set a custom binding in your config file. E.g.:
[bindings]
    S = toggletags spam
- start alot with LANG=C: LANG=C alot
- variant A: type ö
- variant B: type :search ö

Problem: ConfigObj reads the custom binding as a unicode string, but ö is
passed as a binary string. The interaction of both leads to crashes.

Related to #673

Edit: backtrace from variant A (LANG=C alot, type ö) from #673:
```
LANG=C alot -l log -d debug
  File "/home/kelvin/tmp/alot/venv/bin/alot", line 11, in <module>
    load_entry_point('alot==0.7.0.dev0', 'console_scripts', 'alot')()
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/alot-0.7.0.dev0-py2.7.egg/alot/__main__.py", line 130, in main
    UI(dbman, cmdstring)
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/alot-0.7.0.dev0-py2.7.egg/alot/ui.py", line 124, in __init__
    self.mainloop.run()
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/urwid-1.3.1-py2.7-linux-x86_64.egg/urwid/main_loop.py", line 278, in run
    self._run()
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/urwid-1.3.1-py2.7-linux-x86_64.egg/urwid/main_loop.py", line 376, in _run
    self.event_loop.run()
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/urwid-1.3.1-py2.7-linux-x86_64.egg/urwid/main_loop.py", line 1200, in wrapper
    rval = f(*args,**kargs)
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/urwid-1.3.1-py2.7-linux-x86_64.egg/urwid/raw_display.py", line 393, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/urwid-1.3.1-py2.7-linux-x86_64.egg/urwid/raw_display.py", line 493, in parse_input
    callback(processed, processed_codes)
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/urwid-1.3.1-py2.7-linux-x86_64.egg/urwid/main_loop.py", line 400, in _update
    keys = self.input_filter(keys, raw)
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/urwid-1.3.1-py2.7-linux-x86_64.egg/urwid/main_loop.py", line 539, in input_filter
    return self._input_filter(keys, raw)
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/alot-0.7.0.dev0-py2.7.egg/alot/ui.py", line 189, in _input_filter
    prefix=keyseq)
  File "/home/kelvin/tmp/alot/venv/lib/python2.7/site-packages/alot-0.7.0.dev0-py2.7.egg/alot/settings/manager.py", line 356, in get_mapped_input_keysequences
    cand = [c for c in candidates if c.startswith(prefixes)]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 0: ordinal not in range(128)
```

Log entry:
```
DEBUG:ui:Got key (['\xc3', '\xb6'], [195, 182])
```

### DIFF
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -75,6 +75,9 @@ class UI(object):
         # alarm handle for callback that clears input queue (to cancel alarm)
         self._alarm = None
 
+        # force urwid to pass key events as unicode, independent of LANG
+        urwid.set_encoding('utf-8')
+
         # create root widget
         global_att = settings.get_theming_attribute('global', 'body')
         mainframe = urwid.Frame(urwid.SolidFill())


pazz:
can you check what `urwid.get_encoding_mode()` returns from within an alot pyshell?
I get "utf8" without this patch. I presume you dont?

three-comrades:
I get `narrow`, when I start alot (or just python) with `LANG=C alot` and without this patch. This is consistent with when I start python in an empty environment:

```
$ env -i python2 -c "import urwid; print urwid.get_encoding_mode()"
narrow
```

lucc:
In general: Is it possible to create a unittest for that? Should we have that?

For reproducability issues:

I am also an Arch linux. All of the following give me "narrow":
```
env -i python2 -c "import urwid; print urwid.get_encoding_mode()"
env -i LANG=C python2 -c "import urwid; print urwid.get_encoding_mode()"
env -i LANG=C GNUPGHOME=$GNUPGHOME NOTMUCH_CONFIG=$NOTMUCH_CONFIG =alot pyshell <<<$'import urwid\nprint urwid.get_encoding_mode()\n'
```
The last one also gives me a stack trace but I did not find a way to pipe the correct alot commands into the shell command. The env variables GNUPGHOME and NOTMUCH_CONFIG are needed as I store these things in non standard places and alot will not start without them.

`uname -a` says `Linux tp 4.14.15-1-ARCH #1 SMP PREEMPT Tue Jan 23 21:49:25 UTC 2018 x86_64 GNU/Linux`, `python2 --version` is `Python 2.7.14` and alot is on `0.7rc1` = 7a37daee30c271e2aca7546bee910d419e4c2998. Some of the maybe relevant pip2 modules:
```
alot (0.7.0.dev0, /home/luc/vcs/alot)
configobj (5.0.6)
configparser (3.5.0)
gpg (1.8.0)
mock (2.0.0)
notmuch (0.26)
pip (9.0.1)
pygpgme (0.3)
python-magic (0.4.12)
setuptools (38.4.0)
Twisted (17.9.0)
unittest2 (1.1.0)
urwid (1.3.1)
urwidtrees (1.0.1.1)
virtualenv (15.1.0)
```

In my normal environment I get `utf8` from urwid (both plain and inside alot). I have `LANG=en_US.UTF-8` in that environment.

three-comrades:
> In general: Is it possible to create a unittest for that? Should we have that?

I don't think this requires a classical unit test, because there's no real bug in any alot related code. This happens because in this special case urwid passes keys as `str` instead of `unicode`, but this is simply 'invalid' input/an invalid state for alot, isn't it? So maybe an assert in the `_input_filter`  method in `alot/ui.py` or somewhere else would guard against it.

But are there any locations in the code or external circumstances where the encoding used by urwid could be changed after the program start? If not, then I think it's reasonable to just set the encoding once at startup, maybe add a comment and be done.

pazz:
Will this be an issue in python3 at all? I don't think so because there is no
way urwid can misinterpret input as str vs unicode, right?

So to push things on, I'm all for just pushing this patch here without extra
tests. We will soon move development to python3 anyway.

three-comrades:
>  I don't think so because there is no way urwid can misinterpret input as str vs unicode, right?

I don't really know the details and haven't spent enough time reading about it, but when doing the py2->3 switch there were lots of complaints about how Python3 handles bytes (for example: [here](http://click.pocoo.org/5/python3/)) and there is a very recent PEP that may be related, called [Coercing the legacy C locale to a UTF-8 based locale](https://www.python.org/dev/peps/pep-0538/), so I'd expect it to come with its own surprises.

I feel a bit uneasy about the change because I don't really understand yet how key processing works and I'm a bit afraid it may garble input/output for other encodings like iso-8859-X. I imagine it to work this way (does anyone know a good resource?):
- key press
- terminal sends a byte sequence to urwid, and the encoding depends on LC_CTYPE?
- urwid (possibly) decodes the byte or passes it on as bytes - is is this dependent on `urwid.set_encoding`? If the decoding is done with the wrong encoding, the input is messed up?
- for output, urwid encodes a string and sends it to the terminal - if the encodings don't match, the output is garbled?

I tested this with `LANG=de_DE.ISO-8859-1` and for me it seems to fix the same crash there. So I guess it's okay?

PR #1194, opened 2018-02-03
Add support for optional diff and quoting highlighting

ivyl:
Introducing couple of new configurations options in [thread] section:
 * diff_add  - applied to lines starting with +
 * diff_del  - applied to lines starting with -
 * diff_head - applied to lines staring with either "@@" or "diff "
 * nested1 to nested5 - for highlighting nested quoted messages (lines
                        starting with something like "> > >")

Due to how urwid and alot handle displaying text - one widget per line
(or whole message) and styling happening per widget - this change works
only with thread_focus_linewise turned on (the default) and the
highlighting is line-wide.

This is a very rough and ready patch to provide those features as fast
as possible and encourage discussion on a proper implementation.

### DIFF
--- a/alot/defaults/default.theme
+++ b/alot/defaults/default.theme
@@ -31,9 +31,17 @@
     attachment_focus = 'underline','','light gray','light green','light gray','light green'
     body = 'default','','light gray','default','light gray','default'
     body_focus = 'default','','light gray','default','white','default'
+    diff_add = 'default','','light gray','default','light gray','default'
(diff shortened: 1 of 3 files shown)


pazz:
Hi, and thanks for your proposal. 
I generally agree that there should be some nice way of highlighting different text attachments / inline code etc., but I believe that this series is a bit too invasive for the limited scope it will have. Ideally, we would have something (based on external filters and ANSI escape codes #1015 similar to what the ranger filemager has?) that will work for a wider range of text blocks.

PR #1195, opened 2018-02-08
Fix gpg ambiguous key handling

dcbaker:
Currently what happens if there is an ambiguous key, we will present
two potentially identical uids, then add the uid to the list of
potential uid. This results in an infinite loop (although it looks like
it's frozen, it's actually processing input).

To fix this we need to not add the key to the potential keys to
consider, we need to add it to the list of keys we've already
considered. This requires some changes to the core widgets to make work.


dcbaker:
fixed by another commit

PR #1196, opened 2018-02-08
fix ambiguous keys harder

dcbaker:
always pass keyword arguments as thus, to prevent problems like this.


pazz:
Sorry, I don't get it. What exactly does this fix? also, the commit message should read "header", right?

dcbaker:
The problem is that the arguments we're trying to pass are `choices`, and `choices_to_return`, but because the arguments are passed as positional we're actually passing `choices` and `select`.

pazz:
Oh right. Thanks for clearing this up.

Quoting Dylan Baker (2018-02-21 16:42:04)

PR #1197, opened 2018-02-12
Add --tags argument to compose command

kyrias:
Missed this earlier today when I wanted to be able to add alot-compose to a mostly scripted workflow for a project I'm contributing to.


pazz:
This looks like a reasonable patch to me, thanks.
Could you fix the test that fails due to the adjusted syntax?

kyrias:
Fixed!

lucc:
I took the liberty to merge this as I wanted to use it (in a script).

PR #1199, opened 2018-02-12
Fix templates with non-ascii characters.

### LINKED ISSUE #1198: Alot fails to process template with special chars
Software Versions
-----------------

Python version: 2.7.14

Notmuch version: 0.25.1

Alot version (either a number or sha for master): 0.7rc1

Observed behavior
-----------------

Alot fails to process a template with special chars (in this case a german Umlaut):

```
DEBUG:globals:CMDLINE: compose --template=test
DEBUG:ui:search command string: "compose --template=test"
DEBUG:__init__:mode:search got commandline "compose --template=test"
DEBUG:__init__:ARGS: [u'compose', u'--template=test']
DEBUG:__init__:cmd parms {'spawn': None, 'sender': None, 'attach': None, 'cc': None, 'omit_signature': False, 'rest': [], 'bcc': None, 'to': None, 'template': u'test', 'subject': None}
DEBUG:envelope:TEMPLATE: None
DEBUG:envelope:GoT: """
From: varac@...
To: some@oneel.se
Subject: special chars

Sadly, alot fails to process special chars like this: ö.

"""
DEBUG:globals:No encryption by default, encrypt_by_default=none
INFO:envelope:editable headers: set(['To', 'From', 'Subject'])
  File "/home/varac/.local/lib/python2.7/site-packages/twisted/internet/defer.py", line 321, in addCallback
    callbackKeywords=kw)
  File "/home/varac/.local/lib/python2.7/site-packages/twisted/internet/defer.py", line 310, in addCallbacks
    self._runCallbacks()
  File "/home/varac/.local/lib/python2.7/site-packages/twisted/internet/defer.py", line 653, in _runCallbacks
  File "/home/varac/.local/lib/python2.7/site-packages/alot-0.7.0.dev0-py2.7.egg/alot/ui.py", line 681, in call_apply
  File "/home/varac/.local/lib/python2.7/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/home/varac/.local/lib/python2.7/site-packages/alot-0.7.0.dev0-py2.7.egg/alot/commands/envelope.py", line 369, in apply
    content = '%s\n%s' % (headertext, bodytext)
exceptions.UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 54: ordinal not in range(128)
```

Expected behavior
-----------------

Alot should not choke on any unicode chars/special chars/Umlauts in a mail template.

three-comrades:
Fixes #1198 for me.

Two questions:
- Is it okay to just assume that files on Linux are encoded as 'utf-8', or should we guess the encoding (either by reading the file, or by following the hints given by environment variables)?
- The template string is passed to [parse_template](https://github.com/pazz/alot/blob/d03cd3f582818651157aa9588e5723cdaa81dc3e/alot/db/envelope.py#L281) which seems to require byte strings, such that headers like "Subject: =?iso-8859-1?q?p=F6stal?=" can be decoded later. But I don't think anyone would use a template file like that, such that it should be okay to decode the file when it's opened.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -770,8 +770,11 @@ class ComposeCommand(Command):
                           priority='error')
                 return
             try:
-                with open(path) as f:
-                    self.envelope.parse_template(f.read())
+                with open(path, 'rb') as f:
+                    blob = f.read()
+                encoding = helper.guess_encoding(blob)
+                logging.debug('template encoding: `%s`' % encoding)
+                self.envelope.parse_template(blob.decode(encoding))
             except Exception as e:
                 ui.notify(str(e), priority='error')
                 return
--- a/tests/commands/global_test.py
+++ b/tests/commands/global_test.py
@@ -18,6 +18,7 @@
 
 from __future__ import absolute_import
 import os
+import tempfile
 
 from twisted.trial import unittest
 from twisted.internet.defer import inlineCallbacks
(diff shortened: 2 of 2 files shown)


varac:
@three-comrades  wow, great - I mean the quick pull req, I can't review of judge code quality here.

alot/commands/globals.py
```diff
@@ -770,7 +771,7 @@ def apply(self, ui):
                           priority='error')
                 return
             try:
-                with open(path) as f:
+                with codecs.open(path, encoding='utf-8') as f:
```
dcbaker: We use magic for making educated guesses about encoding, and we have a wrapper in  `alot.helpers.guess_encoding`.

something like this probably:
```python
with open(path, 'rb') as f:
     blob = f.read()
encoding = helpers.guess_encoding(blob)
self.envelope.parse_template(blob.decode(encoding))
```

dcbaker:
It would alos be nice if we had a test case for this.

three-comrades:
I used your suggestion and just added a debug logging statement for the encoding.

> It would alos be nice if we had a test case for this.

I added a test that makes sure that the template is not loaded as a byte string, which was the root cause of the crash.

dcbaker:
looks good to me, thanks for making those changes.

varac:
I can confirm that this work for me, great!

PR #1204, opened 2018-02-18
Fix #1191, lockup after running external command.

### LINKED ISSUE #1191: interface is frozen when returning from external calls (editor, shellescape, ...)
Software Versions
-----------------

Python version: 2.7

Notmuch version: 0.25.2

Alot version (either a number or sha for master): 0.6

It happens on the current stable (*v.0.6*) and the one from git (*v.0.7.0dev*)

Observed behavior
-----------------

Before anything, thanks a lot of alot, it's an awesome mua, love it!

This happened recently on both Arch linux and Debian stretch. When using alot, and calling external commands (through the `editor_cmd` config option or through `shellescape` for example), the interface is not responding anymore to any keystroke when coming back from the external call.

Here's my two *external* callers:

* `editor_cmd = 'vim -c "set textwidth=80" -c "set wrap" -c "set spell" -c "set nocp"'`
* `shellescape --refocus=true '<some-script>'; refresh`. 

Both of these were working fine.

Moreover, the entire interface is working perfectly as long as I don't call any of the two above.

Test
-------------------------

Here is the minimal setup I used to test it:

```
mkdir -p /tmp/mails ~/.config/alot
```

notmuch config (`~/.notmuch-config`)
```
[database]
path=/tmp/mails

[user]
name=test
primary_email=test@test.com

[new]
tags=unread;inbox;
ignore=

[search]
exclude_tags=deleted;spam;

[maildir]
synchronize_flags=true
```

alot config (`~/.config/alot/config`):
```
editor_cmd = 'vim -c "set textwidth=80" -c "set wrap" -c "set spell" -c "set nocp"'

[accounts]
    [[test]]
        realname  = test
        address   = test@test.com
        sent_box  = maildir:///tmp/mails/
        sent_tags = inbox,
        sendmail_command = echo "ok"
```

walkthrough:

* init notmuch: `notmuch new`
* start alot: `alot`
* type `m` to compose new email, fill fields and enter something
* exit editor
* everything is frozen (even `CTRL+c` doesn't work)

I'm not sure where to look or how to debug that, nothing is displayed in the logs. Any help is appreciated.

tomjnixon:
See #1191

It would be nice to add tests for this sort of thing, but the current test suite seems to completely mock out urwid, so adding tests for the user interface would be quite complex and probably brittle.


tomjnixon:
Yeah, maybe. The tests for vterm look interesting for this, but it seems like you might as well use something like pyte instead, whose sole purpose is to be a python terminal emulator.

With that you could probably implement something that sent input into stdin, then waited for a regex to match the screen contents in a loop. I just worry that getting the timing right would be hard and likely to cause mysterious failures if it goes wrong.

While I think about it, there is similar code to this in alot/commands/thread.py, and we should check if this code is reqired:

https://github.com/pazz/alot/blob/master/alot/commands/globals.py#L291-L293

I'll take a look at those later.

tomjnixon:
Fixed the other instances of this, which were also affected.

291-293 are required to handle terminal resizes while an external programme is running.

Moved the screen stop/start code into a context manager (`with ui.paused(): ...`), which tidies up the error handling if an exception occurs while the UI is stopped, and adds the terminal resize check to :pipeto and :pyshell.

Had to modify the mocking of the UI in tests to support this. Switching to a MagicMock would also work, but this gives somewhere to put smarter UI mocks.

Issue #1011, opened 2017-01-27
Add a querylist buffer to display saved queries (aka saved search)


lucc:
Since recent versions notmuch offers the `query:your_saved_query_name` syntax ([docs](https://notmuchmail.org/manpages/notmuch-search-terms-7/)). We should have a buffer that works similar to the bufferlist and taglist buffers. It should list all defined queries from the notmuch database, on select the corresponding `search query:foo` should be executed to open the search buffer for that query.

TODO:
* check if the queries can be retrieved with the python bindings
* implement it ;)

three-comrades:
This is something I'd love to use - are you still working on this? If not, I'd like to take a shot at it.

Even if the saved queries can't be retrieved with the python bindings (I guess that's still true - at least I didn't see anything regarding this topic in their NEWS file), they could just be read from the notmuch config file, no?

pazz:
Quoting Julian Mehne (2017-12-06 22:58:17)
> This is something I'd love to use - are you still working on this? If not, I'd
> like to take a shot at it.

Go for it :)

> Even if the saved queries can't be retrieved with the python bindings (I guess
> that's still true - at least I didn't see anything regarding this topic in
> their NEWS file), they could just be read from the notmuch config file, no?

Yep

lucc:
No it is not possible to read them from the config file. The<y are stored in the database. I started a upstream patch to add support for setting and getting these queries from the database. But at the moment this patch is stalled as I was not very motivated for a while. And as this PR is dependent on the upstream patch it is also stalled.

The upstream patch starts at id:7cb0da4d17891d1284b14dbdbe116c65dfaf0195.1496596853.git.l-m-h@web.de in the notmuch mailing list.

The main problem was the naming scheme (the code is finished and is simple). Maybe I can propose another version upstream.

three-comrades:
> No it is not possible to read them from the config file.

Meh, you're right, I misread the man page ... 

Hm, okay - I think I'll just go ahead and for a proof-of-concept I'll just read the queries from the command line for now. If you find the time it would of course be great if you can propose another version.

lucc:
I posted an updated version of the patch upstream: https://notmuchmail.org/pipermail/notmuch/2017/025921.html

lucc:
The upstream patch landed: https://notmuchmail.org/pipermail/notmuch/2017/026020.html

notmuch 0.25.3 was released recently so it should be in the .4 release.

pazz:
[...]

PR #1055, opened 2017-06-01
Py3k or Bust!

dcbaker:
I thought I'd drop this here for others to look at. It's still pretty cobbled together ATM, but it does allow alot to start with python 3.5.

PR #1206, opened 2018-02-21
Named query

elenril:
Slightly polished patches from Julian, as he said in #1011 he has no time for them. The changes done by myself are:
- to avoid confusion between "query" (any notmuch search), "saved search" (a notmuch-emacs feature which this intends to mimic) and "named query" (the official notmuch name for this), I tried to consistently use the "named query" terminology everywhere.
- slightly fancier formatting of the buffer, displaying the number of messages (total and unread)

There are of course plenty other possible improvements, but I think this can be useful enough as it is.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,6 @@
+next:
+* feature: Add a new 'namedqueries' buffer type for displaying named queries.
+
 0.7:
 * info: missing html mailcap entry now reported as mail body text
 * feature: Allow regex special characters in tagstrings
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -19,7 +19,8 @@ from alot.commands import CommandParseError, COMMANDS
 from alot.utils import argparse as cargparse
 
 
-_SUBCOMMANDS = ['search', 'compose', 'bufferlist', 'taglist', 'pyshell']
+_SUBCOMMANDS = ['search', 'compose', 'bufferlist', 'taglist', 'namedqueries',
+                'pyshell']
 
 
 def parser():
--- a/alot/buffers.py
+++ b/alot/buffers.py
@@ -16,6 +16,7 @@ from .walker import PipeWalker
 from .helper import shorten_author_string
 from .db.errors import NonexistantObjectError
 from .widgets.globals import TagWidget
+from .widgets.namedqueries import QuerylineWidget
 from .widgets.globals import HeadersList
 from .widgets.globals import AttachmentWidget
 from .widgets.bufferlist import BufferlineWidget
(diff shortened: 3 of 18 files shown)


lucc:
setup.py should require the correct notmuch version! This feature will most certainly not work with notmuch 0.13.

pazz:
yes: I'm quite sure that this needs notmuch (bindings) v >= `0.26`.
the `get_configs` method was changed in December 2017, and this January was the following release (0.26).. Generally I'm happy to push something like this, as it is a n often requested feature. Of course we'd have to fix all the minor issues the codacity-bot found :)

three-comrades:
If I remember correctly, the main reason why I didn't open a PR was that I felt the new buffer type should share some code with tag lists or another buffer type I don't remember and that my proof of concept duplicated too much code.

elenril:
Quoting Lucas Hoffmann (2018-02-21 11:42:53)
> setup.py should require the correct notmuch version! This feature will most certainly not work with notmuch 0.13.

Right, though as the feature is rather new it seems better to support it
conditionally, rather than require everyone to upgrade.

-- 
Anton Khirnov

elenril:
Quoting Julian Mehne (2018-02-21 17:07:10)
> If I remember correctly, the main reason why I didn't open a PR was
> that I felt the new buffer type should share some code with tag lists
> or another buffer type I don't remember and that my proof of concept
> duplicated too much code.

My impression is that there's no so much code to begin with, so trying
to deduplicate is more trouble than it's worth. Especially since it's
likely to diverge more when/if more functionality is added.

Though if anyone has specific suggestions I'm listening.

-- 
Anton Khirnov

elenril:
Quoting Patrick Totzke (2018-02-23 10:22:20)
> I'm against "conditional" dependencies I must say. There is nothing
> wrong with having the master branch depend on very recent versions of
> dependencies such as notmuch. If you really want to use an outdated
> release you can do that at the expense of newer features.

ok, I'll just bump the dependency then

> Apart from this, notice that this PR will ultimately go into an
> upcomming release, which will be based on the python3 branch developed
> in #1055. So it is likely that the code here will have to be
> refactored in the end..

what does that mean precisely for this PR?

-- 
Anton Khirnov

pazz:
> ok, I'll just bump the dependency then

cheers

> what does that mean precisely for this PR?

only that even making the necessary fixes suggested by the bot and tests, I will
not immediately push it. At the moment, the python3 branch has priority and that
branch may make it necessary to make additional changes here later.
I just wanted to warn you about that.
Thanks for the PR btw :)

pazz:
I think this is a good start. In order to run it on master (now python3) i needed a [small hotfix](https://github.com/pazz/alot/commit/b9d35b5cf3f66cae747a0c52e47a2c0dadd58152), which should be added here.
Some more thoughts: 

- we should also update the themes in alot/extra to include the stuff for this buffer type, and mention (at least in the release mail) that people need to update their theme.
- perhaps there should be some way to add/remove named queries from within alot. Search-mode commands?

pazz:
I have rebased and split this branch to push it along. I'll retire this PR in favor of #1256 here.
Thanks!

PR #1209, opened 2018-02-23
RFC: Add message subject to the message summary widget.

elenril:
Currently, the folded message summary in thread mode only displays the sender and the date. In threads with many different subjects (such as patch series sent with git-send-email) it is very useful to see the subject as well. I've used this patch in my local tree for quite a while and would like to hear opinions on upstreaming it.

Some possible tweaks:
- make it configurable?
- only show the subject when it's different from the thread subject (perhaps also filtering out the "Re:"s at the beginning)?

### DIFF
--- a/alot/widgets/thread.py
+++ b/alot/widgets/thread.py
@@ -58,9 +58,14 @@ class MessageSummaryWidget(urwid.WidgetWrap):
         urwid.WidgetWrap.__init__(self, line)
 
     def __str__(self):
+        mail = self.message.get_email()
+
+        subj = mail.get_all('subject', [''])
+        subj = ','.join([decode_header(s, normalize = True) for s in subj])
+
         author, address = self.message.get_author()
         date = self.message.get_datestring()
-        rep = author if author != '' else address
+        rep = '%s: %s' % (author if author != '' else address, subj)
         if date is not None:
             rep += " (%s)" % date
         return rep


lucc:
A general not on "making things configurable": I also like my tools to be configurable. But I also like my tools to be simple. And maybe even exploreable. So in case we want to make this configurable I suggest the following: Let's *not* introduce a simple boolean option `show_subject_in_summary` but instead let us introduce a format string option `message_summary_format`. Pythons `str.format` method can take keyword arguments to interpolate named format placeholders which I think is ideal for this. This would also solve the issue for possibly all future requests about whatever you want (message size, mime type, attachment count, ...).

About the actual problem and solution: I neither mind nor care ;)

mjg:
I'm using a similar change locally (for the same reasons). Before working on this: There are several existing routes that could be followed. Either format/hook based (like `timestamp_format` and `timestamp_format()`) or theme base (like `seach->threadline`). While the latter requires some refactoring, it seems that the thread widget could benefit from the same type of theming that the search widget has. In fact, selecting a thread means zooming into a line of the search widget, and it would be nice to theme them analogously. What do you think?

pazz:
@mjg I agree that the most general way would be to duplicate the logic from [tread line theming](https://alot.readthedocs.io/en/latest/configuration/theming.html#search-mode-thread-ines).

The easiest to implement, and perhaps sufficient would be to add a hook that takes an `alot.db.Message` and returns a summary string.
Hard-coding only the subject line as proposed here seems too restricted to me.
I will close this PR for that reason; feel free to open a new one that implements one of the above mentioned routes.

PR #1210, opened 2018-02-23
Run tests with python3 on travis

lucc:
This is designed to *not break* even though we do not yet support py3.  One can mark travis jobs as "allowed failures".   As such it can be merged right now even before any actual porting to py3 has happend.

With this we can track the move to py3 on travis and it hopefully helps with #1055.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ python:
   # We can add more version strings here when we support other python
   # versions.
   - "2.7"
-  - "3.3"
-  - "3.4"
   - "3.5"
   - "3.6"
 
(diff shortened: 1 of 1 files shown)


dcbaker:
I'm targeting 3.5 and 3.6 currently, and not bothering with 3.3 or 3.4. I don't know if we really need to support versions that old, and I'd really like to no (asyncio is much better in 3.5 than i 3.4, and doesn't exist in 3.3 and I'd really like to use asyncio instead of twisted), also type annotations are pretty nice and they're new in 3.5

PR #1218, opened 2018-03-03
Tag replied passed

mjg:
These patches are meant to help the sync from  tags to maildir flags to IMAP flags for the states "replied" and "passed": make alot set those tags so that notmuch updates the maildir flags accordingly and IMAP sync software can pick them up from there.

I do think they make good defaults (just like "sent"); backwards-compatibility would require an empty tag list, though.

### DIFF
--- a/alot/defaults/alot.rc.spec
+++ b/alot/defaults/alot.rc.spec
@@ -324,6 +324,9 @@ thread_focus_linewise = boolean(default=True)
         # list of tags to automatically add to outgoing messages
         sent_tags = force_list(default='sent')
 
+        # list of tags to automatically add to draft messages
+        draft_tags = force_list(default='draft')
+
         # path to signature file that gets attached to all outgoing mails from this account, optionally
         # renamed to :ref:`signature_filename <signature-filename>`.
         signature = string(default=None)
--- a/alot/defaults/alot.rc.spec
+++ b/alot/defaults/alot.rc.spec
@@ -322,7 +322,7 @@ thread_focus_linewise = boolean(default=True)
         draft_box = mail_container(default=None)
 
         # list of tags to automatically add to outgoing messages
-        sent_tags = force_list(default=list('sent'))
+        sent_tags = force_list(default='sent')
 
         # path to signature file that gets attached to all outgoing mails from this account, optionally
         # renamed to :ref:`signature_filename <signature-filename>`.
--- a/alot/defaults/alot.rc.spec
+++ b/alot/defaults/alot.rc.spec
@@ -327,6 +327,12 @@ thread_focus_linewise = boolean(default=True)
         # list of tags to automatically add to draft messages
         draft_tags = force_list(default='draft')
 
+        # list of tags to automatically add to replied messages
+        replied_tags = force_list(default='replied')
+
+        # list of tags to automatically add to passed messages
+        passed_tags = force_list(default='passed')
+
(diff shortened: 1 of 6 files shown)


pazz:
how does this relate to what ’notmuch new’ already does? To me this sounds like a sensible behaviour but one that would be implemented at notmuch level instead of in the UI.
Recall that notmuch new will detect that a newly discovered mail is a response by looking at the 'Reply-to' header.

docs/source/configuration/accounts_table
```diff
@@ -100,6 +100,26 @@
...
+
+
+.. _passed-tags:
+
+.. describe:: passed_tags
+
+     list of tags to automatically add to passed messages
+
+    :type: string list
+    :default: passed,
```
dcbaker: stray comma,

alot/defaults/alot.rc.spec
```diff
@@ -327,6 +327,12 @@ thread_focus_linewise = boolean(default=True)
         # list of tags to automatically add to draft messages
         draft_tags = force_list(default=list('draft'))
 
+        # list of tags to automatically add to replied messages
+        replied_tags = force_list(default=list('replied'))
+
+        # list of tags to automatically add to passed messages
+        passed_tags = force_list(default=list('passed'))
```
dcbaker: this is going to parse down to `force_list(default=['p', 'a', 's', 's', 'e', 'd'])`, that seems off to me. I think this should be `force_list(default=['passed'])`

mjg: Again, I followed the existing style for sent and draft; but you are right so obviously that I should have noticed, sorry. (I didn't notice the effect because of explicit config.)

Being new here: Do you prefer a clean-up patch for sent and draft as a new 1/3 for this little series, or fixing it along with 1/2?

docs/source/configuration/accounts_table
mjg: I followed the existing style for sent and draft here, assuming the comma is there to enforce (or illustrate) the list type.

pazz:
Quoting Dylan Baker (2018-03-04 20:15:55)
> > @@ -327,6 +327,12 @@ thread_focus_linewise = boolean(default=True)
>          # list of tags to automatically add to draft messages
>          draft_tags = force_list(default=list('draft'))
> 
> +        # list of tags to automatically add to replied messages
> +        replied_tags = force_list(default=list('replied'))
> +
> +        # list of tags to automatically add to passed messages
> +        passed_tags = force_list(default=list('passed'))
> 
> this is going to parse down to force_list(default=['p', 'a', 's', 's', 'e',
> 'd']), that seems off to me. I think this should be force_list(default=
> ['passed'])

ConfigObj is a bit odd IIRC. It allows (see other force_list entries
in the spec) to write things like 

```
 X = force_list(default=list(From,To,Cc))
```

Notice the missing quotation marks around the inner values.
So the "list(..)" is not exactly pythons list constructor.
I believe that the suggested fix would work,
but you could alternatively just remove the "list(..)" and pass a single string
"replied". See https://github.com/pazz/alot/blob/master/alot/utils/configobj.py#L115

mjg:
I'm still not 100% sure about ConfigObj - is its list() really as_list() or python's list? In any case, leaving out list() creates properly working config and removes the comma from the auto-generated config doc. Whether that is a plus or minus - your call :)

PR #1221, opened 2018-03-05
hooks: Add a new type of hook for getting a sending address

dcbaker:
I get a good deal of mail generated by an exchange mailing list. There
is nothing in these mails to tell that they are sent to any specific
address, but I need to reply to them with a specific address, which is
not my primary. It also doesn't make sense to me to change my primary
address, since I rarely compose new mail with the address I need to
reply with in this case.

This adds a hook that is called when trying to determine which address
to reply from. This hook will be passed the email and the ui instance,
and is a coroutine (it must either return a deferred or be an
inlineCallbacks decorated callable). The later is required for the hook
to be able to use the UI for non-trivial tasks, such as prompting the
user.

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -13,7 +13,7 @@ import tempfile
 from email.utils import getaddresses, parseaddr, formataddr
 from email.message import Message
 
-from twisted.internet.defer import inlineCallbacks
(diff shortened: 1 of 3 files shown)


dcbaker:
I'm opened to other ideas on how to solve this problem, but ti seems hard to solve in a robust way without something generic like this.

PR #1222, opened 2018-03-08
provide retagprompt command in thread mode

mjg:
retagprompt in search mode allows to retag a whole thread. The new
thread mode version allows to do the same for each selected message.

### DIFF
--- a/alot/commands/common.py
+++ b/alot/commands/common.py
@@ -9,13 +9,16 @@ from .globals import PromptCommand
 
 class RetagPromptCommand(Command):
 
-    """prompt to retag selected threads\' tags"""
+    """prompt to retag selected thread's or message's tags"""
     def apply(self, ui):
-        thread = ui.current_buffer.get_selected_thread()
-        if not thread:
+        get_selected_item = getattr(ui.current_buffer, {
+                'search': 'get_selected_thread',
+                'thread': 'get_selected_message'}[ui.mode])
+        item = get_selected_item()
+        if not item:
             return
         tags = []
-        for tag in thread.get_tags():
+        for tag in item.get_tags():
             if ' ' in tag:
                 tags.append('"%s"' % tag)
             # skip empty tags
--- a/alot/commands/common.py
+++ b/alot/commands/common.py
@@ -1,0 +1,25 @@
+# Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
+# This file is released under the GNU GPL, version 3 or a later revision.
+# For further details see the COPYING file
+
+from . import Command
+
+from .globals import PromptCommand
+
+
+class RetagPromptCommand(Command):
+
+    """prompt to retag selected threads\' tags"""
+    def apply(self, ui):
+        thread = ui.current_buffer.get_selected_thread()
+        if not thread:
+            return
+        tags = []
+        for tag in thread.get_tags():
+            if ' ' in tag:
+                tags.append('"%s"' % tag)
+            # skip empty tags
+            elif tag:
+                tags.append(tag)
+        initial_tagstring = ','.join(sorted(tags)) + ','
+        return ui.apply_command(PromptCommand('retag ' + initial_tagstring))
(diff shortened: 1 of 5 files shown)


mjg:
codeclimate finds code duplication with the search mode version of the same command - rightly so, because I took that code and made it operate on messages instead of threads. I don't see an efficient path to refactoring, though.

alot/commands/thread.py
```diff
@@ -1075,6 +1076,25 @@ def apply(self, ui):
             ui.apply_command(ChangeDisplaymodeCommand(visible='toggle'))
 
 
+@registerCommand(MODE, 'retagprompt')
+class RetagPromptCommand(Command):
+
+    """prompt to retag selected message\'s tags"""
```
dcbaker: the escape is unnecessary, right?

mjg: Yes, it's just following the existing style in `search.py`.
Thinking of it: AFAIK there is no way to select multiple threads as once, so the plural _threads'_ in `search.py` appears to be incorrect.

mjg:
As for the refactoring: The two cases need to operate on threads resp. messages. I'll have a shot at it, but it seems there's a lot of potential for refactoring all over the command interface... Maybe putting this in globals.py (without registering it there) would be a start?

mjg:
OK, so this was a good excuse to brush up on decorators, especially on classes decorating classes and how to do what the `@` syntactic sugar does by hand.
The additional commit is marked WIP: If this is the road to go I would reorder the commits to first factor out from search.py to common.py, then use this in thread.py.

dcbaker:
Ah, higher order functions :)

I like this approach, it's nice and DRY.

mjg:
Here's the reordering of the commits. I also restored the original copyright since it's basically just code movement.

alot/commands/common.py
```diff
@@ -0,0 +1,29 @@
...
+
+from . import Command
+
+from .globals import PromptCommand
+
+
+class RetagPromptCommand(Command):
+
+    """prompt to retag selected thread's or message's tags"""
+    def apply(self, ui):
```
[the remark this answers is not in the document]
pazz: you can ignore this issue, codacity just doesn't seem to like the renamed parameter but this is consistent with the rest of the code. Please remove the now obsolete import suggested in the other codacity comment though.
I'll leave the merge up to @dcbaker, because we may want to postpone until the python3 series is finally merged.

mjg:
Is there anything else I can do for this PR?

pazz:
sorry for the delay. I kept everything on hold until the python3 port was ready. I've just merged this in. So, please just rebase this PR to the new master branch and I'll merge it as well.

mjg:
Apparently, something changed about the generation of docs (since py3 was merged) or something. I'll resubmit again to make travis happy.

alot/commands/common.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
+# This file is released under the GNU GPL, version 3 or a later revision.
+# For further details see the COPYING file
+from __future__ import absolute_import
```
pazz: we don't need this line anymode #1250

pazz:
it seems that the tests fail just beacause there is a typo in the test case

lucc:
The tests don't fail for me locally and they don't fail on travis. What do you mean?

pazz:
@lucc sorry my bad: I hadn't refreshed the browser and still saw old travis reports. (i've removed my earlier comment)

alot/commands/common.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (C) 2011-2012  Patrick Totzke <patricktotzke@gmail.com>
```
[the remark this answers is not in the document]
mjg: The content is (not new but) moved from thread.py, which is why I left the CR as is. I don't mind either way - should I update it to 2011-2018 in the 2nd commit where I introduce new code?

pazz:
Quoting Michael J Gruber (2018-06-20 09:12:09)
> as is. I don't mind either way - should I update it to 2011-2018 in the 2nd
> commit where I introduce new code?

I don't mind as well. Let's just keep it as it is.
Thanks

PR #1228, opened 2018-03-12
envelope: Update encryption keys when CC, To, or From is changed

### LINKED ISSUE #1227: Refresh gpg recipients after adding more recipients in evelope buffer
Software Versions
-----------------

Python 2.7.14
notmuch 0.26
alot 0.7.0dev

Observed behavior
-----------------

I compose a mail to alice, exit the edit buffer. The envelope buffer indicates the mail will get encrypted to alice (and myself). Then I decide to CC: bob, but the mail will still get only encrypted to alice and me.
I manually need to toggleencrypt twice until the mail gets encrypted to bob.

This is annoying because I already sent out mails other people couldn't decrypt from this scenario.

Expected behavior
-----------------

Whenever recipients change in envelope mode, alot should try to encrypt to new recipients without manual intervention.

dcbaker:
Currently the encryption keys will only be updated when they are
toggled, which means that if you change a Cc or To then the keys
encrypted to might be wrong, either too many keys will be encrypted to,
or not enough, or just the wrong ones.

This patches fixes this by calling set_encrypt whenever the 'To', 'Cc',
or 'From' headers are changed by set or unset.

Fixes #1227

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -418,12 +418,19 @@ class SetCommand(Command):
         self.reset = not append
         Command.__init__(self, **kwargs)
 
+    @inlineCallbacks
     def apply(self, ui):
         envelope = ui.current_buffer.envelope
         if self.reset:
             if self.key in envelope:
                 del envelope[self.key]
         envelope.add(self.key, self.value)
+        # FIXME: handle BCC as well
+        # Currently we don't handle bcc because it creates a side channel leak,
+        # as the key of the person BCC'd will be available to other recievers,
+        # defeating the purpose of BCCing them
+        if self.key.lower() in ['to', 'from', 'cc']:
+            yield utils.set_encrypt(ui, ui.current_buffer.envelope)
         ui.current_buffer.rebuild()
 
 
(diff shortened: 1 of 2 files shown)


pazz:
This is surprisingly little intrusive, cheers!
About the FIXME's: do you actually intend to deal with these some time? Because as far as I can see, not handling the BCC's is a feature. One could think of adding some ui interaction (or config setting) that lets the user resolve this choice but I'm perfectly fine with not doing this for now (as is the case atm).

Will this will also update/remove the signature key on singed mails when I change the sending address?

dcbaker:
It's been an outstanding Todo for a long time. The problem is that you
can't an encrypted message with a BCC of basically a cc since she recipient
can inspect the the encryption headers to see who received the message. The
proposed solution was to send a second copy to BCC recipients but that's
pretty invasive.

pazz:
shit, I merged this too quickly after locally testing the wrong branch.
This actually introduced the problem that every relevant header change in envelope mode will trigger a search for suitable gpg keys. But this should not happen if `encrypt` wasnt requested. In particular if you have this in your config

```
[accounts]
[[acc]]
encrypt_by_default = False
```

pazz:
This fixes the issue for me
```
diff --git a/alot/commands/utils.py b/alot/commands/utils.py
index 891f8e37..a808b0f9 100644
--- a/alot/commands/utils.py
+++ b/alot/commands/utils.py
@@ -27,6 +27,8 @@ def set_encrypt(ui, envelope, block_error=False, signed_only=False):
         to the key)
     :type signed_only: bool
     """
+    if not envelope.encrypt:
+        return
     encrypt_keys = []
     for header in ('To', 'Cc'):
         if header not in envelope.headers:
```

Issue #1230, opened 2018-03-22
regex alias addresses don't show up as "me" in the search view


vrs:
When I write emails from regex alias addresses, I see my name instead of "Me" in the search view.
I'm already working on a PR for #1177 that should fix this as well.

pazz:
I agree that a method `Account.matches` or similar, that tests if the account is responsible for a given address is a sensible approach.

Issue #1177, opened 2017-11-27
cannot reply to messages that match alias_regexp but not aliases


vrs:
Steps to reproduce:
1. locate an email that's sent to an address of yours that matches your alias_regexp, but is not your address or among the aliases.
2. hit r for reply (or do :reply)

Results: Nothing happens. With verbose logging, alot logs:
```
DEBUG:ui:Got key (['r'], [114])
DEBUG:ui:cmdline: 'reply'
DEBUG:ui:thread command string: "reply"
DEBUG:__init__:mode:thread got commandline "reply"
DEBUG:__init__:ARGS: [u'reply']
DEBUG:__init__:cmd parms {'spawn': None, 'all': False, 'listreply': None}
DEBUG:envelope:TEMPLATE: None
DEBUG:utils:unquoted header: |[snip]|
DEBUG:thread:candidate addresses: [('[snip]', '[snip]')]
DEBUG:thread:candidate addresses: [('[snip]', '[snip]')]
DEBUG:thread:candidate addresses: []
DEBUG:thread:candidate addresses: []
DEBUG:thread:candidate addresses: [('', 'vrs+label@synkretie.net')]
DEBUG:thread:match!: 'vrs+label@synkretie.net' 'vrs\+.+@synkretie.net'
DEBUG:thread:using realname: "vrs"
DEBUG:thread:using address: vrs+label@synkretie.net
DEBUG:thread:reply to: [snip]
DEBUG:utils:unquoted header: |[snip]|
ERROR:ui:Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/user/.local/lib/python2.7/site-packages/alot-0.7.0.dev0-py2.7.egg/alot/ui.py", line 679, in call_apply
    return defer.maybeDeferred(cmd.apply, self)
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1532, in unwindGenerator
    return _inlineCallbacks(None, gen, Deferred())
--- <exception caught here> ---
  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1386, in _inlineCallbacks
    result = g.send(result)
[...]

PR #1231, opened 2018-03-22
Fix alias regexp reply and related bugs

vrs:
Sorry, I would also have preferred a more minimal PR but it kind of escalated. In the process of fixing replies to alias_regexp addresses, I fixed everything else (I think) that also assumed that accounts always have a small and finite amount of addresses associated with them (this uncovered #1230) and these are the results.


vrs:
For the record: I used autopep8 for autoformatting (it's VSC's default, so a reasonably conservative choice). If that's not ok, that should probably be mentioned in the contribution guide.

alot/settings/manager.py
```diff
@@ -476,21 +476,21 @@ def get_accounts(self):
         """
         return self._accounts
 
-    def get_account_by_address(self, address, return_default=False):
+    def get_first_account_with_address(self, address, return_default=False):
```
pazz: how about `account_matching_address` ?

alot/settings/manager.py
```diff
@@ -504,7 +504,7 @@ def get_main_addresses(self):
         return [a.address for a in self._accounts]
 
     def get_addresses(self):
-        """returns addresses of known accounts including all their aliases"""
+        """returns addresses and aliases of known accounts, excluding regex aliases"""
```
pazz: This is used only for tab-completing the "from" headers in compose commands right?

alot/account.py
```diff
@@ -254,11 +255,27 @@ def __init__(self, address=None, aliases=None, alias_regexp=None,
...
-        """return all email addresses connected to this account, in order of
-        their importance"""
-        return [self.address] + self.aliases
+    def matches_address(self, address):
+        """returns whether this account knows about an email address
+
+        :param str address: address to look up
+        :rtype: bool
+        """
+        if self.address == address:
```
pazz: I wonder if we can replace all this stuff about an accounts main address / aliases and stuff it into the `alias_regex`:
A user can then only define that regex, and if she wants multiple addresses that would require to use logical OR's in that re. The test here then reduces to `re.matches`.

The problem with this is just that there may be no finite list of "main addresses" to tab-complete from..

alot/settings/manager.py
vrs: breaks with the get_* scheme of much of the rest of the class, but if that's fine with you I'll use that.

alot/account.py
vrs: There's ui.senderhistory which is already used for some completion, but that's of course empty if the client is freshly set up. I'd keep it the way it is, especially considering not every user is fluent in regex.

vrs:
According to feedback in IRC, I've split up the commits so that each is self-contained and tests run fine for each. I also use account_matching_address now.

alot/account.py
pazz: actually that `senderhistory` is cached in a file and re-read upon ui startup.

pazz: @lucc @dcbaker opinions on this? I believe that we could get rid of `account.aliases` config option in favor of `account.alias_regexp`. This would simplify the settingsmanager code at the expense of more complicated (super)user config files..

vrs: What I meant to say was that a first-time user will have no history to complete from (and first-time UX is somewhat important).

alot/commands/envelope.py
```diff
@@ -511,13 +514,16 @@ def apply(self, ui):
...
                     return
             else:
-                try:
-                    acc = settings.get_account_by_address(envelope['From'])
-                except NoMatchingAccount:
-                    envelope.sign = False
-                    ui.notify('Unable to find a matching account',
-                              priority='error')
-                    return
+                if envelope.account is None:
```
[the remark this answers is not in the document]
vrs: In principle I agree and I'll try to find a better way, but if there was a way to guarantee that an envelope had an account attached to it, that whole dance wouldn't be necessary. However, enforcing that seems like an even more invasive change than what I did here so I stayed conservative at the cost of more error checking.

vrs:
Re splitting into a test commit and a source commit: @dcbaker remarked on irc that he'd prefer to be able to bisect the commits, which seemed reasonable so I committed the relevant tests along with each fix. And couldn't you just use `git log -p tests/` and `git log -p alot/` to effectively split them into test and src changes?

I'm confused as to what commits and issues you're referring to in your last comment so I'll explain them all (current status):
* pep8 fixes: Just changes introduced by autopep8. Not especially interesting.
* attach accounts to envelopes: As detailed in #1177, the account was formerly retrieved via `determine_sender`, but then thrown away. By attaching the account to the envelope, one avoids pointless recomputation and a removes a potential way for different parts of the code to disagree about what account they're using. It's not strictly necessary anymore since account_matching_address is fixed, but not throwing away stuff we've already computed seems like a good idea in general. Also it simplifies a number of tests. Should not result in changes in workflow, except if the user enters an invalid address matching no account during compose, alot now aborts instead of quietly using the default account.
* match addresses against accounts: mostly encapsulates how accounts deal with addresses in the account class, fixes #1230, makes clear_my_address clear alias addresses as well, significantly simplifies determine_sender.
* rename get_account_by_address to account_matching_address: a refactor to make sure nothing outside Account uses address lists.

alot/commands/envelope.py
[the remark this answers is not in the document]
vrs: PS: A thing I considered was making account_matching_address return None instead of raising NoMatchingAccount, this should make all code calling it less convoluted. I refrained from doing so since it was pretty invasive and I didn't want to start a side-discussion about error handling philosophies.

alot/account.py
lucc: if the question is just about these few lines of code I would leave them.

tests/commands/utils_tests.py
```diff
@@ -57,7 +57,8 @@ def setUpModule():
 
     with gpg.core.Context(armor=True) as ctx:
         # Add the public and private keys. They have no password
-        search_dir = os.path.join(os.path.dirname(__file__), '../static/gpg-keys')
+        search_dir = os.path.join(
```
pazz: linebreaks like these are the reason why I would not blindly push everything that autopep8 proposes.

pazz:
Do you mind rebasing this to current master?

vrs:
I'll get to it, was busy the last few weeks but should have time now.

vrs:
Here you go. Had this tab open for way too long.

PR #1239, opened 2018-05-04
crypto: Use session-key to decrypt messages if present in the index

meskio:
notmuch caches the OpenPGP session keys if configured to do so. See
index.decrypt on:
https://notmuchmail.org/manpages/notmuch-config-1/

Using the cached session key decryption of messages can be done without
the need of having the private OpenPGP key. There is some speed up on
decryption, mostly notable on long encrypted threads.

---

This depends on a feature that hasn't being released yet:
https://git.notmuchmail.org/git?p=notmuch;a=commit;h=5e05f00fe59d2f2d7dceadefaab30bcc12b0e8df

Not sure how is this normally handled here. Should we wait for the release to merge it? Should we do a pull-req to develop or another branch? Do we need to wait certain time after the release of the feature on notmuch? Do you prefer me to close it now and open it once the time is right?

### DIFF
--- a/alot/db/message.py
+++ b/alot/db/message.py
@@ -54,6 +54,11 @@ class Message(object):
         self._attachments = None  # will be read upon first use
         self._tags = set(msg.get_tags())
 
+        self._session_keys = []
+        for name, value in msg.get_properties("session-key", exact=True):
+            if name == "session-key":
+                self._session_keys.append(value)
+
         try:
             sender = decode_header(msg.get_header('From'))
             if not sender:
@@ -102,7 +107,8 @@ class Message(object):
         if not self._email:
             try:
                 with open(path, 'rb') as f:
-                    self._email = utils.decrypted_message_from_bytes(f.read())
+                    self._email = utils.decrypted_message_from_bytes(
+                            f.read(), self._session_keys)
             except IOError:
                 self._email = email.message_from_string(
                     warning, policy=email.policy.SMTP)
--- a/alot/crypto.py
+++ b/alot/crypto.py
@@ -206,15 +206,22 @@ def verify_detached(message, signature):
         raise GPGProblem(str(e), code=e.getcode())
 
 
-def decrypt_verify(encrypted):
+def decrypt_verify(encrypted, session_keys=None):
     """Decrypts the given ciphertext string and returns both the
     signatures (if any) and the plaintext.
 
     :param bytes encrypted: the mail to decrypt
+    :param list[str] session_keys: a list OpenPGP session keys
     :returns: the signatures and decrypted plaintext data
     :rtype: tuple[list[gpg.resuit.Signature], str]
     :raises: :class:`~alot.errors.GPGProblem` if the decryption fails
     """
+    if session_keys is not None:
+        try:
+            return _decrypt_verify_session_keys(encrypted, session_keys)
+        except GPGProblem:
+            pass
+
     ctx = gpg.core.Context()
     try:
         plaintext, _, verify_result = ctx.decrypt(encrypted, verify=True)
@@ -226,6 +233,28 @@ def decrypt_verify(encrypted):
         sigs = e.result.signatures
 
     return sigs, plaintext
+
+
+def _decrypt_verify_session_keys(encrypted, session_keys):
+    """Decrypts the given ciphertext string using the session_keys
+    and returns both the signatures (if any) and the plaintext.
+
+    :param bytes encrypted: the mail to decrypt
+    :param list[str] session_keys: a list OpenPGP session keys
+    :returns: the signatures and decrypted plaintext data
+    :rtype: tuple[list[gpg.resuit.Signature], str]
+    :raises: :class:`~alot.errors.GPGProblem` if the decryption fails
+    """
+    for key in session_keys:
+        ctx = gpg.core.Context()
+        ctx.set_ctx_flag("override-session-key", key)
+        try:
+            (plaintext, _, verify_result) = ctx.decrypt(
+                    encrypted, verify=True)
+        except gpg.errors.GPGMEError as e:
+            continue
+        return verify_result.signatures, plaintext
+    raise GPGProblem("No valid session key", code=GPGCode.NOT_FOUND)
 
 
 def validate_key(key, sign=False, encrypt=False):
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -140,7 +140,7 @@ def _handle_signatures(original, message, params):
     add_signature_headers(original, sigs, malformed)
 
 
-def _handle_encrypted(original, message):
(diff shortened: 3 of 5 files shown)


teythoon:
I'm a bit surprised that alot does not use notmuch to get the contents of the mail.  AIUI if it would, the decryption would happen transparently.

teythoon:
So I just read some notmuch source, and it turns out that decryption is only implemented in notmuch-the-binary.  That is a shame...

So IMHO this patch papers over the fact that libnotmuch does not support decryption.  The proper thing to do is to move the decryption support from the binary to the library, and use that.

meskio:
Alot uses the notmuch library, instead of calling the notmuch binary. The mail parsing of notmuch is not in the library, but in the client side.

I was exploring how to change that, and move the email parsing into the library and it requires a bit of refactoring. I think it will be worth it to check with the notmuch devs if they will be willing to merge this change.

I also envision to add support to alot for [memory hole](https://github.com/autocrypt/memoryhole) as I'm getting more and more emails from thunderbird users with the subject encrypted. Having alot relay on notmuch to parse the memory hole headers will also be nice.

teythoon:
I understand that alot is not using the binary but the Python bindings, I worked quite a bit on the Python bindings to make alot and afew work back in the days.

If you find the time and energy, please try to push the crypto stuff into the library.  Supporting memory hole is a good idea too.  Please coordinate your efforts with Daniel (@dkg).

lucc:
About the development style of alot:

I think in order to merge code that depends on a feature in a dependency (e.g. the notmuch python bindings) this feature has to be released and the dependency has to be declared in setup.py (currently notmuch>=0.13). You can see that if you look at the dates and comments about the saved query buffer (#331, #1011, ...).

On the other hand it is perfectly fine to open a PR with a proof of concept and wait/work until it will get merged one day. See all those py3 PRs (finally #1055 will be it I think :).

dkg:
On Fri 2018-05-04 03:02:44 -0700, Ruben Pollan wrote:
> I also envision to add support to alot for [memory
> hole](https://github.com/autocrypt/memoryhole) as I'm getting more and
> more emails from thunderbird users with the subject encrypted. Having
> alot relay on notmuch to parse the memory hole headers will also be
> nice.

i'm going to post a series to notmuch Real Soon Now that deals with
protected headers, and should lay the groundwork for exposing some of
that information from libnotmuch as well.  I would appreciate any review
of that series when it hits the mailing list.

     --dkg

meskio:
Quoting dkg (2018-05-11 20:21:37)
> On Thu 2018-05-10 14:41:10 -0400, Daniel Kahn Gillmor wrote:
> > i'm going to post a series to notmuch Real Soon Now that deals with
> > protected headers, and should lay the groundwork for exposing some of
> > that information from libnotmuch as well. I would appreciate any review
> > of that series when it hits the mailing list.
> 
> that series is out now, but sadly it doesn't expose much of the
> protected headers from libnotmuch, though some of the pieces are there:
> 
> https://nmbug.notmuchmail.org/nmweb/show/20180511055544.13676-1-dkg%40fifthhorseman.net

Nice, I'll test it out.

> i had to think for a bit about why i haven't managed to expose it from
> notmuch. Here's my best understanding:
> 
> * libnotmuch exposes *messages*. The cryptographic details of a
> message are actually associated with a *file* -- if there are two
> files associated with a message, then the cryptographic details get
> blurry -- which file did they belong to?
> 
> * the notmuch cli ("notmuch show" and "notmuch reply") both deal both
> with messages *and* files -- that is, they pick a file for a given
> message, and then operate on it. So they have the ability to display
> a clear and correct cryptographic status for that particular file.
> 
> I'm not sure how to best bridge this gap. I would like to, though, and
> i'd be happy to hear suggestions for what you think the next steps might
> be.

The same way that notmuch cli picks, I guess libnotmuch can pick a file. I will 
expect most notmuch clients to do the same and not expose all the existing files 
for one message. I think picking one is a good default, as long as it list of 
file paths associated to a message is still there for the few clients will not 
use the default option.

In this case I guess the message object could have some methods to retrieve the 
parts, headers, encryption/signature status, ...

The other option I see will be to have a file abstraction on libnotmuch. So from 
message we get an iterator of files and from each file we can get the parts, 
headers and so. But I'm not sure if the extra complexity of this solution is 
worth it.

-- 
meskio | http://meskio.net/
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
 My contact info: http://meskio.net/crypto.txt
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
Nos vamos a Croatan.

meskio:
Notmuch 0.27 has being released with the needed parts for this pull-req. I agree it will be nicer to use notmuch to parse emails, but that will take some time until someone does the work to rework libnotmuch. 

How do we feel about using the session keys from the message properties for now until we get something better from notmuch? Can I improve this pull-req somehow?

meskio:
Is there any blocker to merge this into alot? Can I help to solve them?

pazz:
I don't think we should care too much about versioning in distributions but It might be good to make a release soon, then directly afterwards push this series to master...

lucc:
@pazz works for me.

meskio:
Rebased, fixed tests and added the notmuch version to the setup.py.

I think we are only missing a developer to review it.

alot/db/message.py
```diff
@@ -54,6 +54,11 @@ def __init__(self, dbman, msg, thread=None):
         self._attachments = None  # will be read upon first use
         self._tags = set(msg.get_tags())
 
+        self._session_keys = []
+        for name, value in msg.get_properties("session-key", exact=True):
+            if "session-key" == name:
```
dcbaker: Can you swap the constant and the variable here?

meskio: done

dcbaker:
I'll let someone else do the actually merge. It would be nice if someone who's better at crypto would look at it, but I think it's correct.

lucc:
I would not claim that I am better at crypto than you @dcbaker (I remember many crypto related issues and PRs from you). But I tested the code anyways and it looks good. For the record I did this to test it:

```
cp -r ~/mail ~/mail.tmp
cp $NOTMUCH_CONFIG temp-config
sed -i '/^path/s/mail/mail.tmp/' temp-config
NOTMUCH_CONFIG=temp-config notmuch reindex --decrypt=true
NOTMUCH_CONFIG=temp-config alot search is:encrypted
# and for comparison
alot search is:encrypted
```

I can confirm that the encrypted messages are displayed without gpg-agent and that long threads (7 encrypted messages) are rendered faster.

I will merge this if nobody objects.

pazz:
fine with me, thanks!

PR #1240, opened 2018-05-09
display the exception explaining why we couldn't load hooks

teto:
simply adding this ehlped tremendously finding errors in my hooks file (encoding, wrong imports etc)


alot/settings/manager.py
```diff
@@ -91,8 +91,9 @@ def read_config(self):
         hooks_path = os.path.expanduser(self._config.get('hooksfile'))
         try:
             self.hooks = imp.load_source('hooks', hooks_path)
-        except:
+        except Exception as e:
             logging.debug('unable to load hooks file:%s', hooks_path)
+            logging.debug('because %s' % e)
```
pazz: you could use [logging.exception](https://docs.python.org/2/library/logging.html#logging.Logger.exception) here..

teto:
didn't know about logging.exception. cool. I updated the commit message.

pazz:
looks great. can you squach those commits into one?

teto:
it was alread squashed.

pazz:
silly me. sorry. This looks good to me

alot/settings/manager.py
```diff
@@ -91,8 +91,8 @@ def read_config(self):
         hooks_path = os.path.expanduser(self._config.get('hooksfile'))
         try:
             self.hooks = imp.load_source('hooks', hooks_path)
-        except:
-            logging.debug('unable to load hooks file:%s', hooks_path)
+        except Exception as e:
```
lucc: You don't need the variable name when you use logging.exception() and hence you also don't need the class specification.

teto: I removed it

PR #1247, opened 2018-06-18
use importlib to load hooks code at startup

### LINKED ISSUE #1246: problem with early instantiation of settings (due to `import alot` in hooks file)
Software Versions
-----------------

Python version: 3.6.5 (Arch Linux)

Notmuch version: 0.26.2 (Arch Linux)

Alot version (either a number or sha for master): master: 0.7-127-ge14488f1

Observed behavior
-----------------

I see an exception on the terminal when starting up alot. Functionality seems to be unobstructed. It runs just fine afterwards.

I ran `alot -c ~/alot.tmp -d debug -l logfile`
This is the output I see
```
ERROR:root:unable to load hooks file:/home/luc/.config/alot/hooks.py
  File "/home/luc/vcs/alot/alot/settings/manager.py", line 93, in read_config
    self.hooks = imp.load_source('hooks', hooks_path)
  File "/usr/lib/python3.6/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 684, in _load
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/luc/.config/alot/hooks.py", line 14, in <module>
    from alot.commands.envelope import SignCommand, EncryptCommand
  File "/home/luc/vcs/alot/alot/commands/__init__.py", line 12, in <module>
    from ..settings.const import settings
ImportError: cannot import name 'settings'
inside tmp hooks file
```
The file `~/alot.tmp` looks like this:
```
hooksfile = ~/tmp/config/alot/hooks.py
[accounts]
[[x]]
address = foo@bar
realname = lucas
sendmail_command = msmtp
sent_box = maildir:///home/luc/tmp/sent
draft_box = maildir:///home/luc/tmp/draft
```

The file `~/tmp/config/alot/hooks.py` looks like this:
```
print("inside tmp hooks file")
```
In the default hooks file (`~/.config/alot/hooks.py`) I have this line: `from alot.commands.envelope import SignCommand, EncryptCommand`. If I comment that out, I don't see the error.

The logfile doesn't have any info about that error!

Expected behavior
-----------------

No stack trace.

Additional Info
-----------------

The [`imp`](https://docs.python.org/3/library/imp.html) module is deprecated since 3.4. On StackOverflow they say one should use importlib: https://stackoverflow.com/questions/67631

pazz:
imp is deprecated in favor of importlib.
fixes #1246

### DIFF
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -3,7 +3,7 @@
 # For further details see the COPYING file
 from __future__ import absolute_import
 
-import imp
+import importlib.util
 import itertools
 import logging
 import mailcap
@@ -90,7 +90,9 @@ class SettingsManager(object):
 
         hooks_path = os.path.expanduser(self._config.get('hooksfile'))
         try:
-            self.hooks = imp.load_source('hooks', hooks_path)
+            spec = importlib.util.spec_from_file_location('hooks', hooks_path)
+            self.hooks = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(self.hooks)
         except:
             logging.exception('unable to load hooks file:%s', hooks_path)
         if 'bindings' in newconfig:


pazz:
The only thing that was problematic for me when using this was this line in my hooks file:

```
    from alot.settings.const import settings
```
I had this at the top level, not in some hook. This fails (as it should, but for some reason never did before) because the hooks module is interpreted in the constructor of `SettingsManager`, which `settings` is an instance of.

lucc:
Sorry to say that but I still get the same error.  Please notice that *both* hooks files seem to be loaded for me. I'll update #1246 to make it clear.

pazz:
thanks. I am working on a solution to the second issue right now. Will merge this in and send another PR.

PR #1251, opened 2018-06-19
lazy reading of config files in SettingsManager

pazz:
This prevents SettingsManager from reading the config files right when it is instantiated and instead waits for the main module to call `read_[notmuch]config` with the right path.

This should prevent problems with accidentally reading the default config paths despite being told otherwise (via commandline options)

### DIFF
--- a/alot/settings/manager.py
+++ b/alot/settings/manager.py
@@ -29,55 +29,46 @@ DATA_DIRS = get_xdg_env('XDG_DATA_DIRS',
 
 class SettingsManager(object):
     """Organizes user settings"""
-    def __init__(self, alot_rc=None, notmuch_rc=None):
-        """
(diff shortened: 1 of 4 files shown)


pazz:
args, I forgot to update the tests..

varac:
I tried this MR and it fixes #1246 for me.

alot/settings/manager.py
```diff
@@ -29,55 +29,48 @@
...
         self._theme = None
         self._accounts = None
         self._accountmap = None
-        self.alot_rc_path = alot_rc
-        self.notmuch_rc_path = notmuch_rc
         self._notmuchconfig = None
         self._config = ConfigObj()
         self._bindings = None
-        self.reload()
+        self._configpath = None
```
lucc: This is never used.

PR #1254, opened 2018-06-21
limit chardet on large blobs

### LINKED ISSUE #1253: chardet takes way too long
For large attachments, determing the encoding with chardet takes way too long - in the order of minutes for a few megabytes.

POC patch upcoming.

mjg:
0b5f8bf1 ("require chardet", 2018-03-06) replaced the use of magic for
the determination of encodings by chardet. But chardet looks at the
whole blob which takes enormous amounts of time for large attachments.

Replace the direct call with a BytesIO stream and use a cut off to limit
the number of lines.

Fixes #1253

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -10,6 +10,7 @@ from io import BytesIO
 from io import StringIO
 import logging
 import mimetypes
+import io
 import os
 import re
 import shlex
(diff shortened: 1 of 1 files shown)


pazz:
args, I cannot find a way to add a disapproving review.
Anyways: I'd much rather do something like [this](https://github.com/pazz/alot/commit/6c68e2e1712e6a849f30f98e46921f2841f16a31) instead. It'd be even faster because it avoids calling chardet alltogether

mjg:
Yes, as mentioned in #1253, this was meant as a proof of concept/stop gap measure. Using the emails module is much better. That solves my problem, too. And I guess we'll find a way to make the tests pass.

I don't see a way to cancel my pull request, but I guess closing it without merging should do the job.

pazz:
excellent, thanks Michael!

PR #1256, opened 2018-06-23
New buffer type for notmuch's named query strings

pazz:
This adds a new buffer that displays named queries similar to the taglist mode.
It so far offers only one command: select, which opens a new search buffer for the selected named query.

This is a rebased , fully splitted and refactored version of #1206, for easy review.
TODO:

- [x]  update the theme files in `alot/extra/themes`
- [x] check in auto-generated docs
- [x] add commands to manimulate named queries in this buffer
- [x] add command to save a query in search buffers
- [x] tab completion for named queries in prompts
- [ ] tests
To test this, add some named queries manually via the notmuch CLI:

```
notmuch config set query.foo "bar"
```

### DIFF
--- a/alot/buffers/namedqueries.py
+++ b/alot/buffers/namedqueries.py
@@ -1,0 +1,78 @@
+# Copyright (C) 2011-2018  Patrick Totzke <patricktotzke@gmail.com>
+# This file is released under the GNU GPL, version 3 or a later revision.
+# For further details see the COPYING file
+import urwid
+
+from .buffer import Buffer
+from ..settings.const import settings
+from ..widgets.namedqueries import QuerylineWidget
+
+
+class NamedQueriesBuffer(Buffer):
+    """lists named queries present in the notmuch database"""
+
+    modename = 'namedqueries'
+
+    def __init__(self, ui, filtfun):
+        self.ui = ui
+        self.filtfun = filtfun
+        self.isinitialized = False
+        self.querylist = None
+        self.rebuild()
+        Buffer.__init__(self, ui, self.body)
+
+    def rebuild(self):
+        self.queries = self.ui.dbman.get_named_queries()
+
+        if self.isinitialized:
+            focusposition = self.querylist.get_focus()[1]
+        else:
+            focusposition = 0
+
+        lines = []
+        for (num, key) in enumerate(self.queries):
+            value = self.queries[key]
+            count = self.ui.dbman.count_messages('query:"%s"' % key)
+            count_unread = self.ui.dbman.count_messages('query:"%s" and '
+                                                        'tag:unread' % key)
+            line = QuerylineWidget(key, value, count, count_unread)
+
+            if (num % 2) == 0:
+                attr = settings.get_theming_attribute('namedqueries',
+                                                      'line_even')
+            else:
+                attr = settings.get_theming_attribute('namedqueries',
+                                                      'line_odd')
+            focus_att = settings.get_theming_attribute('namedqueries',
+                                                       'line_focus')
+
+            line = urwid.AttrMap(line, attr, focus_att)
+            lines.append(line)
+
+        self.querylist = urwid.ListBox(urwid.SimpleListWalker(lines))
+        self.body = self.querylist
+
+        self.querylist.set_focus(focusposition % len(self.queries))
+
+        self.isinitialized = True
+
+    def focus_first(self):
+        """Focus the first line in the query list."""
+        self.body.set_focus(0)
+
+    def focus_last(self):
+        allpos = self.querylist.body.positions(reverse=True)
+        if allpos:
+            lastpos = allpos[0]
+            self.body.set_focus(lastpos)
+
+    def get_selected_query(self):
+        """returns selected query"""
+        return self.querylist.get_focus()[0].original_widget.query
+
+    def get_info(self):
+        info = {}
+
+        info['query_count'] = len(self.queries)
+
+        return info
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -1004,7 +1004,7 @@ class ReloadCommand(Command):
                           'help': 'postpone a writeout to the index'}),
         (['alias'], {'help': 'alias to use for query string'}),
         (['query'], {'help': 'query string to store',
-                     'nargs': argparse.REMAINDER})
+                     'nargs': '+'})
     ],
     help='store query string as a "named query" in the database')
 class SaveQueryCommand(Command):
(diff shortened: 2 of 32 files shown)


lucc:
I can not start alot on this branch. I get this exception:
```
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "alot/__main__.py", line 14, in <module>
    from alot.ui import UI
  File "/home/luc/vcs/alot/alot/ui.py", line 14, in <module>
    from .buffers import BufferlistBuffer
  File "/home/luc/vcs/alot/alot/buffers/__init__.py", line 11, in <module>
    from .namedqueries import NamedQueriesBuffer
  File "/home/luc/vcs/alot/alot/buffers/namedqueries.py", line 8, in <module>
    from ..widgets.namedqueries import QuerylineWidget
ModuleNotFoundError: No module named 'alot.widgets.namedqueries'
```

mjg:
The rebase lost the widget `alot/widgets/namedqueries.py`

pazz:
yep, sorry about that.
By the way, this special text widget to display pairs of strings might actually be a bit over the top: we can just as well set up ordinary `urwid.Text` widgets in the buffer. The same holds for `BufferlistBuffer`s which also have a dedicated Text widget. But perhaps we should refactor these together in another PR..

pazz:
There is more to do/fix here:
The commands to write/delete named queries need to be handeled with care because just as tag commands, they write to the notmuch database which can in fact be locked!

I'm also not entirely happy with the two getters for named queries as currently offered by the settings manager. Rather, as the notmuch bindings suggest, we can directly read the pairs as a dict from the bindings.

pazz:
I've refactored the changes to the database manager just now and also removed the problematic setters for now. This means at the moment, this PR contains a read-only version of the named queries functionality. It can only display queries, not create new ones.

To test this, add some named queries manually via the notmuch CLI:

```
notmuch config set query.foo "bar"
```

pazz:
OK, I think this is ready. For some reason travis is stuck, but the (unchanged) tests run fine locally. I could'nt be arsed to write new ones so far *cough*

pazz:
It seems that we did not have a testcase for the database manager at all so far. The last commits adds this, together with a basic test for the new functionality.

You will want to have a closer look at the `setupClass()` method and related calls as I don't have a good understanding of how tests should be set up properly!

tests/db/manager_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+from alot.db.manager import DBManager
+from alot.settings.const import settings
+from notmuch import Database
+
+from .. import utilities
+
+
+class TestDBManager(utilities.TestCaseClassCleanup):
+    """Test the get_account_by_address helper."""
```
lucc: this looks like a copy paste error

pazz: thanks. I just removed this line.

alot/buffers/namedqueries.py
```diff
@@ -0,0 +1,77 @@
...
+        self.filtfun = filtfun
+        self.isinitialized = False
+        self.rebuild()
+        Buffer.__init__(self, ui, self.body)
+
+    def rebuild(self):
+        self.queries = self.ui.dbman.get_named_queries()
+
+        if self.isinitialized:
+            focusposition = self.querylist.get_focus()[1]
```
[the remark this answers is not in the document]
dcbaker: Let's set `self.querylist = None` in the constructor, please. Not only does it fix warnings like this, but it also helps my autocompleter in vim :)

alot/commands/globals.py
```diff
@@ -553,6 +553,21 @@ def apply(self, ui):
             ui.buffer_open(buffers.TagListBuffer(ui, tags, self.filtfun))
 
 
+@registerCommand(MODE, 'namedqueries')
+class NamedQueriesCommand(Command):
+    """opens named queries buffer"""
+    def __init__(self, filtfun=lambda x: x, **kwargs):
```
dcbaker: this doesn't match the definition in the docstring, I think we should either replace the lambda with `bool`, (which will return True for any string that != "").

alot/commands/globals.py
```diff
@@ -979,3 +994,94 @@ def apply(self, ui):
...
+
+    """save alias for query string"""
+    repeatable = False
+
+    def __init__(self, alias, query=None, flush=True, **kwargs):
+        """
+        :param alias: name to use for query string
+        :type alias: str
+        :param query: query string to save
+        :type query: str
```
dcbaker: :type query: str or None

alot/widgets/namedqueries.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (C) 2011-2018  Patrick Totzke <patricktotzke@gmail.com>
+# This file is released under the GNU GPL, version 3 or a later revision.
+# For further details see the COPYING file
+
+"""
+Widgets specific to Namedqueries mode
+"""
+from __future__ import absolute_import
```
dcbaker: Don't need this anymore, this is the default behavior of python 3.

tests/db/manager_test.py
```diff
@@ -0,0 +1,48 @@
...
+        with tempfile.NamedTemporaryFile(mode='w+', delete=False) as f:
+            f.write(textwrap.dedent("""\
+                [maildir]
+                synchronize_flags = true
+                """))
+            cls.notmuch_config_path = f.name
+        cls.addClassCleanup(os.unlink, f.name)
+
+        # define an empty notmuch database in a temporary directory
+        cls.dbpath = tempfile.mkdtemp()
```
dcbaker: should probably add `cls.addClassCleanup(shutil.rmtree, cls.dbpath)` here.

dcbaker: Although, actually, we probably need to add `cls.addClassCleanup(cls.manager.kill_search_processes)` first, right?

alot/commands/globals.py
pazz: I don't quite see how a callable would match type `bool`, but ok..

tests/db/manager_test.py
pazz: it can't hurt, yes

alot/commands/globals.py
dcbaker: what I mean is, `lambda x: x` is  a function that takes a thing `T` and returns `T`, but the documentation says that the callable is a function that takes type `str` and returns type `bool`; `bool()` is a function that takes `T` and returns `bool`, a superset of what the documentation says. In most cases the lambda will work, since any type not bool that is used with a logical operator will have it's `__bool_()` method called, but there are cases where it wouldn't work, like if someone where to write `if filtfun(val) is True: ...`.

pazz:
Oh I see, that makes sense.
So Is it safe/OK to just use `bool` as type as it is in the updated
version? (If so, I'm finished with this series)

Quoting Dylan Baker (2018-07-24 22:15:26)

dcbaker:
yeah, replacing `lambda x: x` with `bool` will be safe.

```python
>>> bool('')
False
>>> bool('Foo')
True
```

dcbaker:
LGTM! thanks.

PR #1259, opened 2018-07-01
Clean up old workaround?

pazz:
Should we remove this workaround for [python issue](https://bugs.python.org/issue14983) that was solved upstream?

WARNING: This is a RFC. I have not tested this!

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -603,19 +603,6 @@ def email_as_string(mail):
     g = Generator(fp, mangle_from_=False, maxheaderlen=78)
     g.flatten(mail)
     as_string = RFC3156_canonicalize(fp.getvalue())
-
-    if isinstance(mail, MIMEMultipart):
-        # Get the boundary for later
-        boundary = mail.get_boundary()
-
-        # Workaround for http://bugs.python.org/issue14983:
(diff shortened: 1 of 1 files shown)


lucc:
I am in favour of this (compare #1260 :) but we should introduce some tests for this first in order to assure ourselves that we will not break anything.

pazz:
I'm not sure how to test this to be honest. We could add add complicated mime mail, sign it, and verify that some other email client is happy with the format. But that could be a one-off test and need not be part of our test suite.
Having read around a bit in the cited issue for python, it seems that this patch has actually made it into all important stdlib versions in 2014.

dcbaker:
Honestly I think this is fine as-is, we shouldn't be testing that code in another source tree is doing the right think. Hopefully they're testing their own code.

lucc:
@pazz As I am a collaborator on this repo I see the "delete branch" button. I was tempted to click it but first the question: do you want to keep old branches around (there are currently 42 in this repo) or should we delete old branches?

pazz:
not at all, I always forget this. You're welcome to click away old
branches once they are merged :)


Quoting Lucas Hoffmann (2018-08-02 11:31:56)

PR #1260, opened 2018-07-01
Remove pre py3 monkey patch

lucc:
I just found this thing by chance but we can now get rid of it.


lucc:
I rebased and now travis pushes the build status back to github :tada:

pazz:
I believe that this is also part of #1270 23cda03.
I'm ok with merging this, and my named queries stuff first, then dcbaker can rebase that branch accordingly.

PR #1261, opened 2018-07-05
widgets/search: Don't try to sort an iterator

dcbaker:
We currently call dict_keys.sort(), which doesn't work because it
doesn't exist. The whole function is kinda strange anyway, since there's
a lot of work done in general that only applies in once case.


pazz:
I'll close this in favor of #1264

PR #1262, opened 2018-07-06
fix and refactor threadline widgets

pazz:
The first commit fixes wide characters in search mode:

This fixes an issue with utf-8 wide characters such as Kanji, fow which urwid needs extra space, and consequently adds additional rows when packing such Textwidgets into Columns as we do in threadlines.

Before:
before

After:
after


pazz:
closed in favor of #1264  (which includes this fix)

PR #1263, opened 2018-07-08
Open signatures as binary

kyrias:
Otherwise it will think that it's octet-stream and won't attach the signature.  Also use file-magic to detect the encoding, because I've had problems with chardet incorrectly identifying my signatures as ISO-8859-1.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -844,7 +844,7 @@ class ComposeCommand(Command):
                     self.envelope.attach(sig, filename=name)
                     logging.debug('attached')
                 else:
-                    with open(sig) as f:
+                    with open(sig, 'rb') as f:
                         sigcontent = f.read()
                     mimetype = helper.guess_mimetype(sigcontent)
                     if mimetype.startswith('text'):


pazz:
These seem to be unrelates issues. Nevertheless,
For your first patch: What's the problem with reading the content as a regular python string?
I do not seem to have any problems with attacheing signatures which are octet-stream. How would I reproduce this? Would you be able to write a test case for thie issue?

For the second patch: I think we once had a good reason to drop magic in favor of chardet, in 0b5f8bf143e76be7ac989e8c8ec6e4a7dc08a78a. I don't know if chardet behaves particularly badly now that we've moved on to python3 but I believe that this should be unrelated.
@dcbaker: as the person who introduced this change, do you have an opinion here?

kyrias:
It only adds it if the mimetype starts with 'text', so the code flat out won't do anything if it's octet-stream...

```
                     mimetype = helper.guess_mimetype(sigcontent)
                     if mimetype.startswith('text'):
                         sigcontent = helper.try_decode(sigcontent)
                         self.envelope.body += '\n' + sigcontent
```

But yeah, chardet is just constantly completely wrong for me now, so something needs be done about it at least.

pazz:
I presume that #1266 obsoletes your second patch?
If so, please remove it here and I will push patch 1 only.

kyrias:
Done.

pazz:
thanks

Issue #1265, opened 2018-07-10
encoding issue


pazz:
I have a mail(thread) that fails to open with the error trace below.

```
  File "/home/pazz/projects/alot/alot/widgets/thread.py", line 238, in _get_body
    bodytxt = self._message.accumulate_body()
  File "/home/pazz/projects/alot/alot/db/message.py", line 264, in accumulate_body
    bodytext = extract_body(eml)
  File "/home/pazz/projects/alot/alot/db/utils.py", line 459, in extract_body
    body_parts.append(string_sanitize(remove_cte(part, as_string=True)))
  File "/home/pazz/projects/alot/alot/db/utils.py", line 392, in remove_cte
    return helper.try_decode(raw_payload)
  File "/home/pazz/projects/alot/alot/helper.py", line 408, in try_decode
    return blob.decode(guess_encoding(blob))
builtins.UnicodeDecodeError: 'cp949' codec can't decode byte 0xb7 in position 1075: illegal multibyte sequence
```

This seems to be `chardet` related as the message has an appendix, and the patch referenced at #1253 solves this issue here for me.

**Software Versions**
- Python version: 3.6.5
- Alot version: master

dcbaker:
closed via: #1266

PR #1266, opened 2018-07-10
Message from bytes

mjg:
Fix for #1253 and #1265

### DIFF
--- a/alot/db/message.py
+++ b/alot/db/message.py
@@ -101,7 +101,7 @@ class Message(object):
         if not self._email:
             try:
                 with open(path, 'rb') as f:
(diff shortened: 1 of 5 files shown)


mjg:
I guess I should rebase this branch on current master in order to kick-off travis-ci. Let's see...

_Cognitive complexity_? You kiddin' me, codeclimate?

pazz:
Don't worry about codeclimate. It's just nagging about old "issues" when you refactor code. We'll ignore this.

dcbaker:
I think this will be fine, since it worked in python 2. It probably will break some very stupid terminals, but I'm okay with that.

PR #1269, opened 2018-07-14
alot/helper: Fix parse_mailto for python 3

kyrias:


kyrias:
I've rephrased the commit message now, but I don't think there's any way to actually test for *this* bug?  The only way would be to grep the source and check whether the py2 paths are used instead of the py3 paths.  I did however add a basic parse_mailto test.

PR #1270, opened 2018-07-18
Replace twisted with asyncio

dcbaker:
I got a bit of free time, and sat down to work on something I'd really wanted to work on since we started the transition to python 3, using asyncio.

I think there are some benefits for us to using asyncio instead of twisted. Asyncio is part of the language, and it gives us nice language features like `async def` and `await`. On the downside it doesn't have as many nice convenience functions as twisted does. It also removes a pretty big, heavy dependency. Twisted is pretty cool, and you can do a lot of cool stuff with it, but we're basically just using the event loop and deferreds.

This series has been lightly tested, it does pass all of the unittests, although there are some obvious holes in the coverage of our test suite :/. I'm currently dog fooding it, but if anyone sees obvious mistakes or problems I'd be glad to start working through them sooner than later (as time at $job and $life allows).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -119,7 +119,7 @@ install:
   - |
     set -e
     if [[ $JOB = docs ]]; then
-      pip install configobj twisted python-magic urwidtrees
+      pip install configobj python-magic urwidtrees
       # Mock all "difficult" dependencies of alot in order to be able to import
       # alot when rebuilding the documentation.  Notmuch would have to be
       # installed by hand in order to get a recent enough version on travis.
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,7 @@
 0.8:
 * Port to python 3. Python 2.x no longer supported
 * feature: Add a new 'namedqueries' buffer type for displaying named queries.
+* feature: Replace twisted with asyncio
 
 0.7:
 * info: missing html mailcap entry now reported as mail body text
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -6,6 +6,9 @@ import logging
 import os
 import sys
 
+from twisted.internet import asyncioreactor
+asyncioreactor.install()
+
 import alot
 from alot.settings.const import settings
 from alot.settings.errors import ConfigError
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -6,9 +6,6 @@ import logging
 import os
 import sys
 
-from twisted.internet import asyncioreactor
-asyncioreactor.install()
-
 import alot
 from alot.settings.const import settings
 from alot.settings.errors import ConfigError
--- a/alot/account.py
+++ b/alot/account.py
@@ -314,14 +314,13 @@ class Account(object):
             return self.store_mail(self.draft_box, mail)
 
     @abc.abstractmethod
-    def send_mail(self, mail):
+    async def send_mail(self, mail):
         """
         sends given mail
 
         :param mail: the mail to send
         :type mail: :class:`email.message.Message` or string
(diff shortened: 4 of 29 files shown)


pazz:
When I try to respond to an email with this, my editor is not spawned as usual and instead the terminal window is just blank. I have both `editor_spawn`, and `editor_in_thread` undefined in the config, which means they default to false. See the log below.

```
DEBUG:globals:No encryption by default, encrypt_by_default=none
INFO:envelope:editable headers: {'Subject'}
DEBUG:globals:using editor_cmd: vim -c "set spell"
DEBUG:globals:{'spawn: ': False, 'in_thread': False}
DEBUG:globals:{'spawn': False}
DEBUG:globals:cmdlist: ['vim', '-c', 'set spell', '/tmp/alot.u8kl93ug.eml']
INFO:globals:calling external command: ['vim', '-c', 'set spell', '/tmp/alot.u8kl93ug.eml']

```

dcbaker:
Okay, I've addressed the editor thing, it has to do with using the `helper.call_cmd`; since that's kinda orthogonal to the point of this merge request I've dropped those patches. I've also fixed a few sending bugs I ran into. I'll see what I can do about making pep8 happy.

dcbaker:
Okay, I've fixed all but one of the pep8 issues, the one left is pretty silly it's an 81 character line, and the last three characters are `)`, I think a newline hurts the readability rather than helping. The other issues are me adding fixme comments.

dcbaker:
And I've fixed a whitespace error I introduced in my fixups...

pazz:
question: do you intent to fix the Todo/xxx/fixme's as part of this series?

dcbaker:
no. I think all of those are worth fixing, but they're not related to the twisted -> asyncio transition. I have some patches for some of those I'm planning to submit if/when this lands.

pazz:
Thanks, that makes sense as this series is large enough as it is. I'll ok the failing codeclimate check.  @lucc: let's testrun this for a few days..

lucc:
I already started using it.  The first thing I noticed was the delay on startup.  My initial command opens three buffers and then switches to the first one (and pops up a notification). I can see all the buffers flicker by for a sort moment.  That's not really a problem, only a glitch but it wasn't there with the old implementation.

Ref:
```
initial_command = bufferlist; search query:inbox; search query:listbox; buffer 0; call hooks.version_notification(ui)
```

dcbaker:
I see the same thing in my setup. I'm not exactly sure why we have that, I don't know if twisted wasn't actually deferring while running the initial command line, or if it was faster.

lucc:
If alot is already running and I do `:search foo ; search bar` it opens both buffers after another on either branch.

dcbaker:
And the raises directive isn't correct either :)

dcbaker:
Sure.

pazz:
It seems that on this branch, calls to external programs that write to stderr result in error notifications in alot. I noticed that when I opened a PDF attachment. This opens my viewer (zathura) as expected. But when I close it again and alot resumes, I get the error notifications.

dcbaker:
- Fixed typo and a couple other sphinx syntax issues
- replace uses of `asyncio.ensure_future` with `loop.create_task`, which is recommended by python upstream

dcbaker:
Fixed the errors after running external commands, also fixed shell calls in external command

pazz:
@dcbaker: can I ask you to go through recent PRs that are ready and merge them, then rebase this branch? I think this is ready to go but I want to avoid having to rebase many other PRs as a result of just merging this one..

PR #1274, opened 2018-07-23
Make DBManager faster

dcbaker:
After looking at the asyncio branch a bit, I realized that the way the search buffers load threads is making the staggered loading problem worse. This is caused by the Pipe filling up with threads that no one ever reads because the pipe is filled until all of the threads are loaded. This is a lot of work for no reason.

On my system I see a slight drop in resident memory load, and the shared memory load is roughly half. This also makes startup faster both with asyncio and with twisted (I built this on top of asyncio, but it really doesn't depend on it, so I rebased on master)


dcbaker:
I'm marking this as RFC, since I'm thinking about other things we could do that might be better.

pazz:
what's the alternative?

dcbaker:
My other thought was to use a single worker process for all search buffers. It would be started at alot start time (before the initial commands are run) and exist communication via a Unix socket. Then each search connects, provides it's query, and receives the thread IDs. The asyncio streams mechanism seems nice. Alternatively maybe just use an async generator, just have to do performance analysis to see if it's fast enough.

pazz:
@dcbaker could you comment on this? As far as I understand, this PR is independent of #1280 and would still be an improvement over the status quo. Does it make sense to rebase this?

pazz:
I'm postponing this for the next release and move it along into milestone 1.0

PR #1275, opened 2018-07-24
Copyedit documentation

alexshpilkin:
Applies a bit of copyediting polish to the documentation. I think you might want to pull everything in (obviously), but the branch should be cherry-pickable.

There are more typos and things to edit, but I’ll limit myself to this for now. One question: what should be done with the inconsistent usage of focused/focussed? (This PR leaves things as they are.)


alexshpilkin:
Hm. Sorry. Let me fix this and rename the branch while I’m at it.

PR #1276, opened 2018-07-24
Copyedit documentation

alexshpilkin:
Applies a bit of copyediting polish to the documentation. I think you might want to pull everything in (obviously), but the branch should be cherry-pickable.

There are more typos and things to edit, but I’ll limit myself to this for now. One question: what should be done with the inconsistent usage of focused/focussed? (This PR leaves things as they are.)

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -38,16 +38,18 @@ from ..utils import argparse as cargparse
 MODE = 'global'
 
 
-@registerCommand(MODE, 'exit')
+@registerCommand(MODE, 'exit', help="shut down cleanly")
 class ExitCommand(Command):
-    """Shut down cleanly.
-
-    The _prompt variable is for internal use only, it's used to control
-    prompting to close without sending, and is used by the BufferCloseCommand
-    if settings change after yielding to the UI.
-    """
+    """Shut down cleanly."""
 
     def __init__(self, _prompt=True, **kwargs):
+        """
+        :param _prompt: For internal use only, used to control prompting to
+                        close without sending, and is used by the
+                        BufferCloseCommand if settings change after yielding to
+                        the UI.
+        :type _prompt: bool
+        """
         super(ExitCommand, self).__init__(**kwargs)
         self.prompt_to_send = _prompt
 
@@ -352,7 +354,7 @@ class PythonShellCommand(Command):
 @registerCommand(MODE, 'repeat')
 class RepeatCommand(Command):
 
-    """Repeats the command executed last time"""
+    """repeat the command executed last time"""
     def __init__(self, **kwargs):
         Command.__init__(self, **kwargs)
 
@@ -367,7 +369,7 @@ class RepeatCommand(Command):
     (['command'], {'help': 'python command string to call'})])
 class CallCommand(Command):
 
-    """Executes python code"""
+    """execute python code"""
     repeatable = True
 
     def __init__(self, command, **kwargs):
@@ -616,8 +618,8 @@ class FlushCommand(Command):
     (['commandname'], {'help': 'command or \'bindings\''})])
 class HelpCommand(Command):
 
-    """display help for a command. Use \'bindings\' to display all keybings
-    interpreted in current mode.'"""
+    """display help for a command (use \'bindings\' to display all keybindings
+    interpreted in current mode)"""
     def __init__(self, commandname='', **kwargs):
         """
         :param commandname: command to document
@@ -983,7 +985,7 @@ class MoveCommand(Command):
                       priority='error')
 
 
-@registerCommand(MODE, 'reload', help='Reload all configuration files')
+@registerCommand(MODE, 'reload', help='reload all configuration files')
 class ReloadCommand(Command):
 
     """Reload configuration."""
--- a/alot/commands/search.py
+++ b/alot/commands/search.py
@@ -136,10 +136,9 @@ RetagPromptCommand = registerCommand(MODE, 'retagprompt')(RetagPromptCommand)
                           'default': 'True',
                           'help': 'postpone a writeout to the index'}),
         (['tags'], {'help': 'comma separated list of tags'})],
-    help='flip presence of tags on this thread. A tag is considered present '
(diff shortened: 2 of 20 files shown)


alot/commands/globals.py
```diff
@@ -40,14 +40,14 @@
 
 @registerCommand(MODE, 'exit')
 class ExitCommand(Command):
-    """Shut down cleanly.
-
-    The _prompt variable is for internal use only, it's used to control
-    prompting to close without sending, and is used by the BufferCloseCommand
-    if settings change after yielding to the UI.
-    """
+    """shut down cleanly"""
```
dcbaker: Please retain the capital `S` and `.` at the end of the first line of the docstring, we try to follow PEP 257 at least somewhat to be consistent with other python projects.

alot/commands/globals.py
```diff
@@ -40,14 +40,14 @@
...
 
     def __init__(self, _prompt=True, **kwargs):
+        """
+        :param _prompt: For internal use only, used to control prompting to
+        close without sending, and is used by the BufferCloseCommand if
+        settings change after yielding to the UI.
+        """
```
dcbaker: Generally in sphinx documentation the `__init__()` method is document as part of the class documentation, and not separately. It appears that sphinx simply appends the `__init__()` documentation to the class documentation. I don't have a strong opinion either way, as long as we're consistent. @pazz, @lucc opinions?

alexshpilkin:
What about the docstring thing in Travis? I have to admit I’ve no idea what it’s complaining about.

alot/commands/search.py
```diff
@@ -135,10 +135,9 @@ def apply(self, ui):
                           'default': 'True',
                           'help': 'postpone a writeout to the index'}),
         (['tags'], {'help': 'comma separated list of tags'})],
-    help='flip presence of tags on this thread. A tag is considered present '
-         'if at least one message contained in this thread is tagged with it. '
-         'In that case this command will remove the tag from every message in '
-         'the thread.')
+    help='flip presence of tags on this thread: a tag is considered present '
+         'and removed if at least one message in this thread is tagged with '
```
dcbaker: you're missing `will be` between `and` and `removed`, or some other change to make this sentence grammatically correct.

alexshpilkin:
Ah, I see what’s wrong. Hang on.

docs/source/generate_commands.py
```diff
@@ -110,7 +110,7 @@ def get_mode_docs():
             modefile.write('The following commands are available in %s mode'
                            '\n\n' % mode)
         else:
-            header = 'Global Commands'
+            header = 'Global commands'
```
dcbaker: This is a title, please leave `commands` capitalized.

alot/commands/globals.py
alexshpilkin: Hm. This is not followed for almost all of the other commands, but OK, I’ll put the non-capitalized version (required for consistency of command descriptions) into the decorator instead.

dcbaker: Our documentation (as you noticed) is kinda all over the place.

alot/commands/search.py
alexshpilkin: I don’t believe I am, unless you insist on the future tense: “a tag (is (considered present) and (removed)) if ...”.

docs/source/generate_commands.py
alexshpilkin: None of the sibling titles are capitalized (and I see no way to change this without making the whole thing ugly):

- 2.4 Commands
    - 2.4.1 Global Commands
    - 2.4.2 Commands in _bufferlist_ mode
    - 2.4.3 Commands in _envelope_ mode
    - 2.4.4 Commands in _namedqueries_ mode
    - _&c._

Perhaps I should fix the (non-)capitalization of 2.1, though.

alot/commands/globals.py
alexshpilkin: By the way, is there any reason for the occasional difference between between the docstring and the `help=` argument of the decorator? (See `MoveCommand`, `ReloadCommand` in the same file.) If not, perhaps it could be fixed as well.

alot/commands/search.py
dcbaker: I asked a co-worker to look at this to make sure I'm not way off, but both of us parsed this as "flip present of tags on this thread: a tag is considered (present and removed) if at least one...". We both agree that the original message reads more naturally.

docs/source/generate_commands.py
dcbaker: Okay, it's annoying that the TOC renders it that way, because it does look somewhat unnatural there.

alot/commands/search.py
alexshpilkin: ”Will be removed“ it is then. I do tend to overcomplicate my sentences sometimes. BTW, the reason for changing it was to turn the whole thing into a single sentence (as it must be if we’re not capitalizing it).

docs/source/generate_commands.py
alexshpilkin: I actually feel all right without titlecase for tertiary headings. My native language doesn’t use titlecase at all, though.

What do you think should be done with the tertiary titles in section 3? Currently about a half of them are titlecased.

alexshpilkin:
Ouch. That is one confusing error message. (Solution: rebase on top of master. The check is failing on the automatic merge commit, not on my branch.)

alexshpilkin:
Changed things as discussed, but left “Global commands” non-titlecased. Added a colon in “The following commands are available ... :”. Titlecased 2.1 “Command-Line Invocation” and 3.1 “Configuration Options”.

alot/commands/globals.py
alexshpilkin: Oh. Missed this, and didn’t know the convention. The change is in any case unnecessary now, because I only moved the parameter description so that it doesn’t leak into the user documentation, but we’ve since agreed to move the user documentation out of the docstring and into the `@registerCommand` decorator.

dcbaker:
Merged. Thank you!

PR #1280, opened 2018-07-31
db/manager: Drop async method

dcbaker:
As far as I can tell using a separate process doesn't actually improve
performance, it makes it worse. The work that we're passing off to the
separate function isn't necessarily work that's well suited to being
handed off, there isn't a lot of computation and the objects that need
to be passed across the pipe are fairly large (at least when considering
a pipe). Converting the function to a generator gives better performance
and simplifies the implementation.

This is an alternate to #1274, I just wanted to get a baseline for the advantages of using a separate processing, and found that this was about as performant as using a SizedPipe, and much simpler.

### DIFF
--- a/alot/buffers/search.py
+++ b/alot/buffers/search.py
@@ -6,7 +6,7 @@ from notmuch import NotmuchError
 
 from .buffer import Buffer
 from ..settings.const import settings
-from ..walker import PipeWalker
+from ..walker import IterableWalker
 from ..widgets.search import ThreadlineWidget
 
 
@@ -26,7 +26,6 @@ class SearchBuffer(Buffer):
         self.sort_order = sort_order or default_order
         self.result_count = 0
         self.isinitialized = False
-        self.proc = None  # process that fills our pipe
         self.rebuild()
         Buffer.__init__(self, ui, self.body)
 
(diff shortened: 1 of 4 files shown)


pazz:
nice, thanks! I'm all in favour of simplifying the code of course, and if things get faster in the process all the better ;-)
I have yet to test this but to point out the obvious, the new walker class needs some documentation and the tests still need to be adjusted.. I'll be back with more detailed feedback soonish.

dcbaker:
Yeah, I wanted to put it out there before I did docs and tests in case there was something obvious I was missing.

pazz:
I guess we're going forward with this one then (after removing the unused imports)?

pazz:
where do we stand on this?
I've just come across some large threads containing HTML mails and am fairly annoyed by how long they load. This branch should help.
Further, there should be some mechanism to only call external html renderers on mails that are unfolded in the thread..

pazz:
I have implemented and merged the extra mechanism (to only render read unfolded messages) in
https://github.com/pazz/alot/pull/1398. Since I did not have problems so far with this branch, I propose we'll merge as it is (after rebasing). thoughts?

pazz:
here we go: https://github.com/pazz/alot/commit/7494812222b98888158f5dfa006ab2fcbbbe6fa9
This actually fixes  an interface issue for me where upon startup, I see some text on screen (too fast to identify) before the search mode shows.

lucc:
@pazz I open several buffers on startup and see them all flash up before the initial command moves to the inbox again. My disk is slow so sometimes I can really see the buffers. Is that the same for you but just faster?

Also what is it about the commit you mention. It looks very similar to this PR? Does it just resolve the conflicts?

pazz:
@lucc yes: that commit just rebases this branch to master and is ready to merge.
I do not load more than one buffer when I start alot, so that cannot be my issue with master.
I would have filed an issue but it is difficult to describe and reproduce.. All I'm saying here is that this particular PR solves this other funky thing for me.

lucc:
I am currently testing this PR (pazz's commit) and #1421 , so far no issues so from my point of view we can merge.

pazz:
merged as #1423

PR #1282, opened 2018-08-01
db/manager: Prepend underscore to method named async

kyrias:
"async" is not an allowed name in Python 3.7.

### DIFF
--- a/alot/db/manager.py
+++ b/alot/db/manager.py
@@ -311,7 +311,7 @@ class DBManager(object):
         db = Database(path=self.path)
         return {k[6:]: v for k, v in db.get_configs('query.')}
 
-    def async(self, cbl, fun):
+    def async_(self, cbl, fun):
         """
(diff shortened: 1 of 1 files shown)


dcbaker:
That changes the presumed scope of the symbol, since by convention python uses `_method` to mean "not public", usually one appends `_` to the end of the method or variable name to avoid these kind of issues. there's also #1280 that proposes dropping the async method entirely. I think we should get this sorted before we make the next release though, so I'll add a bug to the milestone.

kyrias:
Mmh, I can either change the name, or wait for dropping it to happen.

dcbaker:
I don't have a problem which changing it in the mean time, supporting 3.7 is useful and any merge conflicts will trivial.

kyrias:
Done.

dcbaker:
Looks good, thanks!

PR #1285, opened 2018-08-01
Travis python37

dcbaker:
Test alot on python 3.7 in CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,15 +7,15 @@ cache:
   ccache: True
 
 
-# We need a newer version of zlib to build notmuch from source as is available
-# in precise.
-dist: trusty
+# Python 3.7 is not supported in trusty
+dist: xenial
 
 python:
   # We can add more version strings here when we support other python
   # versions.
   - "3.5"
   - "3.6"
+  - "3.7"
 
 # We start two containers in parallel, one to check and build the docs and the
 # other to run the test suite.
--- a/tests/crypto_test.py
+++ b/tests/crypto_test.py
@@ -1,4 +1,5 @@
 # Copyright (C) 2017 Lucas Hoffmann
+# Copyright © 2017-2018 Dylan Baker
 # This file is released under the GNU GPL, version 3 or a later revision.
 # For further details see the COPYING file
 import os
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -7,15 +7,15 @@ cache:
   ccache: True
 
 
-# We need a newer version of zlib to build notmuch from source as is available
-# in precise.
-dist: trusty
+# Python 3.7 is not supported in trusty
+dist: xenial
```
lucc: Since when is that possible? I never saw it in the docs? (But the log claims it works.)

dcbaker: I found it in a bug report about python 3.7 support. It's apparently not fully supported yet, but it seems to work. The suggested workaround for python 3.7 support was to build a docker image, which we can do but it's much more involved than just setting "xenial"

lucc: Can we live with some older libraries than the ones we currently build by hand? Here is what we build by hand and what xenial has:
* gpgme 1.9 -> [1.6](https://packages.ubuntu.com/xenial/libgpgme11)
* libassuan (dep for gpgme) 2.4.3 -> [2.4.2](https://packages.ubuntu.com/xenial/libassuan-dev)
* libgpg-error (dep for gpgme) 1.27 -> [1.21](https://packages.ubuntu.com/xenial/libgpg-error-dev)
* notmuch (we require 0.23 in setup.py) git (xenial is to low but we could switch to a tarball)

dcbaker: I think it's worth trying to use the existing versions of stuff (I'm pretty sure that libassuan can be dropped at least). I'll push a few patches and see what happens.

.travis.yml
```diff
@@ -35,6 +35,7 @@ addons:
       - libtalloc-dev
       - zlib1g-dev
       - libgmime-2.6-dev
+      - libassuan-dev
```
lucc: Is it a dependency of notmuch or gpg?  Se the comment above.,

dcbaker: Doesn't seem happy with libgpgme 1.6: https://travis-ci.org/pazz/alot/jobs/411043268

pazz:
https://travis-ci.org/pazz/alot/jobs/411183305
is this related?

lucc:
No, after restarting it it completes: https://travis-ci.org/pazz/alot/jobs/411183305

pazz:
strange, but ok :) cheers!

pazz:
by the way: should we also mention the new python version in our setup.py?

dcbaker:
ah, yes we should. I'll make a new PR

PR #1290, opened 2018-08-03
Fix 1289

dcbaker:
I like deleting code.


pazz:
I'm happy with deeleting code in favour of using the stdlib as well. 
have you tested that signed/ encrypted mails are accepted by other MUAs?

lucc:
And they are accepted by astroid and neomutt.

Issue #1303, opened 2018-08-15
sendmail return values ignored


pazz:
I've recently updated my account list and while debugging some strange behaviour I noticed that calls to my MTA (msmtp) are unusually snappy.
It turns out that its error code and return values are completely ignored.
The offending commit seems to be 3c0097c3.

**To Reproduce**
Steps to reproduce the behaviour:
1. change your sendmail command to `sendmail_command = fail`
2. compose + send a mail
3. notice that `fail` is a MTA now :)

**Error Log**
nothing special, it looks as if the command succeed:
```
DEBUG:helper:CMD = [b'fail']
INFO:account:sent mail successfully
```

lucc:
Checking my journalctl logs (which record the activity of msmtp for me) I can see that I did not retry to send an email if it failed during the last month but did retry (similar log entry shortly after a failure) some time earlier. So I conclude that this also happened here.

PR #1304, opened 2018-08-15
Use unittest.mock from the stdlib

lucc:
It was added to the stdlib in py3.3 and we already require 3.5.

I am trying to write a test for #1303 and found this on the way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: True
 dist: xenial
 
 python:
-  - "3.5"
   - "3.6"
   - "3.7"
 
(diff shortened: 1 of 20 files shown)


lucc:
Ok I just noticed that this will require py3.6 because we use `Mock.assert_called_once()` so I think it is on hold.

pazz:
should we move towards this? afaik, this PR was postponed because we wanted to avoid a dependency on 3.5+, but in fact I'd also like to move to 3.6 for #1421

lucc:
I am on Arch Linux and have py3.7 so that's fine by me. Even Debian stable is on py3.7!

pazz:
huh? last time I checked this had conflicts? Alright then, let's merge.

lucc:
I rebased so you can review and merge.

pazz:
Thanks :)

PR #1306, opened 2018-08-16
Check the return code from sendmail

### LINKED ISSUE #1303: sendmail return values ignored

lucc:
I took the simple route, so maybe there is a better way to get the exception for the failing command than the simple assert. But this should fix #1303 which was quite annoying.

### DIFF
--- a/alot/account.py
+++ b/alot/account.py
@@ -348,7 +348,9 @@ class SendmailAccount(Account):
 
         try:
             # make sure self.mail is a string
-            out, *_ = await call_cmd_async(cmdlist, stdin=str(mail))
+            out, err, code = await call_cmd_async(cmdlist, stdin=str(mail))
+            if code != 0:
+                raise Exception
         except Exception as e:
             logging.error(str(e))
             raise SendingMailFailed(str(e))
--- a/alot/account.py
+++ b/alot/account.py
@@ -350,7 +350,9 @@ class SendmailAccount(Account):
             # make sure self.mail is a string
             out, err, code = await call_cmd_async(cmdlist, stdin=str(mail))
             if code != 0:
-                raise Exception
+                msg = 'The sendmail command {} returned with code {}{}'.format(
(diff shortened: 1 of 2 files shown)


alot/account.py
```diff
@@ -348,7 +348,9 @@ def __init__(self, cmd, **kwargs):
 
         try:
             # make sure self.mail is a string
-            out, *_ = await call_cmd_async(cmdlist, stdin=str(mail))
+            out, err, code = await call_cmd_async(cmdlist, stdin=str(mail))
+            assert code == 0, \
+                "command exited with code {}.\n{}".format(code, err)
```
dcbaker: I don't think an assert is the right way to handle this. Asserts are for cases that must be true, and not being true is an unrecoverable error, they also don't get run if you use -O. So I think we should do an `if code != 0: raise` instead.

lucc:
Thanks @dcbaker  I amended accordingly.  But I am still unhappy as I raise an exception that is caught in the next line. How can we find out if the code above will ever throw an exception?

pazz:
@lucc: I believe that the call to `call_cmd_async` may raise different exceptions as (AFAIR), binary-not-found treated differently from simply return code \neq 0. Let's keep it as is for now.
@dcbaker: I'f you're OK with the changes please amend your review and merge this in. I'd like to make a release soon and this issue is a blocker. Thanks!

lucc:
For any reviewer: I added a new commit to show the error text in the notification.

PR #1308, opened 2018-08-18
Enable zsh completion of email addresses

lucc:
Currently only the recipient addresses are completed.  It does not make
sense to add the same completion for the sender option as this should
only be used with addresses that are defined in the alot config file.

I have this code ready which we can put in a zsh function to complete the
sender account addresses.
```python
import configobj, sys
config = configobj.ConfigObj(infile=sys.argv[1], encoding="UTF8")
accounts = config.get("accounts", {})
addresses = [accounts[k].get('address') for k in accounts
	     if type(accounts[k]) is configobj.Section]
print(" ".join(addresses))
```
It can be defined inline in the zsh function like this:
```zsh
_alot_account_emails()
{
  python3 - ${XDG_CONFIG_HOME:-~/.config}/alot/config <<-'EOF'
	...
	EOF
}
```
My problem currently is that I don't know enough zsh completion logic to turn
`_alot_account_emails` into a proper completion function that could be used
like the `_email_addresses` function in this patch.

Does anyone of you know how to do it?  Otherwise we can also merge just this
one commit.

### DIFF
--- a/extra/completion/alot-completion.zsh
+++ b/extra/completion/alot-completion.zsh
@@ -24,6 +24,18 @@ _alot_search()
     '--sort=[sort results]:sorting:((newest_first\:"reverse chronological order" oldest_first\:"chronological order" message_id\:"lexicographically by Message Id"))'
 }
 
+_alot_account_emails()
+{
+  python3 - ${XDG_CONFIG_HOME:-~/.config}/alot/config <<-'EOF'
+	import configobj, sys
+	config = configobj.ConfigObj(infile=sys.argv[1], encoding="UTF8")
+	accounts = config.get("accounts", {})
+	addresses = [accounts[k].get('address') for k in accounts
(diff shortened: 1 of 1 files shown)


dcbaker:
zsh completion is mind boggling. I've read all the docs and still never managed to really get things working :(

lucc:
Ok lets give up on the sender address for now. Will merge after a review.

lucc:
I updatet the message of the first commit (contents unchanged). But after looking at my stashed code again I came up with a way to solve the sender problem. Now the addresses are simply interpolated into the argument to `_arguments`. Sorry for the delay and reiteration of the review, I should have come up with this the first time :(

pazz:
typo "picket" --> "picked" in the 2nd commit msg.
I haven't tried it out but looks ok (and harmless enough to push) to me.

lucc:
Fixed typo.

Issue #1044, opened 2017-04-13
confusing command for editing draft


dcbaker:
the use of `editnew` to edit a draft message is extremely confusing, and not at all obvious. I towuld make much more sense to have `editnew` always return a new message, ala `compose`.

I think we should drop `editnew` alltogether, and just have `reply`, `groupreply`, `forward`, `compose`, and `edit` or `editdraft`

pazz:
I agree that the whole drafts handling is not really intuitive.
The thing is that drafts are nothing else but messages usually stored in
a special folder and tagged with "draft". So from notmuch's point of
view these are messages like all others.
Do you have a cleaner solution?

dcbaker:
I think what's really confusing about them is that `editnew` intuitively **doesn't** open a draft to edit further, it creates a new message. We already have `compose` for writing a brand new email, so maybe just renaming `editnew` to `editexisting` or `editdraft`? Or does `editnew` create a new email if there isn't a draft to edit?

pazz:
Quoting Dylan Baker (2017-04-14 17:48:00)
> Does editnew create a new email if there isn't a draft to edit?

No: it works on all mails, as drafts are just specially tagged mails.
I always read this command as "edit_as_new", which is exactly what it
does. It creates a new mail with this one as template.
Maybe there should be an option (default?) to remove the template after
sending the new one.

I just want to avoid giving the impression that this can be used to edit
mails on disk. mails which were sent/received should be considered
immutable.

dcbaker:
ah, the `edit_as_new` makes more sense now, that's just not how I or my co-worker interpreted the command. I'll think about the draft handling a little more. I think we can do better, but I'm not sure exactly how.

fnurl:
Hi, was trying to find out how to continue writing a message saved as a draft. The discussion in this thread has led me to understanding that in alot:

- when composing a message you can save it in a folder configured as the "drafts folder"
- there are no drafts in alot, just a folder where unsent messages can be saved without sending them
- what "save draft" actually means is "store this message in this specific maildir folder without sending it"
- users cannot edit existing messages in the maildir
- this means that users cannot "continue" editing a draft, because 1) messages cannot be edited and 2) there are not special "draft type" messages that are excepted from rule 1.
- the way to "continue editing" a draft is to create a new message using the previous message state as a template then either save the new message as a draft or send it, then delete the previous state(s) of the message

Is this correct?

dcbaker:
[...]

PR #1313, opened 2018-09-14
Option to delete old drafts when sending or saving.

stevenengler:
When you run the 'editnew' command in a thread view, if the email you are copying has the 'draft' flag, it will pass this message on to the envelope object. When the user saves or sends a draft, they will be prompted to delete the old draft.

I do not currently have alot set up to send emails, so someone would need to test that process. But I believe it should work, it's not really any different from the process when saving new drafts.

Let me know if you'd like me to change anything.

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -133,23 +133,59 @@ class SaveCommand(Command):
         path = account.store_draft_mail(
             mail.as_string(policy=email.policy.SMTP))
 
-        msg = 'draft saved successfully'
-
         # add mail to index if maildir path available
         if path is not None:
-            ui.notify(msg + ' to %s' % path)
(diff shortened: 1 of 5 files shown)


stevenengler:
Okay, squashed and built.

alot/commands/envelope.py
```diff
@@ -280,6 +316,47 @@ def __init__(self, mail=None, envelope=None, **kwargs):
                 ui.dbman.add_message(path, account.sent_tags + initial_tags)
                 await ui.apply_command(globals.FlushCommand())
 
+            # if provided, delete the old draft
```
pazz: This looks very much like the code in `SendCommand` above. why the duplication?

polyzen:
Referencing #1044

Issue #1299, opened 2018-08-12
Delivery Status Notification


varac:
I'd like to be able to request an optional positive Delivery Status Notification for certain emails, meaning asking the recipients mail server to send back a message on successful delivery.

Other MUAs have this implemented, i.e Thunderbird, and it is a useful feature.

pazz:
isn't that just some header value that one can already manually add?

varac:
I haven't looked into it in detail. Even if so it would be nice to have an easy way of triggering this from alot, but if it's really just a header then this should be easy to do in a custom key binding.

f-thiele:
I have written a solution for this feature It uses the referenced pull request so I will wait to open a new PR until that one is settled but it can be seen in my fork [here](https://github.com/f-thiele/alot/compare/toggleheader...f-thiele:request_dns_success).

It creates two new commands: `receipt-read` and `receipt-delivered` that take an optional argument. If no argument is supplied and an account for the current `From` defined, it will simply take that address as return address. Otherwise it uses the address given by the user.

---

Implementation can be seen in: https://github.com/f-thiele/alot/commit/5e450d5fbf8babf333834f880ec7d73106ea0f54
Tests can be seen in: https://github.com/f-thiele/alot/commit/41e99f759ca17bccf5c26b2db59feb9bcf5ee55b
Dependent pull request: https://github.com/pazz/alot/pull/1315

PR #1315, opened 2018-09-21
Implement `toggleset` to allow easy set/unset of header

f-thiele:
There is currently no toggletags equivalent for a header field.
To easily switch certain predefined header fields in the config
file on or off it is useful to have the option of binding
```
toggleset key value
```
to a certain key, making it for the user easy to switch between
different scenarios in which specific headers are wanted or
unwanted.

Suggestions for implementation or naming are of course very
welcome.

This addresses a small thing I noticed in an earlier comment of
mine: https://github.com/pazz/alot/issues/1299#issuecomment-423550190.

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -428,26 +428,19 @@ class SetCommand(Command):
         envelope = ui.current_buffer.envelope
 
         if self.action == 'set':
-            do_set = True
+            # add a new header
+            envelope.add(self.key, self.value, self.reset)
         elif self.action == 'unset':
-            do_set = False
-        elif self.action == 'toggleset':
-            # in case we toggle we only `set` a new header if it didn't exist
-            # already before, in case it exists we will simply remove it hence
-            # `do_set = False`
-            do_set = False if self.key in envelope else True
-
-        if self.reset or not do_set:
-            # remove the existing header if
-            #  - we don't append
-            #  - we are in a toggle scenario and it exists already
-            #  - we are in the unset scenario
-
+            # we want to unset the header if it already exists
             if self.key in envelope:
                 del envelope[self.key]
-
-        if do_set:
-            envelope.add(self.key, self.value)
+        elif self.action == 'toggleset':
+            # in case we toggle we only add a new header if it didn't exist
+            # already before, in case it exists we will simply remove it
+            if self.key in envelope:
+                del envelope[self.key]
+            else:
+                envelope.add(self.key, self.value)
 
         # FIXME: handle BCC as well
         # Currently we don't handle bcc because it creates a side channel leak,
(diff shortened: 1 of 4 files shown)


f-thiele:
Following the example of sign/unsign/togglesign the handling of
headers using the set and unset commands has now been unified.
This makes it easier to maintain and avoids the special treatment
of FIXME in two (three with the toggleheader command) functions.

I also renamed `toggleheader` to `toggleset` to be more in line
with `togglesign` and `toggletags`.

Additionally tests are now supplied for the envelope `SetCommand`.
Previous tests of this seemed to not exist.

The dedicated tests are written for a variety of test cases that
cover

  - setting of header
  - unsetting of header
  - toggling header

All of them are tested in a variety of scenarios where unneeded
`append` flag is supplied or values that are not required. Here
we follow and assume the hierarchy of evaluation as below:

 - if header exists and should be toggled / unset we do not care
   whether we also obtain value or append
 - if header does not exist we add whatever value we get, even if
   append is set

As `From`, `Cc`, `To` have special treatment in `SetCommand` we
include as well for the `From` header three tests to be explicit
that this special treatment doesn't break the header handling.

Tests could be slimlined in terms of lines, but are tried to be
as close to other test cases in this file as possible.

alot/commands/envelope.py
```diff
@@ -392,93 +392,72 @@ def openEnvelopeFromTmpfile():
...
+        # a value is only needed if we set or toggle else it is None
+        self.value = ' '.join(value) if value is not None else None
+
         self.reset = not append
         Command.__init__(self, **kwargs)
 
     async def apply(self, ui):
         envelope = ui.current_buffer.envelope
-        if self.reset:
+
```
pazz: I'd prefer to do it without the extra `do_set` variable and sinmply make changes to the envelope within the respective clauses to keep things simple.

lucc:
(I have not read the code yet, just the comments so I hope the following makes sense.)

@pazz `__setitem__` is a special python method that expects three arguments (`self, index, value`). Does it really make sense to add keyword arguments? The extra arg will not be usable with the syntactic sugar `object[index]`.

Maybe it makes more sense to make the `add` method the real worker with the keyword argument and have the `__setitem__` just wrap `add`.

pazz:
@lucc: sounds OK to me, I just wanted to eliminate my silly code duplication.

alot/commands/envelope.py
f-thiele: This is fixed in https://github.com/pazz/alot/pull/1315/commits/ca0bef0067646f94cc2f7b18391cc6f23e128f73

f-thiele:
Thanks @pazz for reviewing. I have implemented the requested changes and ended up using @lucc's suggestion to make `envelope.__setitem__` wrap `envelope.add`.

I'm not quite sure what you would like me to update about the tests though, so they remain untouched.

Issue #1233, opened 2018-04-02
Support gpg header encrytion aka memoryhole (feature request)


varac:
It would be great if alot implements support for header encryption. The [memoryhole project](https://github.com/autocrypt/memoryhole) worked on a specification for that and different mail clients started implementing it (enigmail 2.0 will have the most adoption for now).

From the memoryhole README.md:
> The Memory Hole project aims to fix this problem by stashing relevant headers within the body of the e-mail in a standardized way. Compatible mail user agents should be able to use this structure to display the cryptographically protected headers to the user, indicating their protection.

Since enigmail 2.0 is able to encrypt certain mail headers like subject i.e. if GPG-encryption is used, more and more mails show up in my mailbox with the subject "Encrypted message".
That's great, people adopt it - but it breaks thread recogniztion really bad.

There's already a motion to include support for encrypted headers in notmuch - but this would need enabling gpg decryption while indexing mail - a feature not everybody is comfortable with.

Please implement this feature in alot!

meskio:
The latest release of notmuch includes support for protected headers: https://notmuchmail.org/news/release-0.29/

Right now using this feature in the search window you get the right header, as it comes from notmuch. But the thread window reads the email from the file (instead of from notmuch) and you see the `bad` subject.

As it was discussed in #1239 it will make sense to explore how to use notmuch to get the email content, this will not only provide us with the right headers it will as well simplify many parts of alot removing the need of parsing email in alot. To be able to do that we'll need to refactor notmuch to have the right pieces in libnotmuch. 

In the mean time it makes sense to implement support for parsing and producing encrypted headers in alot. #1316 is an start for that.

PR #1316, opened 2018-09-27
Show decrypted subject in search results.

drebs:
Memory hole is currently implemented in Enigmail and message headers can
be protected in the encrypted payload of a message. This commit modifies
the update of a thread's subject by extracting the subject from the
encrypted headers when the most recent message contains protected
headers (v1). This results in the decrypted header being shown in the
search results.

See:

  https://github.com/autocrypt/memoryhole
  https://wiki.gnupg.org/OpenPGPEmailSummit201607/MemoryHole
  https://www.enigmail.net/index.php/en/user-manual/advanced-operations

This is part of #1233. Another part is missing to show decrypted headers
in the message tree.

### DIFF
--- a/alot/db/thread.py
+++ b/alot/db/thread.py
@@ -49,6 +49,22 @@ class Thread(object):
                 subject = ''
         self._subject = subject
 
+        # maybe decrypt subject in protected headers
+        # (see: https://github.com/autocrypt/memoryhole)
+        msgs = [m for m in thread.get_messages()]
+        last = msgs[-1]
+        first_part = last.get_message_parts()[0]
+        if first_part.get_content_type() == 'application/pgp-encrypted':
+            second_part = last.get_message_parts()[1]
+            encrypted = second_part.get_payload()
+            from email.parser import Parser
+            from alot.crypto import decrypt_verify
+            _, plaintext = decrypt_verify(encrypted)
+            parser = Parser()
+            decrypted = parser.parsestr(plaintext)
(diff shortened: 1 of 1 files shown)


pazz:
Is alot the right place to implement this? How much is/will be implemented at the notmuch level?

drebs:
> Is alot the right place to implement this? How much is/will be implemented at the notmuch level?

Good question, here's a message in notmuch's mailing list regarding this: https://notmuchmail.org/pipermail/notmuch/2018/026405.html

I will write to the person and get back here with more info.

dcbaker:
I think @dkg was the one working on this?

meskio:
I think what it makes sense will be to us notmuch to decrypt messages (and handle memoryhole). But for now this is not possible, as the decryption (I will assume the future memoryhole) lives in the notmuch cli and not in the library. It will be nice to move it into the library, and I think there is people in the notmuch core that will support it, but is not a trivial work.

More info in a previous conversation on the topic: #1239

meskio:
Now that there is proper support in notmuch for protected headers the search results contain the right headers. I thinks this pull-req can be closed.

The missing piece is to have the right header in the thread view. @drebs do you feel like giving it a try? If you don't have time I should be able to look into it next week.

dkg:
@drebs, if you want to find some time to poke around at this together,
please let me know.  Finding me on IRC might be the best way to coordinate.

PR #1319, opened 2018-10-04
Doc

psic:
+ Add a 9th section in the FAQ about using :pipeto less
+ Add a can do / can't do with alot in the description section

### DIFF
--- a/docs/source/description.rst
+++ b/docs/source/description.rst
@@ -2,17 +2,19 @@ Alot is a terminal-based mail user agent for the notmuch mail system.
 It features a modular and command prompt driven interface
 to provide a full MUA experience as an alternative to the Emacs mode shipped
 with notmuch.
-What you can do whit alot:
 
-	* Read mail/thread
+What you can do with alot:
+
+	* Read mail
 	* Search mail
-	* Tag/Untag mail
+	* Tag, untag mail
 	* Filter mail	
 	* Save attachment
-	* Receive mail (using external tool as `mbsync <http://isync.sourceforge.net/mbsync.html>`_ or `offlineIMAP <http://www.offlineimap.org/>`_)
-	* Send/Reply to/Bounce/Forward mail (using external tool as `msmtp <https://marlam.de/msmtp/>`_)
+	* Send, reply to, bounce and forward mail (using external tool as `msmtp <https://marlam.de/msmtp/>`_)
 	* View attachment (using externals tools)
 
 What you can't do with alot:
 	
-	* Search in/save mail
+	* Recieve mail. Alot uses notmuch mail system
+	* Search in mail content. Use the :pipeto command to a pager
+	* Save mail content. Use the :pipeto command to a pager
--- a/docs/source/description.rst
+++ b/docs/source/description.rst
@@ -2,3 +2,17 @@ Alot is a terminal-based mail user agent for the notmuch mail system.
 It features a modular and command prompt driven interface
 to provide a full MUA experience as an alternative to the Emacs mode shipped
 with notmuch.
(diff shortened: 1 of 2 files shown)


docs/source/description.rst
```diff
@@ -2,3 +2,17 @@ Alot is a terminal-based mail user agent for the notmuch mail system.
 It features a modular and command prompt driven interface
 to provide a full MUA experience as an alternative to the Emacs mode shipped
 with notmuch.
+What you can do whit alot:
```
dcbaker: whit -> with

docs/source/description.rst
```diff
@@ -2,3 +2,17 @@ Alot is a terminal-based mail user agent for the notmuch mail system.
...
+
+	* Read mail/thread
```
dcbaker: I don't think the / makes this read nicely. Maybe just `Read mail`

docs/source/description.rst
```diff
@@ -2,3 +2,17 @@ Alot is a terminal-based mail user agent for the notmuch mail system.
...
+	* Search mail
+	* Tag/Untag mail
+	* Filter mail	
+	* Save attachment
```
dcbaker: probably should be `Save attachments`

docs/source/description.rst
```diff
@@ -2,3 +2,17 @@ Alot is a terminal-based mail user agent for the notmuch mail system.
...
+	* Receive mail (using external tool as `mbsync <http://isync.sourceforge.net/mbsync.html>`_ or `offlineIMAP <http://www.offlineimap.org/>`_)
+	* Send/Reply to/Bounce/Forward mail (using external tool as `msmtp <https://marlam.de/msmtp/>`_)
```
dcbaker: How about using commas instead of slashes `Send, reply to, bounce, and forward mail`

docs/source/description.rst
```diff
@@ -2,3 +2,17 @@ Alot is a terminal-based mail user agent for the notmuch mail system.
...
+	* View attachment (using externals tools)
+
+What you can't do with alot:
+	
+	* Search in/save mail
```
dcbaker: These are two different things in a single line, also have you opened an issue for the "save mail" thing? I shouldn't be that hard to do. You could probably use `:pipeto` with `cat` to make that work, and it would be trivial to implement in alot itself.

docs/source/description.rst
```diff
@@ -2,3 +2,17 @@ Alot is a terminal-based mail user agent for the notmuch mail system.
...
 to provide a full MUA experience as an alternative to the Emacs mode shipped
 with notmuch.
+What you can do whit alot:
+
+	* Read mail/thread
+	* Search mail
+	* Tag/Untag mail
+	* Filter mail	
+	* Save attachment
+	* Receive mail (using external tool as `mbsync <http://isync.sourceforge.net/mbsync.html>`_ or `offlineIMAP <http://www.offlineimap.org/>`_)
```
lucc: I am confused: Is there anything we support especially for receiving mail from withing alot? Because if it just the ability to run a shell script this is not a feature specific to receiving mail.

docs/source/description.rst
psic: ok

docs/source/description.rst
psic: why attachments would be plural and mail would be singular ?

docs/source/description.rst
psic: ok

pazz:
I have added a [new FAQ item mentioning `pipeto`](cmd.thread.pipeto) and discarded the "feature list" part of this PR.

PR #1320, opened 2018-10-04
Fix async idle

dcbaker:
It turns out that urwid just doesn't work very will with asyncio. I simply don't have time or motivation to try to fix urwid, so instead I'm proposing trying to get the nice syntax of asyncio, but without the overhead by continuing to use urwid's `TwistedEventLoop`. Since we've installed the twisted reactor as the asyncio loop, starting Twisted's loop starts processing asyncio calls.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -119,7 +119,7 @@ install:
   - |
     set -e
     if [[ $JOB = docs ]]; then
-      pip install configobj python-magic urwidtrees
+      pip install configobj twisted python-magic urwidtrees
       # Mock all "difficult" dependencies of alot in order to be able to import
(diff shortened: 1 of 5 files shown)


dcbaker:
stupid codeclimate, it's mad that an inline comment goes slightly past 79 characters.

pazz:
thanks for taking the time @dcbaker !
Did you poke the urwid devs about this as well?
Should we somewhere document, or at least add a note-to-self, that we'd like to eventually switch to urwids asyncio eventloop?

dcbaker:
I think it would be nice to use the asyncio loop, but that would require urwid fixes, so I don't know if that will happen.

pazz:
understood, thx!

Issue #1268, opened 2018-07-11
Introduce corpus of test messages


pazz:
It should be great to have the infrastructure to write unit tests using a complete notmuch environment, including messages and all. This was mentioned eg. already in [here](https://github.com/pazz/alot/issues/1253#issuecomment-399636806), [here](https://github.com/pazz/alot/issues/463#issuecomment-373945377) etc..

Until we do, I have started collecting half-written tests in the [wiki here](https://github.com/pazz/alot/wiki/Future-Tests). Feel free to add stuff there as a reminder to properly write a test later.

PR #1321, opened 2018-10-10
Add a mail corpus for the test suite

lucc:
This was discussed before (#1268) and there is a wiki entry for this [here](https://github.com/pazz/alot/wiki/Future-Tests). **Currently this is just a WIP branch to get the discussion going.**

I was looking at the open issues for the [0.8 milestone](https://github.com/pazz/alot/milestone/13) and both remaining issues are about mail decoding which looks like a topic best covered by a corpus of some example mails. So I took the liberty to start this.

Questions: Should we keep a notmuch index in git?
1. Pro: maybe faster when running the tests as it doesn't have to be created on the fly
2. Pro: Less setup code for a test that needs it.
3. Con:  Not all tests that need a mail from the corpus might need the index
4. Con: Maybe it will create problems if the tests are run with a different version than the version with which the index was created.
5. Con: Binary files in git. (but it is rather small when `notmuch compact` is called)

### DIFF
--- a/tests/db/utils_test.py
+++ b/tests/db/utils_test.py
@@ -741,3 +741,13 @@ class TestRemoveCte(unittest.TestCase):
         # This should not raise an UnicodeDecodeError.
         utils.remove_cte(mail, as_string=True)
         self.assertTrue(True)
+
+    @unittest.expectedFailure
+    def test_issue_1301(self):
+        fp = open('tests/static/mail/latin-1.eml')
+        mail = email.message_from_file(fp)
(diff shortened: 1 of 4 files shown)


pazz:
Thanks for this, it is a good start!
I believe that calling `notmuch new` is fast enough for such small corpora. and we should avoid the potential issues introduced by a versioning mismatch at all cost.

pazz:
I have rebased, squashed and amended this PR and re-submitted as #1342 and suggest that we close this issue.

lucc:
goto #1342

PR #1322, opened 2018-10-11
Clean up test output a little bit

lucc:


pazz:
are asserts really the thing to use here?

lucc:
That was the easiest way to hide the messages. Is there another one? Ideally we would like to silence just the logger from a certain file (the file under test).  But we don't use different loggers for different files so our messages are handled by the same logger as the unittest messages from ec215f9.

One can use a pattern like this in the code 
```py
import logging
logger = logging.getLoger(__name__)
logger.debug("foo")
```
and then we could possibly silence just this logger in the tests. But that is a bigger (global) change to the code base. Should I do that?

Issue #1232, opened 2018-03-23
Warn if mail is not encrypted to all recipients


varac:
After moving to alot (yay!) I accidentially sent several mails which were not encrypted to all recipients. This is annoying, and alot should warn the user before sending the mail.
Thunderbird/Enigmail does this job quite well.

pazz:
This sounds like a job for a small `pre_envelope_send` hook similar to the [one reminding of you of forgotten attachments](https://github.com/pazz/alot/wiki/Contrib-Hooks#check-for-missing-attachment-before-sending).

dcbaker:
I think this is worth having in core alot, if you send an encrypted message but do not have keys for all recipients you're sending them garbage.

drebs:
Here's a first draft of the functionality:

```python
from twisted.internet import defer

#
# Warn if message is not encrypted to all recipients
#
@defer.inlineCallbacks
def pre_envelope_send(ui, dbm, cmd):
    e = ui.current_buffer.envelope
    if not e.encrypt:
        # do nothing, message is not encrypted (why not!?)
        return

    # get all recipient addresses
    recipients = set(map(lambda s: parseaddr(s)[1], e.headers.get('To', [])))
    recipients.union(map(lambda s: parseaddr(s)[1], e.headers.get('Cc', [])))
    recipients.union(map(lambda s: parseaddr(s)[1], e.headers.get('Bcc', [])))
    if not recipients:
        return

    # get all email addresses of encryption keys
    keys = set()
    for key in e.encrypt_keys.values():
        for uid in key.uids:
            keys.add(uid.email)

    # fail if both sets are not equal
    if not recipients.issubset(keys):
        msg = 'Message is not encrypted to all recipients. Send anyway?'
        if not (yield ui.choice(msg, select='yes')) == 'yes':
            raise Exception()
```

Some things that come to mind:

* Is that the correct way to handle addresses in the `Bcc` header? Would the encryption keys to Bcc'ed addresses also appear in the envelope's `encrypt_keys` attribute?
* Should we handle sets comparison differently? I always encrypt to my own key, even though I am not a recipient of the message. Are there other use cases?
* Should we allow for refreshing the set of keys? If so, how?

Once we get to a nice state I can move the code to the contrib hooks page.

varac:
@drebs Do you like to submit a pull req based on your patch ?

pazz:
[...]

PR #1323, opened 2018-10-11
Check set of encryption keys before sending

drebs:
Closes: #1232

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -168,6 +168,23 @@ class SendCommand(Command):
         self.envelope = envelope
         self.envelope_buffer = None
 
+    def _get_keys_addresses(self):
+        addresses = set()
+        for key in self.envelope.encrypt_keys.values():
+            for uid in key.uids:
+                addresses.add(uid.email)
+        return addresses
+
+    def _get_recipients_addresses(self):
+        tos = self.envelope.headers.get('To', [])
+        ccs = self.envelope.headers.get('Cc', [])
+        return {a for (_, a) in email.utils.getaddresses(tos + ccs)}
+
+    def _is_encrypted_to_all_recipients(self):
+        recipients_addresses = self._get_recipients_addresses()
+        keys_addresses = self._get_keys_addresses()
+        return recipients_addresses.issubset(keys_addresses)
+
     async def apply(self, ui):
         if self.mail is None:
             if self.envelope is None:
@@ -201,6 +218,18 @@ class SendCommand(Command):
                 warning = textwrap.dedent("""\
                     Any BCC recipients will not be able to decrypt this
                     message. Do you want to send anyway?""").replace('\n', ' ')
+                if (await ui.choice(warning, cancel='no',
+                                    msg_position='left')) == 'no':
+                    return
+
+            # Check if an encrypted message is indeed encrypted to all its
+            # recipients.
+            if (self.envelope.encrypt
+                    and not self._is_encrypted_to_all_recipients()):
+                warning = textwrap.dedent("""\
+                    Message is not encrypted to all recipients. This means that
+                    not everyone will be able to decode and read this message.
+                    Do you want to send anyway?""").replace('\n', ' ')
                 if (await ui.choice(warning, cancel='no',
                                     msg_position='left')) == 'no':
                     return


alot/commands/envelope.py
```diff
@@ -168,6 +168,31 @@ def __init__(self, mail=None, envelope=None, **kwargs):
...
+        for header in ['To', 'Cc', 'Bcc']:
+            values = self.envelope.headers.get(header)
+            if not values:
+                continue
+            for value in values:
+                for recipient in value.split(','):
+                    address = email.utils.parseaddr(recipient)[1]
+                    recipients.add(address)
+
+        if recipients:
```
meskio: I find easier to read the pattern:
```
if not recipients:
    return True

# get all email addresses of encryption keys
keys = set()
...
```

alot/commands/envelope.py
```diff
@@ -168,6 +168,33 @@ def __init__(self, mail=None, envelope=None, **kwargs):
         self.envelope = envelope
         self.envelope_buffer = None
 
+    def _check_recipient_keys(self):
+        # get all recipient addresses
+        recipients = set()
+        for header in ['To', 'Cc', 'Bcc']:
```
pazz: in other places, we chose not to include BCCs when encrypting. Let's be consistent.

alot/commands/envelope.py
```diff
@@ -168,6 +168,33 @@ def __init__(self, mail=None, envelope=None, **kwargs):
...
+            values = self.envelope.headers.get(header)
+            if not values:
+                continue
+            for value in values:
```
pazz: why not simply `for value in self.envelope.headers.get(header)` ? (I think you don't need the two lines above).

alot/commands/envelope.py
```diff
@@ -168,6 +168,33 @@ def __init__(self, mail=None, envelope=None, **kwargs):
...
+        if not recipients:
+            return True
+
+        # get all email addresses of encryption keys
+        keys = set()
+        for key in self.envelope.encrypt_keys.values():
+            for uid in key.uids:
+                keys.add(uid.email)
+
+        # fail if both sets are not equal
```
pazz: How about computing the key-addresses in a separate helper to simplify the cognitive complexity?

pazz: you are not checking for equality here.

alot/commands/envelope.py
```diff
@@ -168,6 +168,33 @@ def __init__(self, mail=None, envelope=None, **kwargs):
         self.envelope = envelope
         self.envelope_buffer = None
 
+    def _check_recipient_keys(self):
```
pazz: Please rename to something more descriptive

alot/commands/envelope.py
```diff
@@ -205,6 +232,16 @@ def __init__(self, mail=None, envelope=None, **kwargs):
                                     msg_position='left')) == 'no':
                     return
 
+            # Check if an encrypted message is indeed encrypted to all
+            # recipients.
+            if self.envelope.encrypt and not self._check_recipient_keys():
+                warning = textwrap.dedent("""\
+                    Message is not encrypted to all recipients. Do you want to
```
pazz: can we extend this message to include the consequences?
"This means that not everyone will be able to decode and read this message"
or "This means that the message will be sent out in plain text as well". I can't remember the semantics.

drebs:
I have further refactored a bit, pleas let me know if there's something else to change or improve.

alot/commands/envelope.py
```diff
@@ -168,6 +168,30 @@ def __init__(self, mail=None, envelope=None, **kwargs):
...
+                    if recipient:  # may be an empty string
+                        _, address = email.utils.parseaddr(recipient)
+                        addresses.add(address)
+        return addresses
+
+    def _is_encrypted_to_all_recipients(self):
+        recipients = self._get_recipients_addresses()
+        keys = self._get_keys_addresses()
+        if not recipients.issubset(keys):
+            return False
```
pazz: `return recipients.issubset(keys)` ?
Also, perhaps better call `keys` `addresses_fom_keys` or similar to not cunfuse the value with actual gpg-keys..

lucc:
In python3 comprehensions are generally preferred over map. And there also is a set comprehension IIRC. So you could try
```py
return { t[1] for t in email.utils.getaddresses(tos + ccs) }
```

pazz:
instead of the map-lambda construct you could simply use list comprehensions:

```
[address for (name, address) in email.utils.getaddresses(tos + ccs)]
```

drebs:
I have now mixed @pazz's list comprehension with @lucc's set comprehension to propose a cleaner way to go.

Should we also care about the other helper `_get_keys_addresses` or that one is fine as it is?

pazz:
I would say no: in `_get_keys_addresses` you have a nested `for`-loop, so using comprehensions would only make things unreadable. I'm happy with the code as is.

alot/commands/envelope.py
```diff
@@ -205,6 +222,18 @@ def __init__(self, mail=None, envelope=None, **kwargs):
                                     msg_position='left')) == 'no':
                     return
 
+            # Check if an encrypted message is indeed encrypted to all its
+            # recipients.
+            if self.envelope.encrypt \
+                    and not self._is_encrypted_to_all_recipients():
```
dcbaker: I'd like to see parens instead of \ here:
```python
if (self.envelope.encrypt
        and not self._is_encrypted_to_all_recipients()):
```

This is in keeping with the style of the ```if (await...``` below

varac:
@drebs: Thx alot for you PR, it's great!
I only noticed one minor issue, that is that if you add a comma at the end of the recipient (i.e. `To: drebs@blah.org,`) this check will complain. Alot itself handles it gracefully and would send mails regardless of trailing commas.

drebs:
@varac, I could not immediately reproduce your issue, could you please open a new issue with steps to reproduce? Thanks alot!

PR #1327, opened 2018-10-19
utils: replace unreadable characters when decode payload

meskio:
If the encoding doesn't correspond with some characters on the payload,
instead of crashing with a 'UnicodeDecodeError', let's replace the
invalid characters so we the email is displayed.


pazz:
superceeded by #1344

PR #1329, opened 2018-10-26
crypto: deal with bad signatures on session-decrypt

meskio:
When the key is known or just the signature is invalid gpgme raises
'BadSignatures'. Up to now we were not acknoledging it when using
session-keys, making impossible to open emails without the signing key.

### DIFF
--- a/alot/crypto.py
+++ b/alot/crypto.py
@@ -194,7 +194,7 @@ def verify_detached(message, signature):
     :param bytes signature: the OpenPGP signature to verify
     :returns: a list of signatures
     :rtype: list[gpg.results.Signature]
-    :raises: :class:`~alot.errors.GPGProblem` if the verification fails
+    :raises alot.errors.GPGProblem: if the verification fails
     """
     ctx = gpg.core.Context()
     try:
@@ -214,7 +214,7 @@ def decrypt_verify(encrypted, session_keys=None):
     :param list[str] session_keys: a list OpenPGP session keys
     :returns: the signatures and decrypted plaintext data
     :rtype: tuple[list[gpg.resuit.Signature], str]
-    :raises: :class:`~alot.errors.GPGProblem` if the decryption fails
+    :raises alot.errors.GPGProblem: if the decryption fails
     """
     if session_keys is not None:
         try:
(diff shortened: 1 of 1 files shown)


alot/crypto.py
```diff
@@ -223,16 +223,7 @@ def decrypt_verify(encrypted, session_keys=None):
...
-        plaintext, _, verify_result = ctx.decrypt(encrypted, verify=True)
-        sigs = verify_result.signatures
-    except gpg.errors.GPGMEError as e:
-        raise GPGProblem(str(e), code=e.getcode())
-    except gpg.errors.BadSignatures as e:
-        plaintext, _, _ = ctx.decrypt(encrypted, verify=False)
-        sigs = e.result.signatures
-
-    return sigs, plaintext
+    return _decryp_verify_with_context(ctx, encrypted)
```
dcbaker: shouldn't that be `_decrypt_verify_with_context`?

alot/crypto.py
```diff
@@ -249,14 +240,34 @@ def _decrypt_verify_session_keys(encrypted, session_keys):
...
 
+def _decryp_verify_with_context(ctx, encrypted):
+    """Decrypts the given ciphertext string using the gpg context
+    and returns both the signatures (if any) and the plaintext.
+
+    :param gpg.Context ctx: the gpg context
+    :param bytes encrypted: the mail to decrypt
+    :returns: the signatures and decrypted plaintext data
+    :rtype: tuple[list[gpg.resuit.Signature], str]
+    :raises: :class:`~alot.errors.GPGProblem` if the decryption fails
```
dcbaker: sphinx `raises` takes the type as an argument:
`:raises alot.errors.GPGProblem: if the decryption fails`

alot/crypto.py
meskio: I agree, it was a typo.

meskio:
@dcbaker thanks for the review, I think I fixed your comments.

dcbaker:
looks good to me, unfortunately I can't merge this because codeclimate has decided that 250 (rather than the normally accepted 1000) lines is too long. @pazz

pazz:
@dcbaker: approved manually. Btw, the file length at which codeclimate complains is configurable in the codeclimate rc in our repo. check out https://docs.codeclimate.com/docs/advanced-configuration#section-default-checks

PR #1330, opened 2018-11-05
db/message: accept attachments of type pgp-encrypted

meskio:
Some email clients (like thunderbird) if you attach a .gpg file it will
say the attachment content type is 'application/pgp-encrypted'. We need
to discard only the attachments with content type
'application/pgp-encrypted' if they are followed by an
'application/octet-stream', beacuse this case is a PGP/MIME message:
https://tools.ietf.org/html/rfc3156#section-6.1

### DIFF
--- a/alot/db/message.py
+++ b/alot/db/message.py
@@ -252,10 +252,13 @@ class Message(object):
                 if ct in ['octet/stream', 'application/octet-stream']:
                     content = part.get_payload(decode=True)
                     ct = helper.guess_mimetype(content)
+                    if (self._attachments and
+                            self._attachments[-1].get_content_type() ==
+                            'application/pgp-encrypted'):
+                        self._attachments.pop()
 
                 if cd.lower().startswith('attachment'):
-                    if ct.lower() not in ['application/pgp-encrypted',
-                                          'application/pgp-signature']:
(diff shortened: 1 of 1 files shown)


meskio:
I see, I need to find a better commit message.

The situation is:
1. I attach a `.gpg` file to a non-encrypted email with thunderbird.
2. Thunderbird marks this attachment as content type `application/pgp-encrypted`.
3. Alot discards any mime part with content type `application/pgp-encrypted`. As this is the content type used by PGP/MIME to mark the version (the actual encrypted material is of type `application/octet-stream`).
4. The user interface doesn't display the existence of the attachment, neither displays the content or allows to save it on any way.

The solution requires to be able to differenciate two cases:
* Normal PGP/MIME encrypted messages, where we need to discard the `application/pgp-encrypted` mime part.
* An attachment with `aplication/pgp-encrypted` as content type.

The way I did that was to discard the part only if the following one is of type `application/octet-stream`.

Is it more clear now?

meskio:
Updated the commit message. Is it more readable now?

meskio:
It looks like this pull-req is marked as accepted. What is missing to be merged? Can I do anything to help here?

PR #1332, opened 2018-11-14
make bounce command use the correct sender account

### LINKED ISSUE #1331: Bounce command uses the wrong sending account
SendCommand() determines the sending account from the "From:" header, which is wrong in the case of a bounce. (In most cases this will result in using the default account.)

SendCommand() needs to use the "Resent-From:" header if present. PR upcoming.

mjg:
Bounce correctly determines the address and account to send the bounce
from. It uses the account to choose an address book for "to:" completion
and passes the sender address as "Resent-From:" to SendCommand(). The
latter uses the "From:" header, though, to determine the sending account
again and (in the case of a bounce) wrongly.

Make SendCommand() use "Resent-From:" if present and "From:" else.

Code remark: specifying "False" as the default return value for get()
ist not necessary; it is meant to make this short form clearer to read,
and to safeguard for times when '' or None do not evaluate to False
here.

Fixes #1331 :)

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -253,7 +253,7 @@ class SendCommand(Command):
         if not isinstance(msg, email.message.Message):
             msg = email.message_from_string(
                 self.mail, policy=email.policy.SMTP)
-        address = msg.get('From', '')
+        address = msg.get('Resent-From', False) or msg.get('From', '')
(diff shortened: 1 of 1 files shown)


mjg:
Test suite runs fine locally. Travis is missing some default config, it seems.

pazz:
thanks

PR #1340, opened 2018-11-28
Publish docs and themes

polyzen:


pazz:
what does `graft` do? please rebase, as i've just pushed your other PR that contains only the first commit.

polyzen:
> include all files under dir

-- https://docs.python.org/3.7/distutils/commandref.html#sdist-cmd

Something like this may also be needed:
https://github.com/pexpect/ptyprocess/pull/22/commits/1fa2011142a200563a9b8fbf303a622ab643cff8

pazz:
should extra, docs and tests all be in the manifest file?

pazz:
let's hear from people who package alot. As far ar I understand this file controls [source distributions](https://packaging.python.org/guides/distributing-packages-using-setuptools/#manifest-in), and technically, docs and extra theming stuff, amd certainly tests are not necessary to run alot. Indeed some packagers separate the docs from the rest into different packages.

polyzen:
See this email and my response:
https://lists.archlinux.org/pipermail/aur-general/2018-November/034618.html

I plan on taking over packaging of alot for Arch if my application is accepted.

polyzen:
https://github.com/masmu/pulseaudio-dlna/blob/master/setup.py#L59-L64

polyzen:
@pazz, do you know who the other maintainers are? Perhaps they could be pinged to get some motion on this.

pazz:
do we really want to distribute tests?

polyzen:
termtosvg and xandikos actually decided to install the tests, heh.

polyzen:
Just noticed you sign your GitHub releases, so will be using those unless you're open to also signing the PyPI releases.

Either way, I believe you should go one step further than publishing by using `data_files` to install these files. Seems the built manual would have to be checked in to be able to use it for that as well.

pazz:
I am not responsible for PyPI releases and cannot remember who uploads these.

PR #1295, opened 2018-08-09
Refactor compose command

dcbaker:
I've had this sitting in a branch for a bit, I meant to write more tests, but I don't know when I'll get time to do it. I figured at the very least it would be nice to put it here.

it needs some more work to get it ready to merge, but it's a start.

PR #1345, opened 2018-12-05
Refactor compose command

pazz:
see #1295

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -806,7 +806,7 @@ class ComposeCommand(Command):
             ui.notify(str(e), priority='error')
             raise self.ApplyError()
     
-    def _set_sender(self, ui):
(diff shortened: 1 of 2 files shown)


pazz:
This merge kind of got ugly due. Let's merge this  as soon as we can to avoid painful rebases in future.

@dcbaker: I believe I've screwed up your tests.. could you have a look?

pazz:
Unless anyone objects, I'll merge this on in shortly.

pazz:
@lucc do you see why the tests fail? It looks to me as if the failing tests have nothing to do with thie branch. I wonder why this doesn't happen on the master branch then?

pazz:
Right, I think i've got them now. The tests were indeed broken during my merge.

pazz:
(this is now part of master)

PR #742, opened 2014-12-15
Config to display in thread buffer status bar all tags present in a thread, or tags common to all messages in thread

andresmrm:
Allows to add **{thread_tags}** or/and **{intersection_tags}** to the Thread Buffer statusbar.
- **thread_tags**: shows all tags present in the current thread.
- **intersection_tags**: shows tags common to all messages in the current thread.

**(New P.R. with squashed commits)**

PR #1346, opened 2018-12-07
Np config show tags thread statusbar

pazz:
see #742 
This is a rebase and a fix for the issue i reported there.


pazz:
@lucc: can I interpret your comment as review and go ahead and merge this? I've been holding back on this for a while due to the annoying issue I had with double-width characters, and which is fixed by my last commit..

pazz:
alright then. I'm pretty sure that it's safe to merge :-)

PR #1352, opened 2018-12-14
Set local variable before useage

lucc:
I just found #1345 again which adds some tests for this command as well. I will postpone further tests until after #1345 for now.


pazz:
cheers

PR #1357, opened 2018-12-19
restore support for regex in section names

dtzWill:
As mentioned in documentation (as I recall),
but wasn't working for some time.

Makes it possible to do things like:

```
[tags]
  [[lists/.*]]
    translation = 'lists/(.*)', 'ʟ\1'
```

To replace lists/foo with ʟfoo (or whatever) :).

### DIFF
--- a/tests/settings/manager_test.py
+++ b/tests/settings/manager_test.py
@@ -5,6 +5,7 @@
 """Test suite for alot.settings.manager module."""
 
 import os
+import re
 import tempfile
 import textwrap
 import unittest
@@ -127,7 +128,7 @@ class TestSettingsManager(unittest.TestCase):
         setting = manager.get_notmuch_setting('foo', 'bar')
         self.assertIsNone(setting)
 
-    def test_dont_choke_on_regex_special_chars_in_tagstring(self):
+    def test_choke_on_invalid_regex_in_tagstring(self):
         tag = 'to**do'
         with tempfile.NamedTemporaryFile(mode='w+', delete=False) as f:
             f.write(textwrap.dedent("""\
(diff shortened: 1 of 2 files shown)


pazz:
Could you elaborate on what exactly went wrong and perhaps where you think our tests are incorrect?
Currently our RE tests fail for your patch.
To test locally:

```
python3 -m unittest tests.settings.manager_test.TestSettingsManager.test_dont_choke_on_regex_special_chars_in_tagstring
```

See `tests/settings/manager_test.py` for the test code.

dtzWill:
Well, looking at the test it appears to be specifically checking
that one cannot use regex's in tag strings?

`to**do` is not a valid regex, so it's unsurprising it should fail.

That example in particular makes a weak case for why someone might
want to specify tag strings without escaping regex characters,
but presumably in other situations it would be nice (?).

This patch causes tagstrings to be interpreted as a regex,
which I thought was the intended behavior.

Doing so means that `to**do` is rejected as invalid,
which personally seems quite acceptable.
(however that means the test should be removed,
perhaps in favor of one checking the desired behavior)


From a programming point-of-view, it really seems the author of the code
intended tagstrings to be allowed to be regex's-- otherwise why bother
checking using `re.match`? Note that as-is AFAICT that's useless,
since it's anchored on both ends (`^` and `$`) and the string between
the anchors has all regex-useful things escaped.

In summary:

* the code as-is suggests behavior other than observed
* the (newly, with this change) failing test is at odds with the intended behavior
  * not due to bug in test but test itself, FWIW (no big deal just
    clarifying)
* this patch restores "intended" behavior, breaking this test
* what anyone intended is my guesswork of course,
  but I thought it was mentioned in examples or other docs?

I'll follow-up with a link to such if I find it.

And regardless of intention, my $0.02 is that regex's in tag strings
are useful, very much so :).
(notmuch commands support regex's in tag queries so it's not just me
hehe)

--

Honestly in practice I don't think it's worth escaping tag strings,
while in theory that's a usability problem for folks expecting
fixed-string behavior, .... it doesn't seem likely for folks
to often put things like "**" (the literal, not glob or any such)
as part of a tag name, especially compared to the amount of folks
using tags with a heirarchy-like scheme (foo/bar, foo/todo, foo::bar
foo::todo, etc).

But I don't really know for sure haha and regardless only mean
to restore what seemed the desired behavior, not try to argue
for my preferred behavior :).

Hope this makes sense, thanks!

pazz:
Thanks for this.
You are right that directly interpreting tagstrings in these configs as REs is the intended behaviour
and I fully support the changes you've mentioned above.
Please update/add tests accordingly and I'll approve this PR.
Cheers,
P

dtzWill:
Okay, I tried :innocent: .  I'm not practiced in python, so apologies if is wrong functionally or stylistically, especially not sure what's idiomatic.

dtzWill:
And while visiting this code, test the actual translation feature I want to use :).

tests/settings/manager_test.py
```diff
@@ -127,7 +128,7 @@ def test_read_notmuch_config_doesnt_exist(self):
         setting = manager.get_notmuch_setting('foo', 'bar')
         self.assertIsNone(setting)
 
-    def test_dont_choke_on_regex_special_chars_in_tagstring(self):
+    def test_choke_on_regex_special_chars_in_tagstring(self):
```
lucc: I think this could better be called `test_choke_on_invalid_regex_in_tagstring`.

tests/settings/manager_test.py
```diff
@@ -136,10 +137,67 @@ def test_dont_choke_on_regex_special_chars_in_tagstring(self):
                         normal = '','', 'white','light red', 'white','#d66'
                 """.format(tag=tag)))
         self.addCleanup(os.unlink, f.name)
+        with self.assertRaises(re.error):
```
lucc: Which line raises the exception? I suppose the last so you should just wrap the last one with the `with` block.

tests/settings/manager_test.py
```diff
@@ -136,10 +137,67 @@ def test_dont_choke_on_regex_special_chars_in_tagstring(self):
...
+                    [[{tag}]]
+                        translated = matched
+                """.format(tag=tagprefixregexp)))
+        self.addCleanup(os.unlink, f.name)
+        manager = SettingsManager()
+        manager.read_config(f.name)
+        translate = lambda t: manager.get_tagstring_representation(t)['translated']
+        matched = lambda t: translate(t) == 'matched'
+        self.assertTrue(all(map(matched, ['foo' 'foobar', tagprefixregexp])))
+        self.assertFalse(any(map(matched, ['bar', 'barfoobar'])))
```
lucc: It is not considered idiomatic python to assign lambda expressions to a name. One can use local functions if one wants a name. But if you just want to push them into `map()` it is even more idiomatic to just use a list or generator comprehension:
```py
def matched(t):
    return manager.get_tagstring_representation(t)['translated'] == 'matched'
self.assertTrue(all(matched(t) for t in ['foo', 'foobar', tagprefixregexp]))
self.assertFalse(any(matched(t) for t in ['bar', 'barfoobar']))
```

tests/settings/manager_test.py
```diff
@@ -136,10 +137,67 @@ def test_dont_choke_on_regex_special_chars_in_tagstring(self):
...
+
+    def test_translate_regexp(self):
+        # Test for behavior mentioned in 108df3df8571aea2164a5d3fc42655ac2bd06c17
+        # namely that translations themselves can use regex
+        tag = "notmuch::foo"
+        section = "[[notmuch::.*]]"
+        translation = "'notmuch::(.*)', 'nm:\\1'" # actual config file doesn't need \\
```
lucc: you can use raw strings to prevent double backslashing: `translation = r"'notmuch::(.*)', 'nm:\1'"`. Then you can leave out the comment.

tests/settings/manager_test.py
```diff
@@ -136,10 +137,67 @@ def test_dont_choke_on_regex_special_chars_in_tagstring(self):
...
+            f.write(textwrap.dedent("""\
+                [tags]
+                    {section}
+                        translation = {translation}
+                """.format(section=section, translation=translation)))
+        self.addCleanup(os.unlink, f.name)
+        manager = SettingsManager()
+        manager.read_config(f.name)
+        translate = lambda t: manager.get_tagstring_representation(t)['translated']
+        self.assertEqual(translate(tag), translated_goal)
```
lucc: see above

tests/settings/manager_test.py
lucc: By the way: python has implicit string concatenation of literal strings. So `'foo' 'bar' == 'foobar'`.

pazz:
could you squash this down to one commit for the patch and one for the tests?

pazz:
thanks!

PR #1358, opened 2018-12-20
Rename test files

lucc:
The two main reasons are
- to run `python3 -m unittest discover` without specifying a custom
  `--pattern *_test.py`
- to include the test files automatically when generating the MANIFEST
  file.

This was inspired by #1340. Interestingly `python setup.py test` finds more
tests than `python -m unittest discover`.


polyzen:
> - to include the test files automatically when generating the MANIFEST
file.

Afaik [test_suite](https://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=test_suite) doesn't do that.

[This line](https://github.com/pazz/alot/blob/master/setup.py#L31) could be edited to include the tests in the package, but would also result in them being installed.

polyzen:
I see, I was looking at the wrong docs:
https://docs.python.org/3/distutils/sourcedist.html#specifying-the-files-to-distribute

polyzen:
> ... If you do supply your own MANIFEST, you must specify everything: the default set of files described above does not apply in this case.

pazz:
@lucc no, fine with me.

Issue #1291, opened 2018-08-06
UnicodeDecodeError with Python3


josch:
Hi,

I'm using the current git master (I'm at 5c53edfba561a2ebdca2741c92405dbd69bd469c) with Python3. Which seemed to work fine until today where I got the following backtrace in a search result:

```
DEBUG:ui:global command string: "search -- tag:unread and tag:to-me and not tag:spam"
DEBUG:__init__:mode:global got commandline "search -- tag:unread and tag:to-me and not tag:spam"
DEBUG:__init__:ARGS: ['search', '--', 'tag:unread', 'and', 'tag:to-me', 'and', 'not', 'tag:spam']
DEBUG:__init__:cmd parms {'sort': None, 'query': ['--', 'tag:unread', 'and', 'tag:to-me', 'and', 'not', 'tag:spam']}
DEBUG:manager:Worker process 23126 spawned
DEBUG:manager:Worker process 23126 exited successfully
DEBUG:utils:unquoted header: |mundo ingenia <central@publingenia.club>|
ERROR:ui:Traceback (most recent call last):
  File "/tmp/alot/alot/ui.py", line 711, in apply_command
    cmd.apply(self)
  File "/tmp/alot/alot/commands/globals.py", line 126, in apply
    sort_order=self.order))
  File "/tmp/alot/alot/ui.py", line 405, in buffer_open
    self.buffer_focus(buf)
  File "/tmp/alot/alot/ui.py", line 465, in buffer_focus
    self.update()
  File "/tmp/alot/alot/ui.py", line 656, in update
    self.mainloop.draw_screen()
  File "/usr/lib/python3/dist-packages/urwid/main_loop.py", line 586, in draw_screen
    canvas = self._topmost_widget.render(self.screen_size, focus=True)
  File "/usr/lib/python3/dist-packages/urwid/widget.py", line 144, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/lib/python3/dist-packages/urwid/decoration.py", line 226, in render
    canv = self._original_widget.render(size, focus=focus)
  File "/usr/lib/python3/dist-packages/urwid/widget.py", line 144, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/lib/python3/dist-packages/urwid/container.py", line 1086, in render
    focus and self.focus_part == 'body')
  File "/tmp/alot/alot/buffers/buffer.py", line 19, in render
    return self.body.render(size, focus)
[...]

Issue #725, opened 2014-08-28
Replying to some mails fails with decoding error


telmich:
Trying to reply to a specifc mail I get this error:

'ascii' codec can't encode character u'\xfc' in position 17: ordinal not in range(128)
(check the log for details)

I wonder if decoding to ascii is probably the wrong approach for a terminal having LANG=en_US.UTF-8?

varac:
I get the same, using todays master and python 3:

```
DEBUG:__init__:ARGS: ['reply', '--all']
DEBUG:__init__:cmd parms {'all': True, 'listreply': None, 'spawn': None}
DEBUG:envelope:TEMPLATE: None
DEBUG:utils:unquoted header: |Re: ...|
DEBUG:thread:candidate addresses: [('a', 'a@b.c')]
DEBUG:thread:candidate addresses: [('', 'öwe <oewe@b.c>')]
DEBUG:thread:match!: 'varac@d.e'
DEBUG:thread:using realname: "varac"
DEBUG:thread:using address: varac@d.e
DEBUG:thread:Replying to own message, set recipients to: ['a', 'a@b.c']
ERROR:ui:Traceback (most recent call last):
  File "/home/varac/.local/lib/python3.6/site-packages/alot-0.7-py3.6.egg/alot/ui.py", line 709, in apply_command
    await cmd.apply(self)
  File "/home/varac/.local/lib/python3.6/site-packages/alot-0.7-py3.6.egg/alot/commands/thread.py", line 236, in apply
    my_addresses, mail.get_all('Cc', []))
  File "/home/varac/.local/lib/python3.6/site-packages/alot-0.7-py3.6.egg/alot/commands/thread.py", line 309, in clear_my_address
    new_value.append(formataddr((name, str(address))))
  File "/usr/lib/python3.6/email/utils.py", line 91, in formataddr
    address.encode('ascii')
UnicodeEncodeError: 'ascii' codec can't encode character '\xf6' in position 1: ordinal not in range(128)
```

Alot stumbles on the Cc address, which shows up quoted in the thread buffer:

```
    From         varac <varac@d.e>
    To             a <a@b.c>
    Cc            "öwe <oewe@b.c>"
    ...
```

varac:
This is possible a duplicate of #1300

dcbaker:
[...]

PR #1359, opened 2018-12-24
Add test case for new encoding problems

lucc:
Mentioned in #1291 (sorry it took me so long).

This is just one simple test. We might need to add some more basic test for 8bit or utf8 or so before we can finally fix this.

Maybe the test case can also be formulated without an external file, I did not try that yet.


varac:
It would be good to use utf8 characters in other headers of the mails.

https://github.com/pazz/alot/issues/725#issuecomment-453865460 shows a mail with unicode chars in To, From, Cc, and subject.

Issue #1363, opened 2019-01-07
Warning about msg not being encrypted to all recipients should be case-insensitive


varac:
Warn if mail is not encrypted to all recipients #1232 works well, except the situation where i.e. one recipient has one character in upper case, whereas the gpg has all characters in lower case.

**Software Versions**
- Python version: 3.6.7
- Alot version: master of today

**To Reproduce**
Steps to reproduce the behaviour:
1. compose a mail to `someOne@example.org`
2. encrypt it, using key from `someone@example.org`
3. try to send it
4. Alot will complain

**Error Log**

Nothing relevant in the logs:

```
DEBUG:ui:cmdline: 'send'
DEBUG:ui:envelope command string: "send"
DEBUG:__init__:mode:envelope got commandline "send"
DEBUG:__init__:ARGS: ['send']
DEBUG:__init__:cmd parms {}
```

The UI shows: `Message is not encrypted to all recipients. This means that not everyone will be able to decode and read this message. Do you want to send anyway? (y):yes (n):no`.

varac:
@drebs ^

lucc:
This looks like another symptom of the fact that email address local parts should only be interpreted by the receiving system (the one identified by the domain name of the email address). We had this problem before with local parts that where interpreted case insensitive by the provider but we had trouble matching them against our account addresses in alot.

drebs:
> This looks like another symptom of the fact that email address local parts should only be interpreted by the receiving system (the one identified by the domain name of the email address). We had this problem before with local parts that where interpreted case insensitive by the provider but we had trouble matching them against our account addresses in alot.

Hi, @lucc. I don't understand if your comment is just a remark or if it leads to or suggests some way of handling the present situation.

lucc:
My comment above was a reminder that any solution to the problem must take care to solve essentially the same problems as #1108. Only now for any potential recipient instead of the users own addresses. Essentially it boils down to the question "In which situations do we allow ourselves to deviate from the standard and dare to make assumptions about the local part of an email address?"

pazz:
(see #1364 )

varac:
Thx !

PR #1364, opened 2019-01-07
Compare lowercase recipient and encryption key addresses

drebs:
Alot will check if a certain encrypted message is actually being
encrypted to all its recipients before sending it. Currently, comparison
is taking upper/down-case into account, which leads to malfunction when
recipient is "someOne@example.org" but the key id is actually
"someone@example.org" (or the opposite).

This commit just converts everything to downcase before comparing
addresses.

Closes: #1363

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -174,13 +174,13 @@ class SendCommand(Command):
         addresses = set()
         for key in self.envelope.encrypt_keys.values():
             for uid in key.uids:
-                addresses.add(uid.email)
+                addresses.add(uid.email.lower())
         return addresses
 
     def _get_recipients_addresses(self):
         tos = self.envelope.headers.get('To', [])
         ccs = self.envelope.headers.get('Cc', [])
-        return {a for (_, a) in email.utils.getaddresses(tos + ccs)}
+        return {a.lower() for (_, a) in email.utils.getaddresses(tos + ccs)}
 
     def _is_encrypted_to_all_recipients(self):
         recipients_addresses = self._get_recipients_addresses()


pazz:
Is this safe?
I understand that most email services consider usernames case insensitive.. Does anyone more into crypto  have an opinion?

meskio:
From [RFC 5321](https://tools.ietf.org/html/rfc5321#page-42), it looks like the Local-part "MAY be case-sensitive", but:
> While the above definition for Local-part is relatively permissive,
   for maximum interoperability, a host that expects to receive mail
   SHOULD avoid defining mailboxes [...] where the Local-part is
   case-sensitive.

I will expect most providers to don't allow case-sensitive email addresses.

lucc:
@pazz  the quote @meskio posted is on the spot. The problem about safety in this case is as such:
1. we want to warn the user in order not to do something stupid (send an email to somebody that they can not open)
2. we can not find a key for the recipient so at them moment we give up and warn the user
3. after this PR we will assume that the recipient can read the mail if the address we send to and the address we encrypt to match case insensitively.
4. worst case: the recipient can't read the mail + somebody totally unrelated can read it (unlikely?)
5. side effect: we ignore the standard

pazz:
@lucc: not sure what you are arguing for? In case you want to push this, I'm happy with this. I just wanted to hear a second opinion if this may be a totally bad idea from a security perspective.

lucc:
I am also not sure if I want to push this. I just wanted to clear up what the actual implications (from a security perspective) are.

pazz:
Thanks @lucc and @meskio  for your thoughts on this. I would tend to agree with lucc that it's better to be careful in this situation when there is a possible security issue.
I will close this issue and the related bug report without merging the hotfix.
The original "problem" should in fact be deald with by telling "someOne" off and telling them to sync the address they use and the one mentioned in their key.
Feel free to re-open the issues if necessary.

Issue #1314, opened 2018-09-15
Special characters not displays correctly


varac:
In some mails special characters (in this case german Umlauts) are not displayed correctly in the message view.
I.e. `können` displays as `k�nnen`.
In the same terminal, `notmuch show --format=raw` show the special characters correctly.

In a mail that displays badly I noticed these Content headers:

```
Content-Type           multipart/mixed; boundary="------------E72523F6326A74524F4B3BE4"
Content-Language       de-DE
```

Another mail that displays fine had these headers:

```
Content-Type              text/plain; charset="utf-8"; format="flowed"
Content-Transfer-Encoding 8bit
Content-Language          en-US
```

**Software Versions**
- Python version: 3.6.5
- Notmuch version: 0.26
- Alot version: master

teto:
I also receive japanese messages (using alot master), with the titles displayed perfeclty
```
MIME-Version: 1.0
Content-Type: text/plain; charset="ISO-2022-JP"
Content-Transfer-Encoding: 7bit
Subject: [OFFICE-TYO 15779]
 =?iso-2022-jp?b?W09GRklDRS1BTEwgMTA4NzRdIBskQiFaOm42SDMrO08bKEI=?=
 =?iso-2022-jp?b?GyRCTyJNbSFbGyhCIFsgGyRCJWElcyVGJUolcyU5PnBKcxsoQiAtIA==?=
 =?iso-2022-jp?b?GyRCOERKTBsoQiBdIBskQiVtJTAlJCVzJTIhPCVIJSYlJyUkJTUbKEI=?=
 =?iso-2022-jp?b?GyRCITwlUBsoQih0emstbGdzLWVndzAxKRskQiVhJXMlRiVKJXMbKEI=?=
 =?iso-2022-jp?b?GyRCJTkkTiQqQ04kaSQ7GyhC?=
```
while the actual content appears as:
```

?$B%9%?%C%U3F0L?(B

?$B%0%k!<%W<RFbJs!V?(Bi.X.?$B!W$K?7$7$$5-;v$r7G:\$7$^$7$?!#?(B
```
I don't mind sharing such emails with maintainers (I lurk on irc with the same nickname).

pazz:
@teto could you test this mail on [this](https://github.com/pazz/alot/pull/1344) branch and report back with a logfile?

pazz:
[...]

Issue #1310, opened 2018-08-28
Save an email


WilliamMayor:
**Is your feature request related to a problem? Please describe.**
I'd like to save emails as .eml files. When I get a receipt emailed to me I'd like to save it to my receipts directory for my accountant to deal with.

**Describe the solution you'd like**
A command for saving emails. Or pointers on how to write a custom keybinding that does this.

**Describe alternatives you've considered**
I've tried writing a custom keybind command that uses `pipeto` and combinations of `cat` or `tee` and regular redirects. I can't get it to save the email. Or at least I can, but not if I use spaces in the file name, and adding quotes doesn't work:

    pipeto "tee '/Users/william/Reciepts/DATE NAME.eml'"
    pipeto "cat > '/Users/william/Reciepts/DATE NAME.eml'"
    pipeto "cat" > '/Users/william/Reciepts/DATE NAME.eml'

**Additional context**
I'm very new to alot so I suspect I'm just missing the obvious way to do this. Any and all help appreciated. Thanks!

pazz:
Hi and welcome!

Your use-case sounds as if it might be more helpful to just print the mail as pdf instead.
To do that, you can use muttprint: put the following in your `~/.config/alot/config`:

> print_cmd = muttprint -1 -F Utopia -a From_Date_*Subject* -p TO_FILE:~/print.pdf

then use the `print` command in thread buffers.

If you really wanted to write the email source to a file, there is no command for this as far as I remember.
But you could write your own python hook and call it using `call hookname` from within alot. 

With current alot master, your hook code would look similar to this:

```
async def write_mail(ui):
    # get msg content
    msg = ui.current_buffer.get_selected_message()  # this is an alot.db.Message
    eml = msg.get_email()  # this is an email.Message

    # open file and write str(eml)..
```

Check out the wiki for more inspiration on hooks.
I hope this helps and have fun hacking :)

WilliamMayor:
Hi @pazz, thanks for your reply, and even more thanks for creating alot!

The PDF thing sounds good to me. Looks like I need to install mutt as well? Do I need some kind of HTML to PDF tool as well? Or does muttprint take care of that?

I'll explore the hook option as well.

Thanks!

WilliamMayor:
Hey, I've just tried writing a custom hook. I've added some simple print statements to a function in `.config/alot/hooks.py`. I'm then trying to run that function by running `call FUNCTION` from the alot prompt. It gives me an error that 'name FUNCTION if not defined'.

[...]

PR #1369, opened 2019-01-16
Add test case for encoding problems

teto:
Some of my Japanese mails are not displayed correctly
https://github.com/pazz/alot/issues/1314

For instance the text from tests/static/mail/japanese.eml appear on my terminal as:
```
MA-EYES?$B$4MxMQ<T3F0L?(B

BIRD-BO?$B$N2OOB$G$9!#$*Hh$lMM$G$9!#?(B
```
my terminal is perfectly capable of displaying kanjis, alot even shows the subject correctly, just the body is messed up.

It would be great if someone could pick this up. When I launched the tests, it failed on some other test (using python3.7) 
```
FAIL: test_env_set (tests.helper_test.TestCallCmdAsync)
  File "/home/teto/alot/tests/utilities.py", line 188, in _actual
    return loop.run_until_complete(coro(*args, **kwargs))
  File "/nix/store/ydk0mfpvn9smcmn72wc9i20slv1d2b79-python3-3.7.2/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "/home/teto/alot/tests/helper_test.py", line 424, in test_env_set
    self.assertEqual(ret[0], 'bar')
AssertionError: '' != 'bar'
+ bar

```

### DIFF
--- a/tests/db/utils_test.py
+++ b/tests/db/utils_test.py
@@ -719,6 +719,19 @@ class TestExtractBody(unittest.TestCase):
         expected = "Liebe Grüße!\n"
         self.assertEqual(actual, expected)
 
+    @unittest.expectedFailure
+    def test_simple_japanese_file(self):
+        mail = email.message_from_binary_file(
+                open('tests/static/mail/japanese.eml', 'rb'))
+        actual = utils.extract_body(mail)
+        expected = """
+            MA-EYESご利用者各位
+
+            BIRD-BOの河和です。お疲れ様です。
+        """
+        self.assertEqual(actual, expected)
+
+
 class TestMessageFromString(unittest.TestCase):
 
     """Tests for decrypted_message_from_string.
--- a/tests/static/mail/japanese.eml
+++ b/tests/static/mail/japanese.eml
@@ -1,0 +1,10 @@
+From: teto@github
+To: tests@alot
+Subject: Japanese email
+MIME-Version: 1.0
+Content-Type: text/plain; charset="ISO-2022-JP"
+Content-Transfer-Encoding: 7bit
+
+MA-EYESご利用者各位
+
+BIRD-BOの河和です。お疲れ様です。


lucc:
Are you sure the test file you add is in valid email format according to the RFC? `file tests/static/mail/japanese.eml --mime` tells me it is `message/rfc822; charset=utf-8` but the header claims it is encoded in `7bit` and the charset is `ISO-2022-JP`. That sounds wrong to me.

teto:
I looked at the original mail files in my maildir directory and they are indeed: `message/rfc822; charset=us-ascii`. I've tried to save the test back to us-ascii in vim via `:set fileencoding=us-ascii` or ```
iconv
iconv -f US-ASCII -t UTF-8 
``` but it can't succeed as it's not possible to convert the kanji to a simple ascii I guess. Still the same mail is displayed fine in neomutt.

**Comment by teto** [comment]
here is another case where it doesn't work with a different kind of output
[bad_mail.txt](https://github.com/pazz/alot/files/2835023/bad_mail.txt)
with the edited screenshot.
2019-02-06-145519_636x757_scrot

do note that some japanese messages open up correclty.

**Comment by teto** [comment]
Regarding the mimetype of japanese.eml, it might be because I saved the mail with a custom hook:
```
async def save_mail(ui): 
    # inspired by https://github.com/pazz/alot/issues/1310
    # get msg content 
    msg = ui.current_buffer.get_selected_message()
    # this is an alot.db.Message
    eml = msg.get_email() # this is an email.Message
    # open file and write str(eml)..
    with tempfile.NamedTemporaryFile(mode='w+', prefix="alot-", delete=False) as out:
        out.write(str(eml))

        ui.notify("saved to %s" % out.name, priority='normal', timeout=15)

```
Neomutt displays the mail correctly but I much prefer alot. Really looking forward to a solution for this. I would like to add hooks to translate the mails etc.

**Comment by lucc** [comment]
I am sorry for my laziness regarding the whole encoding and display topic. I hope to have more time available for this starting next week.

**Comment by teto** [comment]
Sorry if that sounded entitled. Take your time it's open source we should enjoy it no problem :) 
I am available on IRC should you have any question.

**Comment by teto** [comment]
how am I supposed to learn Japanese if I can't read my mails :p

**Comment by pazz** [comment]
I have updated the test according to what our other tests do and indeed it fails:

```
FAIL: test_simple_japanese_file (tests.db.test_utils.TestExtractBody)
  File "/home/pazz/projects/alot/tests/db/test_utils.py", line 722, in test_simple_japanese_file
    self.assertEqual(actual, expected)
AssertionError: 'MA-EYES������������������\n\nBIRD-BO���������������������������������������\n' != '\n            MA-EYESご利用者各位\n\n            BIRD-BOの河和です。お疲れ様です。\n        '
- MA-EYES������������������
  
- BIRD-BO���������������������������������������
+             MA-EYESご利用者各位
+ 
+             BIRD-BOの河和です。お疲れ様です。
+       
```

To me it looks like this file is utf8 encoded, but the header explicitly announces the content as `ISO-2022-JP`, so it will not be interpreted as expected.
This all looks just as I would expect as the mail simply is malformed.
We can discuss whether alot should be lenient in these cases but I'd rather stay in line with what python email module does..

Issue #1360, opened 2018-12-24
Missing characters in extracted body


lucc:
For some content transfer encodings and charsets some characters are missing in the extracted body. An example can be seen in the test added in #1359. As that test demonstrates the problem seems to be somewhere below `alot.db.utils.extract_body()`.  We should try to find more relevant combinations of encoding and charset, add tests and ultimatly fix it :)

**Software Versions**
- Python version: 3.7
- Alot version: 0.8-4-g9ff655ba

lucc:
It is marked as an expected failure because the bug is known and we did not yet fix it. I created this test after the bug report was issued in order to help us fix it. See the #1359 I mentioned above.

If you want to provide more test cases that is very much appreciated. So if you have any *other* combinations of encoding and charset I am very interested.

jljusten:
@lucc: Hey. I think you've been the only one to look into this issue, but I guess it's stalled a bit?

We're holding off on packaging 0.8 for debian due to this issue. We would probably cherry-pick a patch for this back to the 0.8 release if a fix is found.

mjg:
The problem bisects back to commit 176cffcd ("refactor alot.db.utils.remove_cte", 2018-12-04) which claims to refactor and make the code more lenient.

In fact, it does one more thing: it changes the fallback for the case of decoding errors to `bp.decode(enc, errors='ignore')` rather than `helper.try_decode(bp)`. Whether this is more lenient or not is a matter of definition; in any case it's the cause of the missing characters. I don't know if this change is intentional and maybe improves other things. The "fix" for me is a one-liner, obviously, but may restore behaviour that this commit wanted to change.  See PR #1375

In general, I'd suggest to refactor in one commit and change behaviour in another one...

pazz:
I guess this can be closed then?

josch:
Yes, it is fixed by PR #1375.

PR #1375, opened 2019-02-04
restore lenient libmagic decoding

mjg:
176cffcd ("refactor alot.db.utils.remove_cte", 2018-12-04) created a few
problems with 8bit quoted-printable e-mails, see #1291 #1360.

This commit restores the old libmagic fallback which did not cause this
problem.

(Maybe just goood as a 0.8 hotfix.)

### DIFF
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -446,9 +446,9 @@ def remove_cte(part, as_string=False):
             sp = helper.try_decode(bp)
         except UnicodeDecodeError as emsg:
             # the mail contains chars that are not enc-encoded.
-            # try again and just ignore those
+            # libmagic works better than just ignoring those
             logging.debug('Decoding failure: {}'.format(emsg))
-            sp = bp.decode(enc, errors='ignore')
+            sp = helper.try_decode(bp)
         return sp
     return bp
 
--- a/tests/db/test_utils.py
+++ b/tests/db/test_utils.py
@@ -711,7 +711,6 @@ class TestExtractBody(unittest.TestCase):
 
         self.assertEqual(actual, expected)
 
-    @unittest.expectedFailure
     def test_simple_utf8_file(self):
         mail = email.message_from_binary_file(
                 open('tests/static/mail/utf8.eml', 'rb'))


josch:
I just backported the change to version 0.8 that we are about to upload to Debian and my results are:

 - I can confirm that this is not a regression of the issue I reported as #1291, so no error message shown with python3 in affected emails
 - It fixes #1360 in that characters are not missing from messages anymore (as also evident by this change fixing that testcase)

pazz:
Thanks for this josh and mjg!

@lucc: what do you think?
Would this justify a minor release?

mjg:
The force-pushed change shows that with this commit, the "expected failure" is turned into success.

What I don't know is whether the behavour change introduced in commit #176cffc was intentional, based on a concrete example, or just on the rationale "if there decode errors then ignore undecodable bytes", which seems reasonable. I'm wondering: Is the message from the test case RFC compliant? It certainly is in a very common format. It is exactly this type that .decode() fails to decode but on which libmagic succeeds. Maybe we can trace what libmagic actually does/guesses to understand what is going wrong? Maybe just a different CTE that we need to pass to decode()?

dcbaker:
Libmagic actually does some sort of analysis to figure out what the content encoding is, the `remove_cte` function trusts that a) the payload is truthful in it's encoding or b) that it does what the RFC requires and is 7bit if it doesn't declare an encoding.

FWIW, I think this is the right thing to do; try to trust the payload, if it lied/is wrong then fall back to libmagic.

Or, bump the version to 3.6 which IIRC can actually do this correctly.

pazz:
@dcbaker: by "bump the version" you mean have alot eepend on python \ge 3.6 ?

jljusten:
@pazz, @dcbaker: I believe we are seeing this with python 3.7 on debian.

lucc:
@pazz sorry I'm late. I also vote for 0.8.1

pazz:
OK, I've just tagged this bugfix release (but won't announce it on the notmuch list)

jljusten:
@pazz: I think maybe you didn't change alot/__init__.py for the version before tagging, so the tarball content doesn't match the v0.8.1 tag. ?

pazz:
@jljusten: thanks and sorry: I only pushed the tags, not the last commit. It should be fixed now.

Issue #1372, opened 2019-02-01
send mail with msmtp dit not work after 0.8


GuillaumeSeren:
Hello thank you for alot it is the best MUA !

Before you submit a bug report, please make sure that the issue still exists on the master branch!
I can reproduce it on 0.8 and master (fff58de316f31267923cad28670ce6c033e93c83).

I use alot since a long time, and I am working to package the 0.8 for gentoo linux,
the problem is the following, when I send a mail it hang's up stay for ever on 'sending..'.

**Software Versions**
- Python version:
python3.6
- Notmuch version: [if relevant]
notmuch-0.27
- Alot version:
alot-0.8 - fff58de3

**To Reproduce**
* Compose a mail
* Send !

**Error Log**
Here the sending part of the log
```
DEBUG:ui:Got key (['y'], [121])
DEBUG:ui:cmdline: 'send'
DEBUG:ui:envelope command string: "send"
DEBUG:__init__:mode:envelope got commandline "send"
DEBUG:__init__:ARGS: ['send']
DEBUG:__init__:cmd parms {}
DEBUG:envelope:FROM: "Mr Guillaume Seren <guillaumeseren@gmail.com>"
DEBUG:envelope:ACCOUNT: "guillaumeseren@gmail.com"
(...)
DEBUG:helper:CMD = [b'msmtp', b'-t', b'--read-envelope-from']
DEBUG:ui:Got key (['q'], [113])
DEBUG:ui:cmdline: 'exit'
DEBUG:ui:envelope command string: "exit"
```
And msmtp is responding as usual
```
Feb 01 12:39:25 host=smtp.gmail.com tls=on auth=on user=guillaumeseren from=guillaumeseren@gmail.com recipients=masked@gmail.com mailsize=423 smtpstatus=250 smtpmsg='250 2.0.0 OK 1549021165 s132sm1443836wmf.28 - gsmtp' exitcode=EX_OK
```
When doing the same with the 0.7 alot get the response from msmtp, and close the small 'sending..' bar,
as alot is now python-3 there is a lot of changes but I can reproduce on master as well.

Guillaume

GuillaumeSeren:
Hey @pazz thank you for the quick feedback.

> 1. have you tried hardcoding the full path to the msmtp binary?

No I kept the same setup / config that worked for the 0.7, will try that.

> 2. I'm not using the msmtp version from my packager either because something similar hung 

[...]

PR #1381, opened 2019-02-11
Spam-marking command

FiskerLars:
I have been increasingly been taken aback from the amount of steps I had to undertake to feed manually tagged spam to feeding my bayesian classifier (spamassassin here). Maybe I got it wrong, but 

1. labelling a thread as spam (7 keypresses)
2. opening up a shell and feeding all recently spam-marked emails to the classifier
3. removing all spam from the inbox

is way to complicated for a daily task.

Thus, I wanted something like 'S' to do all of this for me and created a command with a neat hook. (Plus, further down the line, I had to make the sendmail-send-command synchronous, as I ran into what seems to me like a racing condition for slower sendmails, i.e. /dev/stdin is closed or not opened when sendmail requires the data, leading to a failure.)

Hope this seems useful for you as well. Thanks for the great email client, it is helping me with my communication a lot ;-)


lucc:
Hello @FiskerLars nice that you want to contribute. As it stands it is not clear why alot can not yet do the things you want to do. You know that you can bind longer command sequences and hooks to a single key press?

For the sync/async problem: Can you provide a test case that excises the problem? We are hoping to block the user interface as little as possible so making something explicitly synchronous needs a good justification.

For new features like commands or options: You should definitely provide documentation for those. Otherwise you are the only one who knows how to use it and then all you efforts for the community are in vain. If possible you should also provide tests.

pazz:
is the sendmail thing related to #1372?

I agree that the spam-marking stuff should go into a custom hook and with some documentation also into our [wiki](https://github.com/pazz/alot/wiki/Contrib-Hooks).  Thanks

FiskerLars:
Hi,
sorry for not checking the issue-list beforehand and for having to unrelated patches in one pull request.

The race condition occurs (probably in my specific case) within a wrapper script around esmtp. In order to queue a mail some file-system-action occurs before stdin is read. I can only assume that the pipe is teared down before the read happens. 

For clarification, the read happens very much to the end. At that point stdin is empty. The problem strangely only occurred after this script had been used for well over a year:

`queue_mail() { # ($@)
    local ret=0
    mkdir -p "$qdir" || {
        echo "unable to create queue dir $qdir" 2>&1
        return 1
    }
    mdir="`$mktemp -d "$qdir/XXXXXXXX"`"
    [ -d "$mdir" ] || {
        echo "unable to create tempdir inside $qdir" 2>&1
        return 1
    }
    touch "$mdir/lock"
    echo "$@" >"$mdir/cmd" || ret=1
    cat </dev/stdin >"$mdir/mail" || ret=1
    [ $ret -eq 0 ] || {
        echo "could not enqueue mail in dir $mdir" 2>&1
        return 1
    }
`

Writing on this topic, I could try to have the read from stdin earlier in the code and try that

lucc:
I took the liberty to put block code markers around your code. But it looks fragmented and the braces don't match. Is that the full script you are using? What interpreter is it using? What does it do?

pazz:
I'm closing this because just like @lucc I think this feature should not be part of mainline alot.
Instead, manual spam-marking should for now be done in custom hooks (which you can bind to keys if you like via the `call` command). You may also want to have a look at [afew](https://github.com/afewmail/afew).

Finally, next time you send a PR, please don't mix in a `fix` for the unrelated sendmail issue and open a separate issue instead. Cheers!

PR #1390, opened 2019-04-10
expose message object to reply/forward_prefix hook

NickHu:
This exposes the message object in the `reply_prefix` and `forward_prefix` hooks; now, you can mimic what Gmail does like so:

```python
def forward_prefix(*args, **kwargs):
  # ---------- Forwarded message ---------
  # From: Nick Hu <my@fake.email>
  # Date: Wed, 10 Apr 2019 at 11:03
  # Subject: Re: Cool mail
  # To: Someone Else <recipient@of.cool.mail>
  subject = alot.db.utils.decode_header(kwargs["message"].get("Subject", ""))
  to = alot.db.utils.decode_header(kwargs["message"].get("To", ""))
  return (" ---------- Forwarded message ---------\n"
             " From: {0} <{1}>\n"
             " Date: {2:%a, %-d %b %Y at %H:%M}\n"
              " Subject: {3}\n"
              " To: {4}\n\n\n"
             ).format(*args, subject, to)
```

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -139,7 +139,8 @@ class ReplyCommand(Command):
         timestamp = self.message.get_date()
         qf = settings.get_hook('reply_prefix')
         if qf:
-            quotestring = qf(name, address, timestamp, ui=ui, dbm=ui.dbman)
+            quotestring = qf(name, address, timestamp,
+                             message=mail, ui=ui, dbm=ui.dbman)
         else:
             quotestring = 'Quoting %s (%s)\n' % (name or address, timestamp)
         mailcontent = quotestring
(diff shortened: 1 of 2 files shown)


pazz:
nice one.
I have a few comments: 
1) why only forwards and not replies?
2) please fix the pep8 issue
3) (jsut a thought) Perhaps therse hooks should instead of hardcoding a few "interesting" headers, just receive the whole original message as `email.Message` object..

NickHu:
I only exposed what was needed to replicate what Gmail does, but if you prefer exposing the whole Message object, I can make that change.

pazz:
Sure, thanks. In any case, I think this should be backward-compatible, by
handing the email obj as named parameter..

Quoting Nick Hu (2019-04-10 13:34:35)

NickHu:
@pazz I implemented this as requested; good to merge?

pazz:
thanks for the review @lucc. I agree that this is potentially problematic for people who have already defined hooks without a catch-all parameter a la `**kwargs`. 
I'll merge it anyway because I think it is a nice feature. There should be a warning in the NEWS file once we release.

Issue #1393, opened 2019-04-29
Double width characters in search view breaks column structure


cagprado:
**Describe the bug**
When using `width = 'fit', ...` for search view, if the column contains double width characters (e.g. East Asian), the next column is pushed right one space for each double width char.

I can observe the issue in Authors and Subject columns but I haven't tested for double width characters in other columns. The issue seems to be more easily spotted if both minimum and maximum widths for the `fit` option are the same but it also happens otherwise.

**Software Versions**
- Python version: 3.7.3
- Alot version: Master branch d0297605

**To Reproduce**
Steps to reproduce the behaviour:
1. configure theme to use fixed width columns.
2. query for any emails containing double-width characters either in it's authors list or subject.
3. columns misalign whenever double-width characters are present.

**Error Log**
Interface bug, no error generated.

**Possible Solution Idea**
I didn't investigate too much about it but I could quickly solve my current problem with a single line edit.  I don't know if that breaks other things, though.  Seems like `text.pack()` can still return a width which is larger than the one defined by the theme in some special cases, this change tells urwid what is the maximum width available so the result will not exceed this value.

    *** alot/widgets/search.fix.py        2019-04-29 15:18:24.756820581 +0800
    --- alot/widgets/search.py            2019-04-29 15:18:17.386794207 +0800
    ***************
    *** 156,162 ****
      
          # define width and part_w
          text = urwid.Text(content, wrap='clip')
    !     width = text.pack((maxw or minw,))[0]
          part_w = AttrFlipWidget(text, struct)
      
          return width, part_w
    --- 156,162 ----
      
          # define width and part_w
          text = urwid.Text(content, wrap='clip')
    !     width = text.pack()[0]
          part_w = AttrFlipWidget(text, struct)
      
          return width, part_w

pazz:
Thanks for the detailed bug report and suggested fix!
I believe that I've hot-fixed urwids handling of double width chars elsewhere exactly like this.
Would you mind sending your patch as PR?
Cheers,
P

cagprado:
Sure will do! Thanks!

PR #1397, opened 2019-05-14
Fix double width characters in search view

cagprado:
Hotfix for double width characters causing columns misalign in search
view (closes https://github.com/pazz/alot/issues/1393)


pazz:
(I've pushed this overiding the missing travis check because it fails for other reasons).

PR #1398, opened 2019-05-15
Faster Thread loading

pazz:
This will prevent the whole message from being read and interpreted
at the time we instantiate a MessageTree for display and instead only
create the (cheap!) summary widget.

When a user manually interacts with the Message widgets (for example by
unfolding/toggling source) then the content parts will anyway be
reassembled. The consequence of this patch is that loading large threads
should be much faster.

### DIFF
--- a/alot/commands/search.py
+++ b/alot/commands/search.py
@@ -38,9 +38,9 @@ class OpenThreadCommand(Command):
             query = ui.current_buffer.querystring
             logging.info('open thread view for %s', self.thread)
 
-            sb = buffers.ThreadBuffer(ui, self.thread)
-            ui.buffer_open(sb)
-            sb.unfold_matching(query)
+            tb = buffers.ThreadBuffer(ui, self.thread)
+            ui.buffer_open(tb)
+            tb.unfold_matching(query)
 
 
 @registerCommand(MODE, 'refine', help='refine query', arguments=[
--- a/alot/widgets/thread.py
+++ b/alot/widgets/thread.py
@@ -191,6 +191,19 @@ class MessageTree(CollapsibleTree):
         logging.debug('DHT %s', str(self._default_headers_tree))
         logging.debug('MAINTREE %s', str(self._maintree._treelist))
 
+    def expand(self, pos):
+        """
+        overload CollapsibleTree.expand method to ensure all parts are present.
+        Initially, only the summary widget is created to avoid reading the
+        messafe file and thus speed up the creation of this object. Once we
+        expand = unfold the message, we need to make sure that body/attachments
+        exist.
+        """
+        logging.debug("MT expand")
+        if not self._bodytree:
+            self.reassemble()
+        CollapsibleTree.expand(self, pos)
+
     def _assemble_structure(self, summary_only=False):
         if summary_only:
             return [(self._get_summary(), None)]
--- a/alot/widgets/thread.py
+++ b/alot/widgets/thread.py
@@ -169,7 +169,7 @@ class MessageTree(CollapsibleTree):
         self._default_headers_tree = None
         self.display_attachments = True
         self._attachments = None
-        self._maintree = SimpleTree(self._assemble_structure())
+        self._maintree = SimpleTree(self._assemble_structure(True))
         CollapsibleTree.__init__(self, self._maintree)
 
     def get_message(self):
@@ -191,7 +191,10 @@ class MessageTree(CollapsibleTree):
         logging.debug('DHT %s', str(self._default_headers_tree))
         logging.debug('MAINTREE %s', str(self._maintree._treelist))
 
-    def _assemble_structure(self):
+    def _assemble_structure(self, summary_only=False):
+        if summary_only:
+            return [(self._get_summary(), None)]
+
         mainstruct = []
         if self.display_source:
             mainstruct.append((self._get_source(), None))


lucc:
I have this binding and for some reason it does not work with this patch here.
```
  [[search]]
    enter = select; fold *; unfold tag:unread; move last; unfold
```
The `fold *` somehow *unfolds* all messages in the thread.

The speed improvement is especially noticeable  (for me) with a large thread with many text/html messages that have to be rendered by elinks.

For people who want to test the speed improvements you can find interesting threads like this:
```sh
notmuch search --format=json mimetype:text/html | jq 'sort_by(.matched)' 
alot search thread:SOME_ID
```

pazz:
Quoting Lucas Hoffmann (2019-05-17 08:32:13)
> I have this binding and for some reason it does not work with this patch here.
> 
>   [[search]]
>     enter = select; fold *; unfold tag:unread; move last; unfold
> 
> The fold * somehow unfolds all messages in the thread.


The patch only causes the initial creation of the thread buffer to be faster.
the "fold *" command will iterate over all messages and cause them to be rebuild, regardless of them being collapsed or not. This is why your full binding will not be noticably faster.


 
> The speed improvement is especially noticeable (for me) with a large thread
> with many text/html messages that have to be rendered by elinks.

Yes, this was the idea: to only render a message once it is being unfolded.

> For people who want to test the speed improvements you can find interesting
> threads like this:
> 
> notmuch search --format=json mimetype:text/html | jq 'sort_by(.matched)'
> alot search thread:SOME_ID
>
Thanks, that's helpful.

lucc:
I found a mail which triggers this exception and kills alot.  The Trace is not printed in the logfile but direrctly on the tty:
```
  File "/home/luc/.local/bin/alot", line 11, in <module>
    load_entry_point('alot', 'console_scripts', 'alot')()
  File "/home/luc/vcs/alot/alot/__main__.py", line 137, in main
    UI(dbman, cmdstring)
  File "/home/luc/vcs/alot/alot/ui.py", line 141, in __init__
    self.mainloop.run()
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 286, in run
    self._run()
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 384, in _run
    self.event_loop.run()
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 1340, in run
    reraise(*exc_info)
  File "/usr/lib/python3.7/site-packages/urwid/compat.py", line 58, in reraise
    raise value
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 1354, in wrapper
    rval = f(*args,**kargs)
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 1313, in _twisted_idle_callback
    callback()
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 572, in entering_idle
    self.draw_screen()
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 586, in draw_screen
    canvas = self._topmost_widget.render(self.screen_size, focus=True)
  File "/usr/lib/python3.7/site-packages/urwid/widget.py", line 144, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/lib/python3.7/site-packages/urwid/decoration.py", line 226, in render
    canv = self._original_widget.render(size, focus=focus)
  File "/usr/lib/python3.7/site-packages/urwid/widget.py", line 144, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/lib/python3.7/site-packages/urwid/container.py", line 1086, in render
    focus and self.focus_part == 'body')
  File "/home/luc/vcs/alot/alot/buffers/thread.py", line 136, in render
    return self.body.render(size, focus)
  File "/usr/lib/python3.7/site-packages/urwid/widget.py", line 144, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/lib/python3.7/site-packages/urwid/widget.py", line 1765, in render
    canv = get_delegate(self).render(size, focus=focus)
  File "/usr/lib/python3.7/site-packages/urwid/widget.py", line 144, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/lib/python3.7/site-packages/urwid/listbox.py", line 471, in render
    (maxcol, maxrow), focus=focus)
  File "/usr/lib/python3.7/site-packages/urwid/listbox.py", line 353, in calculate_visible
    self._set_focus_complete( (maxcol, maxrow), focus )
  File "/usr/lib/python3.7/site-packages/urwid/listbox.py", line 734, in _set_focus_complete
    middle,top,bottom=self.calculate_visible((maxcol,maxrow),focus)
  File "/usr/lib/python3.7/site-packages/urwid/listbox.py", line 356, in calculate_visible
    focus_widget, focus_pos = self._body.get_focus()
  File "/home/luc/.local/lib/python3.7/site-packages/urwidtrees/widgets.py", line 75, in get_focus
    return self._get(self._focus)
  File "/home/luc/.local/lib/python3.7/site-packages/urwidtrees/widgets.py", line 68, in _get
    res = self[pos], pos
  File "/home/luc/.local/lib/python3.7/site-packages/urwidtrees/widgets.py", line 54, in __getitem__
    entry = self._tree.get_decorated(pos)
  File "/home/luc/.local/lib/python3.7/site-packages/urwidtrees/nested.py", line 86, in get_decorated
    return self._get_decorated_entry(self._tree, pos)
  File "/home/luc/.local/lib/python3.7/site-packages/urwidtrees/nested.py", line 82, in _get_decorated_entry
    entry = tree.decorate(pos[0], entry, is_first=isf)
  File "/home/luc/.local/lib/python3.7/site-packages/urwidtrees/decoration.py", line 475, in decorate
    line = urwid.Columns(cols, box_columns=range(len(cols))[:-1])
  File "/usr/lib/python3.7/site-packages/urwid/container.py", line 1768, in __init__
    if focus_column is None and w.selectable():
AttributeError: 'NoneType' object has no attribute 'selectable'
```
Maybe this already helps to find the problem. I can construct a MWE during next week.

pazz:
Interesting! And this only happens on this branch?
It looks like the message summary is empty, or missing some component. But this should then also be a problem on the master branch...


Quoting Lucas Hoffmann (2019-05-18 18:30:06)

lucc:
It does not happen on the master branch. The conditions are as follows:
1. you need an encrypted message, so `:search is:entrypted`.
2. `:select` a thread
3. move to the last message `G`. All other messages are not relevant
4. `:fold` this message; yes `:fold` even though it looks folded already. (You can also do `:fold *` instead of `:move last; fold`)
5. `:move last`/`G` to the last *line*
6. Now `:unfold` this message -> this triggers the exception
7. I was not able to trigger the bug on any unencrypted message.

I noticed that the initial `:fold *` actually unfolds the messages visually. Additionally any attachments are *not* displayed during the visual unfolding from `:fold`. Only with a further `:unfold` the attachments appear. 

I assume that the `:fold` in a newly leaves the interface in some inconsistent state. Maybe the messages should be considered folded if only the message summary is shown. So the internal state would match the visible state.

lucc:
OK I can now trigger the bug on any encrypted message that is longer than one line. 
1. `select` the thread
2. `fold` the message (so that it will be visually unfolded)
3. move to any line except the first line eg `:move down; move down`
4. `unfold` -> booom!

Strangely enough the an `application/gpg` attachment is only shown with encrypted messages after a second `unfold` or upon a `:select` that implicitly unfolds the message.  Both in `master` and here.

pazz:
@lucc  I could reproduce with other messages as well. The problem seemed to be that you should not have needed to "fold" to see the message: fold means hide, unfold means show.
You did not see the msg content because It wasn't computed directly after the `select`, which was a mistake. Please test again.

pazz:
:lucc I'm tempted to push this soon. OK with you?

lucc:
I have used this code for a while now and the previously mentioned bug is fixed.  My problem with unfolding all messages after a plain `:select`ion of a thread is that I don't see any speed improvement over master.  For me one of my longest thread with many html messages takes about 3 seconds to display and handle key presses. The only difference between master and this branch is that master displays the search window for 3 seconds and then the unfolded thread wheras this branch switches from the search window to the folded thread window pretty fast and then takes three seconds until all messages are unfolded and the ui becomes responsive again.

As the pull request is called "faster thread loading" I do not see the point.  I would prefer the threads to not be unfolded. I can still add a manual `unfold *` to my keybindings if I (or any user) want that.

Another thing I noticed is that for some threads (I don't yet know the exact conditions but it is consistently reproducible with the same threads every time) not all messages are unfolded upon opening the thread buffer.  And for some threads not all messages get folded on the first `fold *` (also reproducible but I don't know the exact properties of the thread that trigger it).

In my opinion at least the first point should be fixed before a merge.

pazz:
Thanks for your response!

Quoting Lucas Hoffmann (2019-06-29 23:58:20)
> The only difference between master and this branch is that
> master displays the search window for 3 seconds and then the unfolded thread
> wheras this branch switches from the search window to the folded thread window
> pretty fast and then takes three seconds until all messages are unfolded and
> the ui becomes responsive again.

Yes, this is exactly the intended change.
The thinking is that the thread is opened quickly but of course, reading every message from disk and firing up an html renderer still takes its time (perhaps we can run these concurrently somehow?).

> As the pull request is called "faster thread loading" I do not see the point. I
> would prefer the threads to not be unfolded. I can still add a manual unfold *
> to my keybindings if I (or any user) want that.

My use case is that I have some very long threads where I'm only interested in the newest message, the only one tagged "inbox". The default behaviour of the 'select' command from my inbox would be to open the thread and unfold only this last message, which is pretty quick.

 
> Another thing I noticed is that for some threads (I don't yet know the exact
> conditions but it is consistently reproducible with the same threads every
> time) not all messages are unfolded upon opening the thread buffer. And for
> some threads not all messages get folded on the first fold * (also reproducible
> but I don't know the exact properties of the thread that trigger it).
> 
> In my opinion at least the first point should be fixed before a merge.

Only those matching the previous search query should be unfolded.
This is the intended and, at least on my end, observed behaviour.

lucc:
I would argue that unfolding the messages that match the search query of the search buffer where we did `select` the thread works for some workflows only. I tend to have longer threads where severall messages are tagged `inbox` so these would all get unfolded although I am only interested in the newest one, which in my case is tagged `unread`. And if I execute any other query the interesting message might or might not be matching the search query by which I found the thread and might or might not be in a thread where many other messages match the same query. So I argue that this automatic unfolding scheme only works nicely for a limited range of scenarios.

In my opinion the better default is to not unfold any message and advise the user to unfold some via a mapping or by hand.

pazz:
Fine. This sounds sensible and would also clean up the code base.

lucc:
@pazz how is it going forward on this?

I closed the milestone 0.8.1 just now because that is already released. I also added this PR to 0.9 because I think it would be nice to have there.

pazz:
Sorry I haven't had the time to spend on this. However, the requested change to remove the automatic unfolding should be separate from this PR, which only seeks to  separate message parsing from thread-buffer opening..
I propose we merge this as is (it did not give me any problems and I've been using it a while), and open a new issue for the proposed change.

Issue #1347, opened 2018-12-07
UnicodeEncodeError in formataddr


pazz:
This should not happen when replying to a msg from someone with a funny unicode name:
```
DEBUG:thread:Replying to own message, set recipients to: ['"Michaël C <mc@X>, DC <d@Y>"']
ERROR:ui:Traceback (most recent call last):
  File "/home/pazz/projects/alot/alot/ui.py", line 709, in apply_command
    await cmd.apply(self)
  File "/home/pazz/projects/alot/alot/commands/thread.py", line 239, in apply
    to = ', '.join(self.ensure_unique_address(recipients))
  File "/home/pazz/projects/alot/alot/commands/thread.py", line 321, in ensure_unique_address
    urecipients = [formataddr((n, str(a))) for a, n in res.items()]
  File "/home/pazz/projects/alot/alot/commands/thread.py", line 321, in <listcomp>
    urecipients = [formataddr((n, str(a))) for a, n in res.items()]
  File "/usr/lib/python3.6/email/utils.py", line 91, in formataddr
    address.encode('ascii')
UnicodeEncodeError: 'ascii' codec can't encode character '\xeb' in position 5: ordinal not in range(128)
```

**Software Versions**
- Python version: 3.6.7
- Alot version: master (b6606a5402aa2c50d202e7feb27e74d2c7202681)

pazz:
This happens only when group replying to self-sent messages.
the problem is that the in line [213](https://github.com/pazz/alot/blob/master/alot/commands/thread.py#L213),
`recipients` contains strings that have not yet been encoded into ascii.
In case it is a normal reply, those have been encoded in the preceeding line.

pazz:
It turns out that the mail has a bad To header:

```
From: "Michaël C <mc@X>, DC <d@Y>"
```

This means that `email.Message.get_all('To', [])` returns `'"Michaël C <m@c>"'`.
Now,

```
ReplyCommand.clear_my_address([],['"Michaël C <m@c>"'])
```
fails
but 
```
ReplyCommand.clear_my_address([],['Michaël C <m@c>'])
Out[32]: ['=?utf-8?q?Micha=C3=ABl_C?= <m@c>']
```

varac:
I also get this when trying to tab-complete a recipient from an address book lookup that returns Special characters like german Umlauts.

pazz:
i think this might not be related. can you provide a trace?

pazz:
[...]

PR #1404, opened 2019-05-25
fix #1402

### LINKED ISSUE #1402: UTF8 recipients with a comma get split into incorrect parts when replying
**Describe the bug**
If I receive an e-mail from a sender such as `"Ö, Ä" <x@y.z>` and reply then the recipients will be a list with one recipient 'Ö' and another recipient `Ä <x@y.z>`. In other words: the UTF8-encoded comma in the name is mistaken for a header value list separator.

**Software Versions**
- Python version: python3-3.7.3-3.fc30.x86_64
- Alot version: f81fea2f (current master)

**Error Log**
I have amended `db/utils.py:decode_header()` with debug lines for the passed in header (before `re.sub`) and decoded header right before the `return`.

 When I display the message in alot, the debug reads:
```
DEBUG:utils:passedin header: |"Ö, Ä" <x@y.z>|
DEBUG:utils:unquoted header: |"Ö, Ä" <x@y.z>|
DEBUG:utils:decoded header:  |"Ö, Ä" <x@y.z>|
```
Everything fine so far.

When I reply the following relevant lines (others omitted) are shown before the editor opens:
```
DEBUG:thread:candidate addresses: [('Ö, Ä', 'x@y.z')]
DEBUG:thread:reply to: =?utf-8?q?=C3=96=2C_=C3=84?= <x@y.z>
DEBUG:utils:passedin header: |=?utf-8?q?=C3=96=2C_=C3=84?= <x@y.z>|
DEBUG:utils:unquoted header: |=?utf-8?q?=C3=96=2C_=C3=84?= <x@y.z>|
DEBUG:utils:decoded header:  |Ö, Ä <x@y.z>|
```
So, this is the step where the back and forth of encoding and decoding seems to lose the quotes around the problematic address. But I don't see which step actually introduces the quotes - they are not part of the UTF8-string.

In any case, the `', '.join()` in line 227 of `alot/commands/thread.py´ apparantly needs to be somewhat more careful about assembling that comma separated list.

pazz:
When decoding headers containing email addresses;
Some MUAs (exchange) will add headers in the form:

  To: "Last, First" <x@y.z>

Prior to this commit, alot would remove the quotes (apparently they
violate RFC 2047). However, this then would lead to problems where the
additional comma is interpreted as separator between several recipients.

This commit causes alot to not remove the quotes.
#1402

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -10,7 +10,7 @@ import subprocess
 import tempfile
 import email
 import email.policy
-from email.utils import getaddresses, parseaddr, formataddr
+from email.utils import getaddresses, parseaddr
 from email.message import Message
 
 import urwid
@@ -28,6 +28,9 @@ from ..completion import ContactsCompleter, PathCompleter
 from ..db.utils import decode_header
 from ..db.utils import extract_headers
 from ..db.utils import extract_body
+from ..db.utils import formataddr
+from ..db.utils import clear_my_address
+from ..db.utils import ensure_unique_address
 from ..db.envelope import Envelope
 from ..db.attachment import Attachment
 from ..db.errors import DatabaseROError
@@ -203,7 +206,7 @@ class ReplyCommand(Command):
             # make sure that our own address is not included
             # if the message was self-sent, then our address is not included
             MFT = mail.get_all('Mail-Followup-To', [])
-            followupto = self.clear_my_address(account, MFT)
+            followupto = clear_my_address(account, MFT)
             if followupto and settings.get('honor_followup_to'):
                 logging.debug('honor followup to: %s', ', '.join(followupto))
                 recipients = followupto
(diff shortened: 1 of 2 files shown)


pazz:
I have not checked wether this introduces problems with gpg encoded or signed messages.

pazz:
I have added the second commit to this beanch because it solves a very similar issue #1347 to #1404.

mjg:
This does not solve #1402, unfortunately. I'm not sure where to discuss this, but as far as I understand:

#1347 is about header parts which are not decoded when alot expects them to be.

#1402 (my BR) is about a comma wrapped in a properly UTF8 encoded header (realname part) which - after decoding - alot mistakes for a list separator (because it does not add quotes whehere necessary).

#1404 (this PR) removes code which was supposed to remove erronous double quotes.

This sounds like three different things to me.

Note that realname parts containing commas (quoted when ascii, unquoted but encoded when utf8) are produced by Thunderbird as well. So they are common (RFC or not).

I'll see if I can come up with a test - which part of the test suite should it go to?

PR #1406, opened 2019-06-04
ContactsCompletion: use db.utils.formataddr

### LINKED ISSUE #1378: Wrong tab completion for addressbook entries with non-ascii characters.
With alot 0.8 from Debian unstable I can observe the following bug:

In abook I have the address of a person called "Andreas Nüchter" stored. When I compose an email and type "Andr" when prompted for the `To:` field, then it gets completed to:

    To:=?utf-8?q?Andreas_N=C3=BCchter?= <xxx@yyy>,

This is a regression. This bug did not happen with alot 0.7.

mjg:
email.utils.formataddr does more encoding than we need at this point -
headers will be encoded when they get inserted anyway.

Use db.utils.formataddr instead. Fixes #1378

### DIFF
--- a/alot/completion.py
+++ b/alot/completion.py
@@ -14,6 +14,7 @@ from . import commands
 from .buffers import EnvelopeBuffer
 from .settings.const import settings
 from .utils import argparse as cargparse
+from .db.utils import formataddr
 from .helper import split_commandline
 from .addressbook import AddressbookError
 from .errors import CompletionError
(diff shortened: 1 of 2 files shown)


pazz:
Ah yes, tests fail. I am not sure if we just need to update the tests or if they indicate a required change.
I've just checked that (setting to in envelopes) the promt correctly interprets `double "quote" person <dquote@example.com>`, i.e., keeps the string as is. This is what I'd expect. So the command completer should not artificially add another pair of quotes as indicated by the failing test.

mjg:
Sorry, I should know better than rushing out a PR without running all tests, before leaving the desk... I'll check the tests.

PR #1388, opened 2019-03-28
Ignore non-ascii email-addresses in completion

Tyilo:
Both a bug report and a fix.

**Describe the bug**
I'm using `notmuch` as addressbook, however I have received a mail from the address `André@DigitalOcean` (note the non-ascii character in the address).
This causes alot to crash when trying to complete. 

**Software Versions**
- Python version: 3.7.2
- Notmuch version: 0.28.3
- Alot version: master

**To Reproduce**
Steps to reproduce the behaviour:
1. Create the file `/tmp/custom_abook` with the contents:
```
#!/bin/bash
echo '{"name": "", "address": "André@DigitalOcean", "name-addr": "André@DigitalOcean"},'
```
2. Use the following abook config:
```
		[[[abook]]]
			type = shellcommand
			command = /tmp/custom_abook
			regexp = '\[?{"name": "(?P<name>.*)", "address": "(?P<email>.+)", "name-addr": ".*"}[,\]]?'
			shellcommand_external_filtering = False
```
3. Open alot, write a mail and try to auto-complete the recipient.

**Error Log**
Backtrace:
```
Traceback (most recent call last): 0x7f57c08954c8>)
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/tyilo/repos/alot/alot/__main__.py", line 146, in <module>
    main()
  File "/home/tyilo/repos/alot/alot/__main__.py", line 137, in main
    UI(dbman, cmdstring)
  File "/home/tyilo/repos/alot/alot/ui.py", line 141, in __init__
    self.mainloop.run()
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 286, in run
    self._run()
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 384, in _run
    self.event_loop.run()
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 1340, in run
    reraise(*exc_info)
  File "/usr/lib/python3.7/site-packages/urwid/compat.py", line 58, in reraise
    raise value
  File "/usr/lib/python3.7/site-packages/urwid/main_loop.py", line 1354, in wrapper
    rval = f(*args,**kargs)
  File "/usr/lib/python3.7/site-packages/urwid/raw_display.py", line 404, in <lambda>
[...]

Issue #1409, opened 2019-06-11
Use Python3's email.message.EmailMessage


pazz:
We've moved on to python3 which is great, but we do not yet make use of the new email libs API to create messages. 

This allows to transparently create messages with "attachments" and "html alternatives" without the need to maually fumble sround with MIME parts. See [here](https://docs.python.org/3/library/email.examples.html) for examples.
Switching to message creation using this API should also deal with the weird encoding issues we have for header values (e.g. #1408 and many others).

Taking the long view, I believe that alot should replace the explicit "envelope" class in favour of making envelope mode simple operate on `email.message.EmailMessage`s.

As a first step, it should be fairly easy to replace the ceode in `alot.db.envelope.Envelope` that creates an email message, using the examples from the module docs linked above.

Ideally, we would rely on some external module to create PGP/MIME or S/MIME signatures etc.
Does anyone have experience with python3's `EmailMessage` and PGP/MIME?

dkg:
I don't have any experience with wrapping python3's EmailMessage in PGP/MIME or S/MIME.  i'd start with focusing on PGP/MIME -- but in any case key management seems like the hardest problem, harder than just how you do the wrapping.

I've just recommended that [the gpg python module grow the interface you're looking for](https://dev.gnupg.org/T4578) so that there is one canonical place for doing the right thing with this, and alot wouldn't need to maintain it directly.

As another possible approach without adding too many dependencies, is that, notmuch already uses GMime internally -- and GMime has python bindings via GObject Introspection (see debian package `gir1.2-gmime-3.0`) that might be useful, as they include cryptographic controls.  This is still a lot closer to the wire format than i think you want, though, not nearly at the level of abstraction offered by EmailMessage yet.

dkg:
(the gpg python module seems like a good place to fix things *if* you're using gpg, which i currently assume alot is doing; if the idea is to use something other than gpg, then there may be other steps you want to consider)

dkg:
You  might also take a look at some of the hints [in juliank's recent blog post](https://blog.jak-linux.org/2019/06/13/encrypted-email-storage/) (i don't think you want to do exactly what he's doing there, but it's related and i think he's got the structural parts in decent shape)

pazz:
Thanks for the comments and pointers @dkg!
[...]

PR #1410, opened 2019-06-12
Use db.utils.formataddr instead of email.utils.formataddr

Tyilo:
This should fix the problem described in #1388

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -32,6 +32,7 @@ from ..completion import TagsCompleter
 from ..widgets.utils import DialogBox
 from ..db.errors import DatabaseLockedError
 from ..db.envelope import Envelope
+from ..db.utils import formataddr
 from ..settings.const import settings
 from ..settings.errors import ConfigError, NoMatchingAccount
 from ..utils import argparse as cargparse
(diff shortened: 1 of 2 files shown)


pazz:
We need to double check the tests. See also
https://github.com/pazz/alot/commit/d916b5e42ffd6b5128333321d68365744ca08b1d
for the tests I wrote for the formataddr method.

In the long run, I'm *really* looking forward to using the new python3 email API, which seems to deal much better with encoding issues.

teto:
> In the long run, I'm really looking forward to using the new python3 email API, which seems to deal much better with encoding issues.

what's preventing it ? time ?

pazz:
yes, mostly.

I have no idea yet how to combine the new email API (search for email.message.EmailMessage), which abstracts from manual MIMEMessage building, with pgp/mime. Ideally, someone would write a module that we could use to construct S/MIME or PGP/MIME messages from these EmailMessages. This *could* be easy, but I have not done any research on it.

It should be fairly easy to switch to the new API and break the crypto features in alot, but I assume this will be unpopular. :)

teto:
(sry I removed my post when I saw your topic https://github.com/pazz/alot/issues/1409 :) )

pazz:
can we close this? the issue #1388 is closed already..

Tyilo:
Sure

PR #1413, opened 2019-07-06
Options overwrite default values for sent_tags and draft_tags

### LINKED ISSUE #1412: Cannot overwrite defaults for `sent_tags` and `draft_tags`
**Describe the bug**

If the user sets the config values for `sent_tags` and `draft_tags`, the hardcoded default values `sent` and `draft` are not overwritten, but added to the list of customized tags. This was already mentioned in #918, but has not been pursued further.

**To Reproduce (on the example of `sent_tags`)**
1.  Set `sent_tags` for some account in your config: `sent_tags = my_custom_tag` 
2.  Send a message from this account
3.  The sent message has two tags: `my_custom_tag` and `sent`

**Expected behaviour**
- The sent message has one tag, `my_custom_tag` and default tag `sent` is not added

For me this is unwanted behaviour. But as it exists like this for years now, I'm not sure if I should open a PR. Any comments?

sgelb:
In addition, remove hardcoded default values from account.py and use defaults from alot.rc.spec

This has multiple consequences:
- default values for `sent_tags` and `draft_tags` are solely set in alot.rc.spec
- options `sent_tags` and `draft_tags` in user config do overwrite these defaults. It is now possible to not tag sent mails and drafts at all with `sent_tags = ` and `draft_tags = `
- this is a breaking change if these options were used (though a very small one). To keep the old behaviour, the user has to add `sent` and/or `draft` manually to the options.

This fixes #1412

### DIFF
--- a/alot/account.py
+++ b/alot/account.py
@@ -203,15 +203,6 @@ class Account:
                  abook=None, sign_by_default=False,
                  encrypt_by_default=u"none", encrypt_to_self=None,
                  case_sensitive_username=False, **_):
-        sent_tags = sent_tags or []
-        if 'sent' not in sent_tags:
-            sent_tags.append('sent')
-        draft_tags = draft_tags or []
-        if 'draft' not in draft_tags:
-            draft_tags.append('draft')
-        replied_tags = replied_tags or []
-        passed_tags = passed_tags or []
-
         self.address = Address.from_string(
             address, case_sensitive=case_sensitive_username)
         self.aliases = [


pazz:
Two comments: I'm fine with internal refactoring but the default behaviour really should stay, that sending a mail adds the "send" tag. Perhaps amend the `alot.rc.spec`?
Secondly, let's avoid mutable default parameters (empty lists). This is a well known anti pattern in python.

sgelb:
Thanks for the second comment, another day I learned something!

Just curious, but is there an argument besides backward compability that the "default" tags can't be overwritten even if the user decides to set them explicitly in their configuration?

If the default behaviour should stay, than this PR is useless and can be closed. In this case, a fix for #1412 would be to clarify in the documentaion that the options do not replace a default value, but append additional tags.

pazz:
No: I'm very happy with giving users the option to change the default send-tags. I would just like to keep the default value as "sent", because this is what most people would want and expect (cf notmuch emacs or sup).

lucc:
If I understand @pazz correctly he means that the default behaviour *if the user did not set `*_tags`* in the config, should not change. So if the user does not set `*_tags` it should default to `["send"]` and `["draft"]`.

That is how I understand it and that is also what I would prefer.

alot/account.py
```diff
@@ -198,20 +198,11 @@ class Account:
...
                  realname=None, gpg_key=None, signature=None,
                  signature_filename=None, signature_as_attachment=False,
-                 sent_box=None, sent_tags=None, draft_box=None,
-                 draft_tags=None, replied_tags=None, passed_tags=None,
+                 sent_box=None, sent_tags=[], draft_box=None,
+                 draft_tags=[], replied_tags=[], passed_tags=[],
                  abook=None, sign_by_default=False,
                  encrypt_by_default=u"none", encrypt_to_self=None,
                  case_sensitive_username=False, **_):
-        sent_tags = sent_tags or []
```
lucc: In light of the comment just now I mean that this line could be 
```py
sent_tags = sent_tags or ["send"]
```
and for drafts the corresponding one.

pazz: Wouldn't this overwrite the empty list as parameter?
I mean if a user sets `send_tags = ,` in the ocnfig, (indicating no tags, i.e., the empty list), then `sent_tags or ["send"] ` would default to the latter, which is what @sgelb wanted to change..

And yes, you interpreted my comment above correctly :)

lucc: You are right, this would make it impossible to completely unset the list from the config... Then we have to code it differently.

sgelb:
> If I understand @pazz correctly he means that the default behaviour _if the user did not set `*_tags`_ in the config, should not change. So if the user does not set `*_tags` it should default to `["send"]` and `["draft"]`.

Ah, it seems there was a misunderstanding. That behaviour has not changed.

Hopefully I'm right, but the only place I found where accounts are created is in the method called in https://github.com/pazz/alot/blob/f9689575e771aaa2e6cd4da13b71b88cf8e9e246/alot/settings/manager.py#L127

Here, `self._config` is the result of merging `alot.rc.spec` and the user config. It will contain the default values from `alot.rc.spec` **or** the options from the user config. That means that `*_tags` in the Account constructor are always lists containing either the defaults or the user config options.
So I think the empty lists as default arguments in `account.py` can be replaced with `None` to fix my error. This would be the changed behaviour:

- with no user config options for `*_tags`, the defaults from `alot.rc.spec` are used
- `alot.rc.spec` is the single place of truth for the default values of `*_tags`
- with user config options for `*_tags`, they **overwrite** the defaults. The user can set these options "to nothing" (`sent_tags = ,`), these mails would not be tagged at all.

alot/account.py
pazz: `send_tags = ["send"] if send_tags  is None` ...

sgelb:
I squashed the two commits. No idea why travis fails to update it progress here, it built and tested the new commit 3 days ago.

PR #1422, opened 2019-08-08
Add new hook 'thread_message_text_translate'

rex1fernando:
This adds a hook which can translate the text of each message before it is displayed in thread mode. I personally have found it very useful to strip out the original messages from replies; this makes thread view much more compact and useful, since this way each message widget only has the corresponding message text and not all of the older messages in the thread.

### DIFF
--- a/alot/widgets/thread.py
+++ b/alot/widgets/thread.py
@@ -99,7 +99,12 @@ class TextlinesList(SimpleTree):
                 structure.append((FocusableText(line, attr, attr_focus), None))
         else:
             structure.append((FocusableText(content, attr, attr_focus), None))
+
+        if len(structure) == 0:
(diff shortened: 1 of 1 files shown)


pazz:
Thanks for sending a PR together with this feature request and welcome to alot!
I'd rather have such hook inside the Message object, not the view (widget) on it.
However, I am currently refactoring the "get body text" mechanism there (see #1421)..

rex1fernando:
Thanks for your reply! So should I cancel this pull request and wait until #1421 is merged, and then add the hook to the Message object?

pazz:
Yes, I think that would make sense. Thx

rex1fernando:
Ok, will do :)

PR #1425, opened 2019-08-17
refactor prompt completer code

pazz:
This just splits the file completion.py, which was too large before, into several files.
We should to that with commands as well, but I figured I'd start with the completion code as it does not conflict (much) with open PRs..


pazz:
The last commit addresses the codacy-bot issues; codacity issues can be ignored as they are just copies of the "fixed" ones from the removed `completion.py` file...

alot/completion/command.py
```diff
@@ -0,0 +1,222 @@
...
+# This file is released under the GNU GPL, version 3 or a later revision.
+# For further details see the COPYING file
+
+import logging
+
+from alot import commands
+from alot.buffers import EnvelopeBuffer
+from alot.settings.const import settings
+from alot.utils.cached_property import cached_property
+from . completer import Completer
```
lucc: Why these spaces after the dots?

alot/completion/__init__.py
```diff
@@ -0,0 +1,17 @@
+from .completer import Completer
```
lucc: Why all these imports? They look unused to me.

pazz: yes they are. I just wanted to make sure nothing breaks by making these names available through `alot.completion` as before..

alot/commands/globals.py
```diff
@@ -25,10 +25,10 @@
 from .. import helper
 from ..helper import split_commandstring
 from ..helper import mailto_to_envelope
-from ..completion import CommandLineCompleter
```
lucc: Now I see what that file was about. I like it better the way it is now.

pazz:
cheers for the quick response @lucc !

PR #1426, opened 2019-08-17
inconditionnally add subject

teto:
I have recenlty set ask_subject to false but when entering neovim, I wnated to add a Subject but didn't remember the exact  syntax, ended up writing `Topic:` or `Title:` instead which ofc got ignored.
I think it might be best to add the fields even if empty.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -978,6 +978,7 @@ class ComposeCommand(Command):
             self.envelope.tags = [t for t in self.tags.split(',') if t]
 
     async def _set_subject(self, ui):
+        subject = ""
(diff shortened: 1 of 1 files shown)


pazz:
first there is a typo in your commit msg, but more importantly, i believe that this would invalidate subjects set from command line options via `alot compose --subject`..

teto:
Subject appears twice then. Didn't know about that use case. Will look into
it.

2019年8月18日(日) 2:20 Patrick Totzke <notifications@github.com>:

pazz:
OK, thanks. I will close this PR then. Please just open a new one in case you get around to adress the issue above. Cheers

Issue #898, opened 2016-10-02
Message header ordering isn't preserved from the editor


BlacklightShining:
Okay, so this is _really_ minor. It's also (probably) really easy to change, though! Headers are currently stored internally in a `dict`, which doesn't preserve ordering; Python provides `collections.OrderedDict`, which does. As far as I know, you can just use it as a drop-in replacement.

pazz:
Quoting Blacklight Shining (2016-10-02 15:28:24)
>    Okay, so this is really minor. It's also (probably) really easy to change,
>    though! Headers are currently stored internally in a dict, which doesn't
>    preserve ordering; Python provides collections.OrderedDict, which does. As
>    far as I know, you can just use it as a drop-in replacement.

Yes but why would you want to? The header ordering is completely
irrelevant is it not?

BlacklightShining:
For the most part, yeah, at least before the message goes through MTAs and gets trace headers added to it (which [MUST NOT](https://tools.ietf.org/html/rfc2119) be reordered). Like I said, really minor issue.

pazz:
ok, I'm open for pull requests on this of course :)

pacien:
Standard `dict`s are now officially ordered since Python 3.7.
For some reason, headers still seem to be shuffled when composing a message, even with this version of the Python interpreter.

pazz:
can you be a little more precise as to which two header lists you are comparing? pre/post edit header values listed in envelope buffers?

pacien:
envelope buffer -> `:edit` -> editor: headers are shuffled
editor -> save and quit -> envelope buffer: the order is kept

Headers seem to be randomly re-ordered each time I compose a new message, either with `:compose` or using the command line sub-command:

```
❯ alot compose --sender 1 --to 2 --subject 3
To: 2
From: 1
Subject: 3

❯ alot compose --sender 1 --to 2 --subject 3
Subject: 3
To: 2
From: 1

❯ python --version     
Python 3.7.4

❯ alot --version                            
0.8
```

pazz:
gotya. I don't have a solution for you but I can point to the code that results in the observed behaviour:
https://github.com/pazz/alot/blob/master/alot/commands/envelope.py#L385
and
https://github.com/pazz/alot/blob/master/alot/db/envelope.py#L322.

What happens is that alot will determine, based on your config, which headers will be edited. Those will be removed from the dictionary, turned into part of the template send to the editor, and afterwards re-interpreted and added to the dictionary (at the end).

PR #1428, opened 2019-08-20
commands/envelope: keep editable headers order

pacien:
This should mitigate issue #898 by keeping the order of editable headers from either the configuration or the current envelope.

__This change set hasn't been tested.__

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -27,6 +27,7 @@ from ..helper import string_decode
 from ..settings.const import settings
 from ..settings.errors import NoMatchingAccount
 from ..utils import argparse as cargparse
+from ..utils.collections import OrderedSet
 
 
 MODE = 'envelope'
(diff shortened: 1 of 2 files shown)


pacien:
Travis CI seems to fail due to an external error:

```
Cloning into 'notmuch'...
fatal: unable to look up notmuchmail.org (port 9418) (Temporary failure in name resolution)
```

pazz:
Yes, seems to be down. Give it a few hours, it's late at night in the US, where afaik the people maintaining that site are based.

pazz:
Having had a second look at this I wonder why we don't just use OrderedDict for `envelope.headers` instead?

pacien:
That wouldn't help because the order in the editable file is given by the iteration over `edit_headers`, which order is randomised because of the use of `set`: https://github.com/pazz/alot/blob/9e782988bb2a9383dda8770bf3b0103f69f64c3f/alot/commands/envelope.py#L380

pazz:
Sure. What I was suggesting was to have `envelope.headers` as ordered dict
and in the line you point to instead iterate over *that* ordered dict, filtered by membership in `edit_headers`..

pacien:
Edit headers that are not present in the envelope headers would need to be inserted somewhere. For this, it is reasonable to use the order given in the configuration, which requires the set to be ordered.

pazz:
Can you quickly confirm: you have pushed this commit to several places but it hasn't changed since I've had a look at it right?
Moreover, I'd be great if you could add a more elaborate commit msg, that at least refers to the original issue, and mentions that the whole thing will be obsolete once we move beyond python 3.7 as a requirement.

pacien:
> Can you quickly confirm: you have pushed this commit to several places but it hasn't changed since I've had a look at it right?

Those were to add a `__repr__` method for debugging and fix a typo. No important change otherwise.

> Moreover, I'd be great if you could add a more elaborate commit msg, that at least refers to the original issue, and mentions that the whole thing will be obsolete once we move beyond python 3.7 as a requirement.

Amended to add a comment and a better commit message.

pazz:
cheers

PR #1429, opened 2019-08-29
set headers inconditionnally

teto:
follow up of https://github.com/pazz/alot/pull/1426 but for all fields.
Now tested with `alot compose --subject`

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -386,6 +386,7 @@ class EditCommand(Command):
                 del self.envelope[key]
 
             for value in vlist:
+                logging.debug('key: %s value [%s]', key, value)
                 # newlines (with surrounding spaces) by spaces in values
                 value = value.strip()
                 value = re.sub('[ \t\r\f\v]*\n[ \t\r\f\v]*', ' ', value)
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -769,7 +769,7 @@ class ComposeCommand(Command):
         self.headers = headers or {}
         self.sender = sender
         self.subject = subject
-        self.to = to or []
+        self.to = to
         self.cc = cc or []
         self.bcc = bcc or []
         self.attach = attach
@@ -925,7 +925,8 @@ class ComposeCommand(Command):
 
     async def _set_to(self, ui):
         account = self.envelope.account
-        if 'To' not in self.envelope.headers:
+        # if 'To' not in self.envelope.headers:
+        if self.to is None:
             allbooks = not settings.get('complete_matching_abook_only')
             logging.debug(allbooks)
             abooks = settings.get_addressbooks(order=[account],
@@ -939,7 +940,7 @@ class ComposeCommand(Command):
 
             to = to.strip(' \t\n,')
             ui.recipienthistory.append(to)
-            self.envelope.add('To', to)
+        self.envelope.add('To', to)
 
     async def _set_gpg_encrypt(self, ui):
         account = self.envelope.account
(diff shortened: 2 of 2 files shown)


alot/commands/globals.py
```diff
@@ -724,7 +724,7 @@ class ComposeCommand(Command):
     def __init__(
             self,
             envelope=None, headers=None, template=None, sender=u'',
-            tags=None, subject=u'', to=None, cc=None, bcc=None, attach=None,
+            tags=None, subject=None, to=None, cc=None, bcc=None, attach=None,
```
teto: when using `alot compose`, the default is overriden with None which caused me some trouble so better default to None IMO.

alot/commands/envelope.py
```diff
@@ -386,6 +386,7 @@ def openEnvelopeFromTmpfile():
                 del self.envelope[key]
 
             for value in vlist:
+                logging.debug('key: %s value [%s]', key, value)
```
teto: It helped me see which keys were set to None. I will remove it I think

alot/commands/globals.py
```diff
@@ -966,26 +967,22 @@ def _set_base_attributes(self):
         # set forced headers for separate parameters
         if self.sender:
             self.envelope.add('From', self.sender)
-        if self.subject:
```
teto: to and subject are set in __apply anyway so we don't need these.

teto:
to be clearer, I think this is ready, I just would like to know if the resulting behavior (always printing headers such as cc/bcc) is fine. I will squash/clean afterwards.
I personnally prefer this behavior but as this was not the default so far, maybe it's not the case for everyone :s

alot/commands/globals.py
```diff
@@ -978,14 +978,15 @@ def _set_base_attributes(self):
             self.envelope.tags = [t for t in self.tags.split(',') if t]
 
     async def _set_subject(self, ui):
+        subject = ""
```
pazz: Perhaps this should not be hard-coded but a default taken from the config, which defaults to something non-empty such as "[no subject]"?

alot/commands/envelope.py
pazz: If some of these values are "None" then something went wrong (no pun intended). Setting headers to value "None" makes litte sense and those header should rather be removed IMO

alot/commands/globals.py
```diff
@@ -939,7 +940,7 @@ def _set_gpg_sign(self, ui):
 
             to = to.strip(' \t\n,')
             ui.recipienthistory.append(to)
-            self.envelope.add('To', to)
+        self.envelope.add('To', to)
```
pazz: why is `to` defined?

alot/commands/globals.py
pazz: see above. are they added to non-empty values?

alot/commands/globals.py
```diff
@@ -975,15 +975,14 @@ def _set_base_attributes(self):
...
-                'Subject' not in self.envelope.headers:
+        subject = self.subject
+        if settings.get('ask_subject') and subject is None:
             subject = await ui.prompt('Subject')
             logging.debug('SUBJECT: "%s"', subject)
             if subject is None:
                 raise CommandCanceled()
 
-        self.envelope.add('Subject', subject)
+        self.envelope.add('Subject', subject or u'')
```
pazz: in python3, `u'' == ''` iirc

pazz:
Please rebase and address my questions/change requests if you want this to go into the upcoming release.

pazz:
I plan to make a release soon. Can you comment on the status of this?

teto:
You can ignore it for now. 

NB. nice hat !

pazz:
ok. thanks :-)

pazz:
How do we stand on this? If you want to withdraw the PR then please just close it. But I'm not against merging it in case it addresses my minor comments above..

PR #1432, opened 2019-09-06
Fix html only mails

mjg:
Make HTML-only mails display again.
(No, they'll never be "great".)

### DIFF
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -477,7 +477,7 @@ def extract_body(mail):
     """
 
     if settings.get('prefer_plaintext'):
-        preferencelist = ('plain',)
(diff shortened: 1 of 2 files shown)


pazz:
It looks good to me save for the failing test (which seems due to a typo in the expected text?).

mjg:
> It looks good to me save for the failing test (which seems due to a typo in the expected text?).

Hmm, the tests do succeed here - I tested the tests :)

BUT: None of the two results are correct. We expect the HTML part if that is the only one, and the test comparisons where messed by the default html mailcap handler which is different for travis and for me. I have to apply the mailcap decorator (cat).

pazz:
thanks!

PR #1436, opened 2019-09-26
Bump twisted to >= 18.4.0 fix #1420

### LINKED ISSUE #1420: Send mail and initial_command broken with alot-0.8.1 and <twisted-18.4.0
Hello !

**Describe the bug**
Last week I had to change my laptop ssd and reinstall partly my system,
then I stumble upon #1372 and #1373 again, I was sure this comes from one of the dependencies not right, because I thought it was working fine on my last ssd.
I am sure the problem stopped occuring because I upgrade one of the dependencies in my userspace, I only have search which one.

There might be other problems with this issue, but I noticied those:
* The initial_command did not work (event the default one).
* The send mail was stuck on Sending, but I can see in the msmtp log that it was gone well.

**Software Versions**
- Python version: 3.6.5
- Notmuch version: 0.28.4
- Alot version: alot-0.8.1 (reproducible on master)
- msmtp 1.8.5

So I try different version of twisted and there is no more issue on the 18.4.0, so I will bump the ebuild in gentoo to this version at least, but I wanted to share the issue with you all.

And so maybe bumping the setup.py install_requires array to this version would prevent the issue for other ?

GuillaumeSeren:
Hey,
Following the issue #1420 I have bump twisted to be at least 18.4.0,
this patch need some checking to be sure everyone can have this version.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -28,7 +28,7 @@ A full list of dependencies is below:
 * `urwid <http://excess.org/urwid/>`_ toolkit, ≥ `1.3.0`
 * `urwidtrees <https://github.com/pazz/urwidtrees>`_, ≥ `1.0`
 * `gpg <http://www.gnupg.org/related_software/gpgme>`_ and it's python bindings, ≥ `1.9.0`
-* `twisted <https://twistedmatrix.com>`_, ≥ `10.2.0`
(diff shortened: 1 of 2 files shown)


lucc:
There is a nice overview for such questions at repology.org: https://repology.org/project/python:twisted/versions

In my opinion it looks good.

Issue #746, opened 2015-01-08
Color-code quoted text in message bodies


elehack:
It would be very nice to have quoted text color-coded like it is in Mutt. This seems related to #639, perhaps addressable by the same underlying technical solution?

Issue #302, opened 2012-02-03
asyncronous calls in AbooksCompleter


pazz:
abook.lookup really.
cf issue #294

PR #1441, opened 2019-10-10
[RFC] Colour quoted text

Nelyah:
## Context 

I feel like being able to render quote text according to the quoted level with colours is important. At least it is for me in terms of readability.

This PR is to answer this particular feature. An issue has been opened #746, but it was answered with #612 and then #1015. These are great features, but don't quite target this specific issue.  

## Modifications

I modified the file `alot/widget/thread.py` to call the `parse_text_colour` function, if we are looking at words linewise. The function `parse_text_colour` asks the config to check if we want to parse quoted text. If yes, it proceeds and calls the function `parse_quotes`. 

The latter returns the theme value associated with the quote_level in the theme file (if there is one). 

## Example configuration

This is working well for me and it passes the tests. This is how I set it up in my config file to make either `>` or four spaces a quote symbol.:

    quote_symbol = '( {4}|>)'
    parse_quotes = True

I am fairly new to this project, please let me know if there are obvious things I am missing.


Nelyah:
Thank you @pazz, this is a wonderful piece of software you got there!

Here is a screenshot of a sample mail. I set up the colours being close from one another, but this can be changed with the options `quote_level_i` in the configuration file (where i is the quotation level wished for). I declared those only up to i=7. Let me know if you'd wish more colour contrast.

Screenshot from 2019-10-10 18-17-58


> Would this only change the urwid representation or does your code changes the text of follow-up reply emails?

No, this should only change the urwid representation. The only real modification to existing code is to change the `attr` theming value (in `alot/widget/thread.py`). I am getting the new `attr` value by parsing the mail body and searching it for quotation markers (which can be defined in the configuration file)

alot/widgets/colour_text.py
```diff
@@ -0,0 +1,55 @@
...
+    :param str line: The line of text to be parsed
+    :param int max_quote_level: Search for quotes up to that level
+    :return: Tuple (is_quoted, quote_colour)
+    """
+    symbol = settings.get('quote_symbol')
+    is_quoted = False
+    quote_colour = None
+
+    for quote_level in range(1, max_quote_level+1):
+        quote_regex = r'^\ *(' + symbol + '\ *){' + str(quote_level) + r'}'
```
lucc: Why are there backslashes before whitespace? Why are these raw strings? Please use `str.format` instead of `str+str`.

alot/widgets/colour_text.py
```diff
@@ -0,0 +1,55 @@
...
+    """Return colour attribute as (attr, attr_focus)
+
+    :param str line: line of text to be parsed
+    :return: The theme attribute to apply
+    """
+    if settings.get('parse_quotes'):
+        return parse_quotes(line)
+
+
+def parse_quotes(line, max_quote_level=7):
```
lucc: Maybe it is better to move the 7 into a local variable. The argument is not changed from anywhere (if I see that correctly). Please also document that there is the magic number 7 here because there are 7 levels predefined in the config spec. I think the code would work if a user defines a higher level in their config that is not in the spec. But I don't think we want to support this case.

Please fix the newline in the docstring. It should be between the text and the parameters not the params and the return.

alot/widgets/colour_text.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    is_quoted, quote_colour = get_quote_colour(line, max_quote_level)
+
+    if is_quoted and quote_colour is not None:
+        return quote_colour
+    else:
+        return None
+
+
+def get_quote_colour(line, max_quote_level):
```
lucc: It would be good to provide some tests for this code.

alot/widgets/colour_text.py
Nelyah: You're right, I pushed a commit to use `str.format` instead of `str+str`.

As far as I know however, it is good practice to use raw strings when writing regexes (which may contain special characters). It is not yet the case in this one but I prefer to do it consistently.

I escape the space characters to make them more visible and improve the regex readability. Should I delete them? They are not otherwise necessary.

alot/widgets/colour_text.py
Nelyah: I pushed another commit to move the 7 to a local variable. I explain in the docstring up to what level we go and in a comment why is that.

From what I have tested, if a user defines `quote_level_8` (`>7`) in their theming config, that attribute is not considered. When a line is quoted, my code will match the highest possible quote level up to 7, then apply the theming attribute of that highest quote level found.

For instance:

    This is not a quoted line
    >>>>>>> This is 7th quote level
    >>>>>>>> This is 8th quote level
    >>>>>>>>> This is 9th quote level

Except for the not quoted line, they will all match the theming attribute of the 7th quote level. If no such attribute was defined in the theming config or if the line is not quoted, then the return value of `settings.get_theming_attribute` is `'default'`. If such a value is returned, we use the `theme.body` theming attribute instead.

alot/widgets/colour_text.py
Nelyah: Ok! I'll work on it.

Nelyah:
I just added the tests as well

alot/widgets/colour_text.py
lucc: I think escaping stuff that doesn't need escaping does not make the code more readable. It just makes me wonder weather I or the author did not understand (the) regex(es) .

tests/widgets/test_colour_text.py
```diff
@@ -0,0 +1,110 @@
...
+
+class TestQuoteLevel(unittest.TestCase):
+
+    @mock.patch(
+        'alot.commands.globals.settings._config', SETTINGS)
+    @mock.patch(
+        'alot.commands.globals.settings._theme',
+        theme.Theme(DUMMY_THEME.splitlines()))
+    def test_quote_level_get_equals_corresponding_attribute(self):
+        self.assertEqual(settings.get('quote_symbol'), '>')
```
lucc: We don't have to test `settings.get` here.

tests/widgets/test_colour_text.py
```diff
@@ -0,0 +1,110 @@
...
+    line_focus = '', '', '', '', '', ''
+    line_odd = '', '', '', '', '', ''
+[search]
+    focus = '', '', '', '', '', ''
+    normal = '', '', '', '', '', ''
+    [[threadline]]
+        focus = '', '', '', '', '', ''
+        normal = '', '', '', '', '', ''
+[thread]
+    quote_level_2 = '', '', '', '', '#880', '#886'
```
lucc: Please use different colors for each value so that no false positive can happen because of color code collision.  I think best would be "recognizable" codes: `#201` and `#202` for level 2, `#301` and `#302` for level 3 ...

tests/widgets/test_colour_text.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @mock.patch(
+        'alot.commands.globals.settings._config', SETTINGS)
+    @mock.patch(
+        'alot.commands.globals.settings._theme',
+        theme.Theme(DUMMY_THEME.splitlines()))
+    def test_quote_level_get_equals_corresponding_attribute(self):
+        self.assertEqual(settings.get('quote_symbol'), '>')
+
+        self.assertIsNone(parse_quotes(''))
```
lucc: I would put these None cases in one test method of their own.

tests/widgets/test_colour_text.py
```diff
@@ -0,0 +1,110 @@
...
+        self.assertIsNone(parse_quotes('test no quote'))
+
+        # Undefined value
+        self.assertIsNone(parse_quotes('>'))
```
lucc: Why is this None?  I was expecting it to be the color for level 1. But that is not defined. So maybe this is the default value from the spec?  In this case I would make one extra test method that tests default values.

tests/widgets/test_colour_text.py
```diff
@@ -0,0 +1,110 @@
...
+        self.assertEqual(parse_quotes('>> >> >> >>'), AttrSpec('#180', '#186'))
+
+    @mock.patch(
+        'alot.commands.globals.settings._config', SETTINGS_test_symbol_regex)
+    @mock.patch(
+        'alot.commands.globals.settings._theme', theme.Theme(DUMMY_THEME.splitlines()))
+    def test_regex_symbol_is_correct(self):
+        # Match 4th level quote
+        self.assertEqual(parse_quotes('Aaa A'), AttrSpec('#580', '#586'))
+        self.assertEqual(parse_quotes('Aa<>tb A'), AttrSpec('#580', '#586'))
```
lucc: I would also add tests for stuff like `some line with >> marker in the middle`, `some line with the marker at the end: >>` .

tests/widgets/test_colour_text.py
Nelyah: Fixed in the last push.

tests/widgets/test_colour_text.py
Nelyah: Fixed in the last push. I also moved the theme variable as global to my script as to avoid reading it for each test case.

tests/widgets/test_colour_text.py
Nelyah: Done. I put together default values for quotes and non quotes in the same method as they return `None`.

tests/widgets/test_colour_text.py
Nelyah: When an undefined `quote_level` theming attribute is requested but not found, `'default'` is returned by `settings.get_theming_attribute`.

The return value from `get_quote_colour` is either `None` if no quote were found, `'default'` if a quote was found but the theming was undefined or an `AttrSpec` instance if this is a quote but its theming value is defined.

In the first two cases, `parse_quotes` will return `None` and this is wanted.   
>  In this case I would make one extra test method that tests default values.

I added on such test case.

tests/widgets/test_colour_text.py
Nelyah: Done.

alot/widgets/colour_text.py
Nelyah: Fair enough. I fixed it in the last push.

Nelyah:
Thank you for the feedback, @lucc! And sorry if this has taken some time and back and forth.

alot/widgets/colour_text.py
```diff
@@ -0,0 +1,42 @@
+import logging
```
pazz: please add a licence header as in all other files. You are welcome to add your name in there of course

alot/widgets/colour_text.py
```diff
@@ -0,0 +1,42 @@
...
+    quote_colour = None
+    symbol = settings.get('quote_symbol')
+    if symbol is None:
+        symbol = r'>'
+    for quote_level in range(1, max_quote_level+1):
+        quote_regex = r'^\ *(' + symbol + '\ *){' + str(quote_level) + r'}'
+
+        if re.match(quote_regex, line):
+            is_quoted = True
+            logging.info(f'Requesting attribute quote_level_{quote_level}')
```
pazz: should be `debug`, not `info` imo

alot/widgets/colour_text.py
```diff
@@ -0,0 +1,42 @@
...
+    """Search for quotes. 
+
+    :returns: attr, None if no quote are available
+
+    """
+    default_colour = settings.get_theming_attribute('thread', 'body')
+    is_quoted = False
+    quote_colour = None
+    symbol = settings.get('quote_symbol')
+    if symbol is None:
```
pazz: use the fallback option when getting the config setting? https://github.com/pazz/alot/blob/master/alot/settings/manager.py#L239
(note that the defaul is "> ", with a trailing space)

alot/widgets/colour_text.py
```diff
@@ -9,7 +9,8 @@ def parse_text_colour(line):
     :returns: attr
 
     """
-    return parse_quotes(line)
+    if settings.get('parse_quotes'):
```
pazz: otherwise return `None` (implicitly)?

alot/defaults/alot.rc.spec
```diff
@@ -17,6 +17,10 @@ input_timeout = float(default=1.0)
 # .. note:: this config setting is equivalent to, but independent of, the 'search.exclude_tags' in the notmuch config.
 exclude_tags = force_list(default=list())
 
+# Show quotes be parsed
+parse_quotes = boolean(default=True)
+quote_symbol = string(default='>')
```
pazz: why not use the existing `quote_prefix`?

Nelyah:
I'll do the licence change and add the quote_prefix (didn't see it previously). 
However squashing commits (which I have already pushed to origin) through a rebase means I am rewriting all of my commits as my new master diverges from origin/master, isn't it? 
I'm not quite a git professional, but let me know if this is what I should do (unless there is a better solution for it)

lucc:
You can do an interactive rebase. Like this `git rebase --interactive origin/master message-body-colour`. Learn more about it [from the docs](https://help.github.com/en/github/using-git/about-git-rebase).

pazz:
Where are we on this? It'd be a shame to let this get too old and diverge from master too much to be useful..

Nelyah:
Oh, I'm sorry. 

I had some things happening at the time and it just slipped out of my mind. I will get back to this in the next few days! Sorry again for the delay...

Nelyah:
I rebased and resolved a few merge conflicts (mainly in `thread.py` with `FocusableText` being replaced by `ANSIText`), but have not yet added anything else.

I added a few `[tags]` to my commits to help myself (and maybe you as well) since it's been a while I worked on this. I'll remove them before merging this.

alot/defaults/alot.rc.spec
Nelyah: I didn't see it before!  

However, I notice that it has a space added to it. That would mean that a line like:

```
>> Level 2 quote level
```

wouldn't get interpreted correctly, because the regex `r'^ *({} *){{{}}}'.format(symbol, quote_level)` wouldn't match. It would still be possible to change it but the default wouldn't work really well. 

I'm not sure in what context  `quote_prefix` is used. A first solution would be to remove the space and accommodate for this wherever it is used. A second could be to rename my `quote_symbol` to something more explicit, like `quotation_regex_detection`.

alot/widgets/colour_text.py
Nelyah: Does it  make sense here to specify a fallback? If I understand correctly we would still get the default settings value of the variable we're requesting (in this case from `quote_symbol` : `>`), even if no fallback is defined.

tests/widgets/test_colour_text.py
```diff
@@ -0,0 +1,119 @@
+# Copyright (C) 2017  Lucas Hoffmann
+# This file is released under the GNU GPL, version 3 or a later revision.
+# For further details see the COPYING file
+
+import unittest
+from urwid import AttrSpec
+from unittest import mock
+
+from alot.settings.const import settings
```
pazz, line 9: Codacy Issue found: [Unused settings imported from alot.settings.const](https://app.codacy.com/manual/patricktotzke/alot/pullRequest?prid=4429219)

alot/widgets/colour_text.py
Nelyah: Just added it

tests/widgets/test_colour_text.py
Nelyah: Removed it in 32cebe8

alot/defaults/alot.rc.spec
lenormf: I don't think this new option is necessary. The two most commonly used characters for quoting text are `>` and `|`, and I've only ever seen the bracket being used ([wikipedia reference](https://en.wikipedia.org/wiki/Posting_style#Quoted_line_prefix)).

It would make the PR leaner to assume only those two characters highlight quotes in messages, plus it would avoid some weird unit tests verifying that the code behaves as expected when using `[Aa]` for `quote_symbol`.

alot/widgets/colour_text.py
```diff
@@ -0,0 +1,65 @@
...
+    """Cycle through quotation levels for the line
+
+    :param str line: The line of text to be parsed
+    :param int max_quote_level: Search for quotes up to that level
+    :return: quote_colour (either string 'default' or an AttrSpec object)
+    """
+    symbol = settings.get('quote_symbol')
+    quote_colour = None
+
+    for quote_level in range(1, max_quote_level+1):
```
lenormf, line 53: I think running 7 regex matches on a line that has 7 levels of nested quoting is too much.

I made it work (with a dirty patch) on my local branch with only two calls to `re.match()`, for any amount of nesting, I think it's the better way to go about it.

Plus if we remove the `quote_symbol` option (c.f. my other comment), it will be easier to parse spaces properly ([RFC reference](https://tools.ietf.org/html/rfc3676#section-4.5))

tests/widgets/test_colour_text.py
```diff
@@ -0,0 +1,118 @@
...
+        odd = '', '', '', '', '', ''
+"""
+
+DUMMY_THEME = theme.Theme(DUMMY_THEME_STRING.splitlines())
+
+SETTINGS = {'quote_symbol': '>',
+            'colourmode': 256
+            }
+
+SETTINGS_test_symbol_regex = {'quote_symbol': '([Aa]|<>.b|>)',
```
lenormf, line 76: As mentioned in another comment, this is a weird thing to test for. Maybe we don't need the code to be that generic.

alot/widgets/colour_text.py
```diff
@@ -0,0 +1,65 @@
...
+        quote_regex = r'^ *({} *){{{}}}'.format(symbol, quote_level)
+        if re.match(quote_regex, line):
+            logging.debug(
+                'Requesting attribute quote_level_{}'.format(quote_level))
+            quote_colour = settings.get_theming_attribute(
+                'thread', 'quote_level_{}'.format(quote_level))
+
+        else:
+            # If there is no match at some point,
+            # we simply use the last level match colour
```
lenormf, line 63: I think cycling back to the colours for the first level of quoting is a better option:

* 8 → 1
* 9 → 2
* …
* 15 → 1
* 16 → 2

etc.

Basically a modulo.

PR #1444, opened 2019-10-28
Implement command for confirming command sequences.

### LINKED ISSUE #1368: Confirmation option for command pipelines
I'd like to bind sequences of --all tagging operations in search buffers (so that I can easily, say, tag all e-mails from a search query as deleted, and untag the inbox and unread tags). This is currently a bit risky because even with a leader like 'g delete' I might accidentally hit that at some point.

I was thinking of the possibility of having a "confirm" command as the first command in a sequence which prompts the user similar to bug_on_exit. If answered "n", it aborts the entire sequence.

Alternatives could be to add a --confirm or --prompt option to tagging operations but then the sequence would just continue when answered "n" on the first.

I'm also considering just binding 'g delete y' to have effectively the same sequence of key presses, but that doesn't actually *prompt* me for what I'm about to do and seems to have some trouble with the timing of the third keypress.

MacGyverNL:
First attempt at implementing a confirmation command intended for use in
command sequences, as discussed in #1368.

When called (optionally with a free-form string message), prompts the
user for confirmation. If the user declines, a SequenceCanceled
exception is raised, which trickles up the error handling into
apply_commandline where it is explicitly caught outside the common error
handler. This aborts the execution of subsequent commands in the
commandline, but does not show up as an error in the log.

Closes #1368

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -17,7 +17,7 @@ import shlex
 import urwid
 
 from . import Command, registerCommand
-from . import CommandCanceled
+from . import CommandCanceled, SequenceCanceled
 from .utils import update_keys
 from .. import commands
 
@@ -1174,3 +1174,30 @@ class RemoveQueryCommand(Command):
         # flush index
         if self.flush:
             ui.apply_command(commands.globals.FlushCommand())
+
+
+@registerCommand(
+    MODE, 'confirmsequence',
+    arguments=[
+        (['msg'], {'help': 'Additional message to prompt',
+                   'nargs': '*'})
+    ],
+    help="prompt to confirm a sequence of commands")
+class ConfirmCommand(Command):
+    """Prompt user to confirm a sequence of commands."""
+
+    def __init__(self, msg=None, **kwargs):
+        """
+        :param msg: Additional message to prompt the user with
+        :type msg: List[str]
+        """
+        super(ConfirmCommand, self).__init__(**kwargs)
+        if not msg:
+            self.msg = "Confirm sequence?"
+        else:
+            self.msg = "Confirm sequence: {}?".format(" ".join(msg))
+
+    async def apply(self, ui):
+        if (await ui.choice(self.msg, select='yes', cancel='no',
+                            msg_position='left')) == 'no':
+            raise SequenceCanceled()
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -16,7 +16,7 @@ from .buffers import BufferlistBuffer
 from .buffers import SearchBuffer
 from .commands import globals
 from .commands import commandfactory
-from .commands import CommandCanceled
+from .commands import CommandCanceled, SequenceCanceled
 from .commands import CommandParseError
 from .helper import split_commandline
 from .helper import string_decode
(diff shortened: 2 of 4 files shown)


MacGyverNL:
FYI I clearly have no idea what I'm doing with the documentation. Maybe it's me, maybe it's Travis-CI, but help getting that sorted would be appreciated.

alot/commands/globals.py
```diff
@@ -1174,3 +1174,30 @@ def apply(self, ui):
         # flush index
         if self.flush:
             ui.apply_command(commands.globals.FlushCommand())
+
+
+@registerCommand(
+    MODE, 'confirmsequence',
+    arguments=[
+        (['msg'], {'help': 'Additional message to prompt',
```
pazz: additional sounds as if it was optional, but it shouldn't be right?

alot/ui.py
```diff
@@ -145,6 +145,10 @@ def _error_handler(self, exception):
             self.notify(str(exception), priority='error')
         elif isinstance(exception, CommandCanceled):
             self.notify("operation cancelled", priority='error')
+        elif isinstance(exception, SequenceCanceled):
```
pazz: just to make sure I understand: Without introducing this new exception, and raising `CommandCanceled` instead, would only abort the current Command but not the whole sequence?

alot/commands/globals.py
MacGyverNL: It is, though -- there's a defaut message that doesn't require you to give additional information about what's being prompted. But if it should be mandatory, I'd be happy to change.

pazz: I guess that would make sense, don't you think?

alot/ui.py
MacGyverNL: Yes. That's what I tried initially, but CommandCanceled gets handled in self.apply_command and not re-raised, so never trickles up to self.apply_commandline, so a normal CommandCanceled in a sequence doesn't cancel the sequence.

MacGyverNL:
> Perhaps we should also add tests?

Probably, but I'm not sure what kind of test this would even be. Executing this in a pipeline and then checking whether the second command also gets run?

alot/commands/globals.py
MacGyverNL: The default message is just "Confirm sequence?"; I didn't see the need for making any additional prompt necessary. Maybe the entire prompt should just be passed in as a mandatory argument?

alot/ui.py
pazz: Ok thanks for clarifying.
IIRC, CommandCanceled was there to enable confirmation prompts inside a single command, such as when sending a message.
I'm not sure if we really need two different types of cancellation exceptions. I mean, it seems to me that all other use cases would be fine if the whole *sequence* of commands is killed. This would only be a problem if you chain commands with prompts, and *really* want to continue despite having negatively answered some prompt. useful?

MacGyverNL: I agree that that doesn't sound very useful. We could take the CommandCanceled handling out of the unified exception handler function and handle it differently in the exception handlers of apply_command and apply_commandline. I didn't want to mess too much with the unified exception handling function when initially creating this.

pazz:
I think this is useful as is and we can always tweak the default message and add tests later.

PR #1447, opened 2019-11-04
Remove unicode literals syntax from python2

lucc:
In order to confuse new developers less.

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -116,7 +116,7 @@ def string_decode(string, enc='ascii'):
 def shorten(string, maxlen):
     """shortens string if longer than maxlen, appending ellipsis"""
     if 1 < maxlen < len(string):
-        string = string[:maxlen - 1] + u'…'
+        string = string[:maxlen - 1] + '…'
     return string[:maxlen]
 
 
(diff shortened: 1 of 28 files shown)


alot/helper.py
```diff
@@ -179,7 +179,7 @@ def shorten_author_string(authors_string, maxlength):
         au = authors.pop()
         if len(au) > 1 and (remaining_length == 3 or (authors and
                                                       remaining_length < 7)):
-            authors_chain.appendleft(u'\u2026')
```
pazz: is this still necessary then? I presume that source files are utf-8 encoded?

tests/commands/test_global.py
```diff
@@ -84,12 +84,12 @@ def test_set_gpg_sign_by_default_but_no_key(self):
...
 
     def test_get_template_decode(self):
-        subject = u'This is a täßϑ subject.'
-        to = u'recipient@mail.com'
-        _from = u'foo.bar@mail.fr'
-        body = u'Body\n地初店会継思識棋御招告外児山望掲領環。\n€mail body €nd.'
+        subject = 'This is a täßϑ subject.'
+        to = 'recipient@mail.com'
+        _from = 'foo.bar@mail.fr'
+        body = 'Body\n地初店会継思識棋御招告外児山望掲領環。\n€mail body €nd.'
```
pazz: compare this with the comment above, where we still use escape codes

lucc:
I only found 3 cases.

pazz:
Thanks for this!

PR #1448, opened 2019-11-04
Test on Python 3.8

polyzen:
Edit: We are trying to move to Python 3.8 at Arch Linux:
https://rebuilds.foutrelis.com/log/alot.log


pazz:
Good to know that it doesn't work out of the box.
The error log on Travis seems to suggest that the unittest module has changed.. feel free to add a pr to fix whatever needs fixing for moving the tests, and alot, to the new version.

Issue #1437, opened 2019-09-26
RFC2822 2.2.3. (folded) Long Header Fields


pacien:
**Describe the bug**

Long header fields that span over multiple lines as defined in section 2.2.3. of RFC2822 aren't correctly parsed by alot.

A multi-line header and the ones following are recognised as part of the message body instead of being headers when going back from the editor to the envelope view.

**Software Versions**

- Python version: 3.7.4
- Notmuch version: [if relevant]
- Alot version: 0.8

**To Reproduce**
Steps to reproduce the behaviour:
1. compose a new message, open the text editor
2. prepend a multi-line header like
```
Cc: a@a.a,
    b@b.b
```
3. save and exit the editor, go back to the envelope view
4. observe that this header and the ones following it aren't seen as headers but instead as part of the body

**Error Log**
N.A.

pazz:
Yes, this is because headers, as edited manually, are not parsed by the email module but instead [here](https://github.com/pazz/alot/blob/master/alot/db/envelope.py#L294]. 
This is to be able to read utf-8 encoded strings which according to the RFC would have to be encoded in ascii, see also here for comments: https://github.com/pazz/alot/blob/master/alot/commands/envelope.py#L353

I agree that this is not ideal. Feel free to suggest improvements! Perhaps we should simply use pythons email module to parse user edited messages as well, but I'm not sure how that reacts to accidentally malformed values that include e.g. umaluts but are not correctly escaped.

PR #1453, opened 2019-11-15
envelope: correctly handle folded headers

pacien:
Should fix https://github.com/pazz/alot/issues/1437.

### DIFF
--- a/alot/db/envelope.py
+++ b/alot/db/envelope.py
@@ -291,50 +291,42 @@ class Envelope:
 
         return outer_msg
 
-    def parse_template(self, tmp, reset=False, only_body=False):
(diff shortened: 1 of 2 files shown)


pacien:
The unit test is nice to have. I don't have any Windows machine to test with CRLF but it should work.

pazz:
in any case, It'd be good to have a test for interpreting multi-line headers as fixed in this PR.

pacien:
I'm confused. There is already a unit test for that included in this PR. Do we need one testing CRLF too? It should work the same as LF is contained in it.

pazz:
Yes, I've overlooked that. cheers.

PR #1457, opened 2019-11-23
envelope: add setting for custom Message-ID domain

pacien:
This allows users to use a custom domain in automatically generated
Message-ID headers instead of the local hostname.

### DIFF
--- a/alot/db/envelope.py
+++ b/alot/db/envelope.py
@@ -274,7 +274,8 @@ class Envelope:
         headers = self.headers.copy()
         # add Message-ID
         if 'Message-ID' not in headers:
-            headers['Message-ID'] = [email.utils.make_msgid()]
+            domain = settings.get('message_id_domain')
+            headers['Message-ID'] = [email.utils.make_msgid(domain=domain)]
 
         if 'User-Agent' in headers:
             uastring_format = headers['User-Agent'][0]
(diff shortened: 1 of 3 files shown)


pacien:
Amended to regenerate docs.

mjg:
Wow, that was fast.

I'm not sure this PR is the right implementation for a good idea. It indeed is necessary to keep local hostnames from leaking. This is possible already with a hook like this one:
```
async def pre_envelope_send(ui, dbm, cmd):
    e = ui.current_buffer.envelope
    if 'Message-ID' not in e.headers:
        address = e.get('Resent-From', False) or e.get('From', '')
        _, address = email.utils.parseaddr(address)
        e.headers['Message-ID'] = [email.utils.make_msgid(*address.split('@'))]
```
In most cases, it will be necessary to specify the domain per account. In many cases, it will be the domain of the account's address; some will like more flexibility. But I don't think a global setting would serve many use cases.

pazz:
@mjg args, yes you are right. I did not think about multiple accounts.
I believe that a native config option (instead of a hook) is nice, but perhaps it should be moved into the accounts section. 

Sorry for my trigger-happyness. PRs welcome of course, and I'd make a fix a priority before tagging the next release.

pazz:
@mjg: OK with the fix?

mjg:
@pazz Thanks. I have not checked yet whether the hook would still work. Note that accounts can have several addresses that you send from (aliases) which are not necessarily on the same domain. People use gmail or similar to send "from" their project alias (fedoraproject, debian,...). I guess I would still prefer the hook for those (or an implementation which uses the domain of the from-address, maybe based on a boolean setting).

pazz:
Your hook should still work because it overwrites the value just before sending.

I think catering for multiple domains for the same account is overkill and people can fall back to your hook if the message_id domain really is that important to them :-)

Issue #1454, opened 2019-11-18
Can't attach a file if filename contains a semicolon


mmartin:
**Describe the bug**
Can't add an attachment if the filename contains semicolon.

**Software Versions**
- Python version: 3.6.9
- Alot version: 0.8.1

**To Reproduce**
Steps to reproduce the behaviour:
1. compose new e-mail
2. try :attach /path/to/file;name
3. error: unknown command: name
4. try :attach '/path/to/file;name'
5. error:  No closing quotation
6. try :attach "/path/to/file;name"
7. error: No closing quotation

pazz:
Right. That sucks. I believe that [this](https://github.com/pazz/alot/blob/master/alot/helper.py#L26) is the problem.

As a hotfix, try using wildcards: `:attach /path/to/file*name`..

mmartin:
Yes, that was the problem. I wrote a custom parser. It's not pretty or efficient, but works good enough for me.

But I found an another problem – tab-completion does not work – `:attach '/<Tab>` returns nothing.

```python
def split_commandline(s):
    """
    splits semi-colon separated commandlines
    """
    commands = []
    buffer = ''
    state = 0 # 0 - default; 1 - escape; 2 - inside single qute; 3 - inside double quote
    prev_state = 0 # for handling escaped symbols

    for c in s:
        if state == 0:
            if c == ';':
                commands.append(buffer.strip())
                buffer = ''
                continue
            elif c == '\\':
                prev_state = state
                state = 1
            elif c == '\'':
                state = 2
            elif c == '"':
                state = 3
            buffer += c

        elif state == 1:
            state = prev_state
            buffer += c

        elif state == 2:
            if c == '\'':
                state = 0
            elif c == '\\':
                prev_state = state
                state = 1
            buffer += c

        elif state == 3:
            if c == '"':
                state = 0
            elif c == '\\':
                prev_state = state
                state = 1
            buffer += c

    if buffer:
        commands.append(buffer.strip())

    return commands
```

PR #1458, opened 2019-11-29
ignore quoted separators when splitting command lines

pacien:
Fixes https://github.com/pazz/alot/issues/1454:
Can't attach a file if filename contains a semicolon

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -23,20 +23,12 @@ import urwid
 import magic
 
 
-def split_commandline(s, comments=False, posix=True):
+def split_commandline(s):
     """
-    splits semi-colon separated commandlines
+    splits semi-colon separated commandlines, ignoring quoted separators
     """
-    # shlex seems to remove unescaped quotes and backslashes
-    s = s.replace('\\', '\\\\')
-    s = s.replace('\'', '\\\'')
-    s = s.replace('\"', '\\\"')
-    lex = shlex.shlex(s, posix=posix)
-    lex.whitespace_split = True
-    lex.whitespace = ';'
-    if not comments:
-        lex.commenters = ''
-    return list(lex)
+    splitter = r'''((?:[^;"']|"(\\\\|\\"|[^"])*"|'(\\\\|\\'|[^'])*')+)'''
+    return re.split(splitter, s)[1::4]
 
 
 def split_commandstring(cmdstring):
--- a/tests/test_helper.py
+++ b/tests/test_helper.py
@@ -124,9 +124,20 @@ class TestSplitCommandline(unittest.TestCase):
         expected = [base]
         self._test(base, expected)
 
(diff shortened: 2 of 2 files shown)


mmartin:
This seems not to work correctly, if there's an escaped quote inside the quotes.

This:

```echo "foo; \"bar; baz";sleep 1```

gets split into:

```
echo "foo; \"bar
 baz
sleep 1
```

**Comment by pacien** [comment]
I fixed the regexp and added a test for quoted escaped quotes.

File: alot/helper.py [hunk 1]
```diff
@@ -23,20 +23,12 @@
...
-    s = s.replace('\\', '\\\\')
-    s = s.replace('\'', '\\\'')
-    s = s.replace('\"', '\\\"')
-    lex = shlex.shlex(s, posix=posix)
-    lex.whitespace_split = True
-    lex.whitespace = ';'
-    if not comments:
-        lex.commenters = ''
-    return list(lex)
+    splitter = r'''((?:[^;"']|"(\\\\|\\"|[^"])*"|'(\\\\|\\'|[^'])*')+)'''
```
roosemberth, line 30: >     splitter = r'''((?:[^;"']|"(\\\\|\\"|[^"])*"|'(\\\\|\\'|[^'])*')+)''' 
>                                                                     ^

Why do you match one or more groups? (`+`). This will not recognize the string `''`.

tests/test_helper.py
```diff
@@ -124,9 +124,16 @@ def test_single(self):
         expected = [base]
         self._test(base, expected)
 
-    def test_unicode(self):
-        base = 'echo "foo";sleep 1'
-        expected = ['echo "foo"', 'sleep 1']
+    def test_quoted_separator(self):
+        base = '''echo "foo; bar";sleep 1 ; echo "foo; \\"bar; baz";''' \
+               ''' echo "foo; bar \\\\" "baz"; test 'hi' 'hi' '''
```
roosemberth, line 129:
```suggestion
               ''' echo "foo; bar \\\\" "baz"; test 'hi' 'hi'; word; two words '''
```

alot/helper.py
pacien: The spec (from tests.test_helper.TestSplitCommandline.{test_simple,test_single}) does not want us to match empty strings.

tests/test_helper.py
pacien: Amended.

pazz:
I'm tempted to revert these changes in light of the newly introduced bugs: #1507 #1508

pacien:
I'm not sure that I understand how those two issues relate to this changeset which only deals with the in-app command prompt input parsing.

pazz:
Of course: you are right. I misunderstood the issue report and thought that the linked commit was responsible for a regression. All good :)

PR #1462, opened 2020-01-06
marking and unmarking of spam threads in search mode in a single command

FiskerLars:
I assign them to S and H keys respectively, by default.

Adding a spam learning hook is still TODO

### DIFF
--- a/alot/commands/search.py
+++ b/alot/commands/search.py
@@ -19,6 +19,7 @@ from ..db.errors import DatabaseROError
 MODE = 'search'
 
 
+
 @registerCommand(MODE, 'select')
 class OpenThreadCommand(Command):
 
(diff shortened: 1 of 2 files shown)


lucc:
I do not really understand why there needs to be an extra command for this and why it needs to go into alot. Why can it not be a binding or a hook in your config?

FiskerLars:
OK, I try to handle it as a hook/binding somewhere

lucc:
I am mostly asking *why*! If there is a good reason to do it in alot and not in a hook we can discuss it.  At the moment it just looks like a binding to tag/untag with `spam,killed`. If that is everything it is better done in the users config. If there is some special hook logic that can not be done with a `:call` and is beneficial to other users it has a chance of being merged.

lucc:
If you have an interesting hook or so you can also contribute it to the wiki.

PR #1465, opened 2020-01-29
Clear the screen before initial paint

andir:
If the terminal previously did show another curses application (e.g.
VIM) that content might be shown upon starting alot until the first
command has been executed and the result printed. By simply clearing the
screen before entering the mainloop we can work around that.

### DIFF
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -134,6 +134,9 @@ class UI:
         logging.info('setup gui in %d colours', colourmode)
         self.mainloop.screen.set_terminal_properties(colors=colourmode)
 
+        # clear the screen before the initial frame
+        self.mainloop.screen.clear()
+
         logging.debug('fire first command')
         loop.create_task(self.apply_commandline(initialcmdline))
 


pazz:
Thanks, this looks simple enough and harmless to me.
Such a patch should not be necessary though if urwid works as expected. Can you give a bit more context how to reproduce this? which os, shell, and slot version are you using?

andir:
> Thanks, this looks simple enough and harmless to me.
> Such a patch should not be necessary though if urwid works as expected. Can you give a bit more context how to reproduce this?

1. start termite
2. start vim
3. type some text
4. `:q!`
5. `termite`

while termite loads it will briefly show the old buffer contents of VIM.

> which os,

NixOS
> shell,

zsh, probably more importantly this happens with termite, and my custom terminal (https://github.com/andir/ate) which probably means a few more VTE based terminals as well.

> and alot version are you using?

0.8.1

teto:
I am using alot 0.9 with the terminal kitty and it seems alot doesn't use altscreen anymore, (I believe it used to). I am using nixos as well, might be related.

PR #1474, opened 2020-03-07
Fix the missing HTML message

atenart:
Fixes the missing HTML message, displayed when an HTML body can't be
rendered because there is no functional mailcap entry. The message
refers to the online FAQ, but the item was the wrong one.


pazz:
thanks

PR #1475, opened 2020-03-13
[WIP] Mimetree and mime part support. Resolve #862.

### LINKED ISSUE #862: Feature: Add a tree view for the mime parts of a message
This is a feature request (or discussion) for a new alternative way to display a message in thread buffers. It could be accessible with a new command for thread buffers `togglemimetree` which should be used similar to `togglesource` or `toggleheaders`.

If activated the mime tree of the message should be displayed below the currently displayed header lines (so this command should work independently of `toggleheaders`). Each line should correspond to one mime part and the lines should be indented like the messages in thread view to indicate the tree structure. 

Some commands are of interest when viewing the mime tree:
- something like `select` to display a rendered version of the current mime part (either inside the tree or in a new buffer)
- something like `select` currently works on attachments to open mime parts with external tools
- something like `pipeto` to pipe only this mime part to an external shell command
- something like `save` to save only this mime part to a file
- others?

This feature is inspired by the [view-attachments](http://www.mutt.org/doc/manual/#attach-menu) of mutt. In mutt it looks like this: 
view-attachments

ryneeverett:
This is a rebase and continuation of #894.

- [ ] rebase
- [ ] move pipeto functionality into separate branch/PR
- [ ] default keybindings for `togglemimetree` and `togglemimepart`
- [ ] show current mime part

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -514,6 +514,11 @@ class EditNewCommand(Command):
     forced={'mimetree': 'toggle'},
     arguments=[(['query'], {'help': 'query used to filter messages to affect',
                             'nargs': '*'})])
+@registerCommand(
+    MODE, 'togglemimepart', help='switch between html and plain text message',
+    forced={'mimepart': 'toggle'},
+    arguments=[(['query'], {'help': 'query used to filter messages to affect',
+                            'nargs': '*'})])
 class ChangeDisplaymodeCommand(Command):
 
     """fold or unfold messages"""
@@ -590,7 +595,14 @@ class ChangeDisplaymodeCommand(Command):
             all_headers = not mt.display_all_headers \
                 if self.all_headers == 'toggle' else self.all_headers
             if self.mimepart:
-                mt.set_mimepart(ui.get_deep_focus().mimepart)
+                if self.mimepart == 'toggle':
+                    message = mt.get_message()
+                    mimetype = {'plain': 'html', 'html': 'plain'}[
+                        message.mime_part.get_content_subtype()]
+                    mimepart = get_body_part(message.get_email(), mimetype)
+                elif self.mimepart is True:
+                    mimepart = ui.get_deep_focus().mimepart
+                mt.set_mimepart(mimepart)
             if self.mimetree == 'toggle':
                 tbuffer.focus_selected_message()
             mimetree = not mt.display_mimetree \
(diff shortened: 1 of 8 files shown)


ryneeverett:
> 4. perhaps it is time to finally introduce a single-message buffer? In any case, we should review any changes proposed here under the assumption that this will come in future, and we should try to write code in a way that miminizes later overhead.

I implemented a single message buffer in one of my earliest drafts. As I recall I didn't even have to create a custom buffer I just passed the mimepart into the base `Buffer` class and it worked perfectly. However I didn't find this to be a useful or pleasant UI so rather than add a new `Buffer` subclass to add a pretty header I switched to the mime part toggling strategy. Long story short this is fairly trivial now that the mimepart is easily accessible.

> 6. Unit tests and Codeclimate should not fail before we merge

It seems like Codeclimate fails if there are any "issues" and any change to a function that codeclimate doesn't like is considered an issue. One of the failures is for adding the new commands to a list of completion variables in a method codeclimate considers too complex. Another is actually for decreasing the complexity by replacing `extract_body` (Cognitive Complexity 11) with `extract_body_part` (Cognitive Complexity 8). I doubt you really want those unrelated methods simplified in this branch.

lucc:
Thank you @ryneeverett for working on this. I had previously given up because I had hit a wall.

I like the idea of a single message buffer where I can see a mime tree and expand and collaps the individual mime parts just like I can currently see a message tree in a thread buffer and expand and collapse the individual messages.

ryneeverett:
> I like the idea of a single message buffer where I can see a mime tree and expand and collaps the individual mime parts just like I can currently see a message tree in a thread buffer and expand and collapse the individual messages.

I personally don't see any use for this functionality but it would be pretty easy to implement on top of this branch. I've added a reference to the email body part to the mime tree leaves and I've broken out the `extract_body_part` function to return the displaystring from an individual body part, so the mimetree itself now has all the data needed to construct such a CollapsibleTree.

ryneeverett:
> 2. perhaps we can add a section or paragraph on mime trees in the usage docs?

I'm open to suggestions but I'm not sure where I'd add an additional section beyond the automatic documentation in the "Commands in ‘thread’ mode" section page. Adding a top-level usage page on these features strikes me as excessive.

ryneeverett:
Closing in favor of #1480 and #1481.

Issue #1478, opened 2020-03-21
Allow `loop_hook` to run asynchronously


lenormf:
**Is your feature request related to a problem? Please describe.**

I want to use the `loop_hook` to run `offlineimap` regularly, and update the buffers that contain unread messages. Here is my implementation:

```python
def loop_hook(ui):
    try:
        subprocess.check_call("offlineimap",
                              stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
    except subprocess.CalledProcessError as e:
        ui.notify("`offlineimap` returned %d" % e.returncode, "error")
        return False

    for search_buffer in ui.get_buffers_of_type(alot.buffers.SearchBuffer):
        info = search_buffer.get_info()
        querystring = info["querystring"].split(" ") if "querystring" in info else []
        if "tag:unread" in querystring and info["result_count"] > 0:
            ui.notify("buffer reloaded")
            search_buffer.rebuild()
```

I'd like this function to run asynchronously, to prevent it from locking the UI when it's running.

**Describe the solution you'd like**

Make the hook asynchronously by default.

**Describe alternatives you've considered**

Declaring the hook as `async` yields an error:

```
/usr/lib/python3.8/site-packages/alot/ui.py:31: RuntimeWarning: coroutine 'loop_hook' was never awaited
  callable_(*args, **kwargs)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

**Additional context**

Version 0.9

PR #1479, opened 2020-03-21
periodic-hook async

gauteh:
…lar style as call)

### DIFF
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -28,7 +28,9 @@ from .widgets.globals import ChoiceWidget
 async def periodic(callable_, period, *args, **kwargs):
     while True:
         try:
-            callable_(*args, **kwargs)
+            t = callable_(*args, **kwargs)
+            if asyncio.iscoroutine(t):
+                await t
         except Exception as e:
             logging.error('error in loop hook %s', str(e))
         await asyncio.sleep(period)


pazz:
sure, but what is the use case? I usually try to avoid adding such "features" just for the sake of it because the code base is already quite bloated...

gauteh:
I was trying to set up a way to sync mail with responsive UI that still could update the UI when done:

```
async def run(cmd):
  import logging
  import asyncio

  proc = await asyncio.create_subprocess_shell(
      cmd,
      stdout=asyncio.subprocess.PIPE,
      stderr=asyncio.subprocess.PIPE)

  stdout, stderr = await proc.communicate()

  logging.debug(f'[{cmd!r} exited with {proc.returncode}]')
  if stdout:
    logging.debug(f'[stdout]\n{stdout.decode()}')
  if stderr:
    logging.debug(f'[stderr]\n{stderr.decode()}')

def poll(ui, refresh = True):
  ui.notify("polling..")
  import logging

  async def inner():
    await run('tmux wait-for -L p1 \; split-window -d -v -p 20 "/home/gauteh/.bin/met/mail-sync ; tmux wait-for -U p1"')
    await run('tmux wait-for -L p1 \; wait-for -U p1')

    if refresh:
      ui.current_buffer.rebuild()
      ui.update()

  return inner()

def loop_hook(ui):
  return poll(ui, False)
```

but it seems that this will always reset the cursor/buffer, so I could not really get it to work the way I hoped.

pazz:
But the issue with resetting a cursor (I presume you mean urwid focus) is unrelated to the non-existence of async periodic hook calls correct?

gauteh:
Yes, so this pr does not enable my use case anyway.

ons. 6. mai 2020, 11:29 skrev Patrick Totzke <notifications@github.com>:

pazz:
But #1499 might..
I will include your patch for completeness sake.
Thanks

gauteh:
Thanks :)

lenormf:
Could this be a fix for #1478 ?

pazz:
> Could this be a fix for #1478 ?

Yes, thanks for the refedence here @lenormf . It should indeed also allow your construction.

PR #1480, opened 2020-03-23
Take 3: Mimetree and mime part support. Resolve #862.

### LINKED ISSUE #862: Feature: Add a tree view for the mime parts of a message

ryneeverett:
This supersedes #1475 in response to the review there.

### DIFF
--- a/alot/buffers/thread.py
+++ b/alot/buffers/thread.py
@@ -56,6 +56,8 @@ class ThreadBuffer(Buffer):
         info['message_count'] = self.message_count
         info['thread_tags'] = self.translated_tags_str()
         info['intersection_tags'] = self.translated_tags_str(intersection=True)
+        info['mimetype'] = (
+            self.get_selected_message().mime_part.get_content_type())
         return info
 
     def get_selected_thread(self):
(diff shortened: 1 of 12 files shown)


pazz:
@lucc what do you think?

PR #1483, opened 2020-04-02
reply: use `email.utils.formataddr` as in compose command

### LINKED ISSUE #1482: `from` header is not correctly encoded when replying
First, thanks for prodiving such a good companion for notmuch! alot is awesome!

In thread buffer when invoking reply command, realname associated to my account is not correctly encoded in `From` header (some non ascii char are leaking in header)
It seems this is related to `determine_sender` which uses custom `format_addr` instead of `email.utils.format_addr`.

I provide a pull request to make my point clearer, feel free to discard it if it has nothing to do with this encoding issue ;)

stregouet:
to format address used in `From` header we should use `formataddr` from
stdlib, this will handle encoding properly (we shouldn't leak non-ascii
characters in headers).
Actually this is same behaviour in `compose` command

(closes #1482)

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -88,7 +88,7 @@ def determine_sender(mail, action='reply'):
                     logging.debug('using realname: "%s"', realname)
                     logging.debug('using address: %s', address)
 
-                    from_value = formataddr((realname, address))
+                    from_value = email.utils.formataddr((realname, address))
                     return from_value, account
 
     # revert to default account if nothing found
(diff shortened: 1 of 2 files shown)


patricktotzke:
No: the stdlibs formataddr is broken, or at least [has a strange semantics that introduced problems in alot](9b75fec97b4ad23178a730623ba691fbe666a747). We're deliberately not using that any more.

stregouet:
yes pazz told me on issue #1482  about problem with stdlib formataddr so why using it with compose command?

pazz:
I believe this can be closed? If we use the stdlib routing explicitly somewhere then this is by accident and should be fixed independently I think.

stregouet:
yes you're right. I close it

Issue #1484, opened 2020-04-02
Emails with non-ASCII subjects are bounced by SMTP server without UTF-8 support


aepsil0n:
**Describe the bug**
When sending emails using alot that include non-ASCII characters to a SMTP server that does not support UTF-8 in mail headers, the email is bounced by the server.
Likely due to a bug in Python, see [issue 40157][python-issue]

**Software Versions**
- Python version: 3.6-3.8
- Alot version: 0.8-0.9

**To Reproduce**
1. Get a SMTP server, that does not support UTF-8 in headers.
2. Compose a mail with non-ASCII characters in the subject
3. Send it via that server under 1.

**Error Log**
What I get from the SMTP server:
`550 Subject contains invalid characters. (in reply to end of DATA command)`

The bounced message looks like this (somewhat redacted for privacy reasons):

```
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: quoted-printable
Subject: Test üäö
From: one.of.my@mails.com
To: another.mail@of.mine
Message-ID: <redactedmessageid@somecomputer>
User-Agent: alot/0.8
Date: Thu, 02 Apr 2020 12:41:21 +0200
```

The important aspect is `Subject: Test üäö`. This has not been encoded. I have traced this down through the source code to [alot/commands/envelope.py:245](https://github.com/pazz/alot/blob/cf433946921fe7475c3457affea02fd3328faf46/alot/commands/envelope.py#L245).

It was a bit strange to see this behaviour given the existence of `email.policy.SMTPUTF8`. But apparently they both encode headers the same way. I have the impression that this is a bug in the Python standard library, so I also opened [issue 40157][python-issue] on the Python bug tracker.

As a workaround, removing the policy parameter and thereby falling back to the default compat32 policy, fixes the problem for me. Does that have any unintended side-effects?

[python-issue]: https://bugs.python.org/issue40157

pazz:
@aepsil0n Could you check if #1485 fixes your issue?

aepsil0n:
Thanks for the heads-up. I just checked the issue with the current master (8e96d504) and can confirm that it fixes my issue.

pazz:
Wait: what? This supposed fix is not yet part of the master branch..

aepsil0n:
There are two commits (5855e100 and 3ea84785) by @stregouet on master. You merged them in #1485. That is why I simply pulled in the recent master branch to check. Were you not referring to these commits?

[...]

Issue #1434, opened 2019-09-12
Transcode non-ASCII headers using MIME encoding


pacien:
**Is your feature request related to a problem? Please describe.**

Messages composed with alot may be rejected by some mail servers if headers contain non-ASCII characters.

```
  linux-kernel@vger.kernel.org
    host vger.kernel.org [209.132.180.67]
    SMTP error from remote mail server after end of data:
    550 5.7.1 Content-Policy reject msg: Message headers can not have 8-bit non-ASCII characters in it; Use MIME encodings if such are needed!
```

**Describe the solution you'd like**

The user should be warned before sending such message and be offered the possibility to have those headers automatically transcoded. An option to do that automatically could also be added.

**Describe alternatives you've considered**

Normalising those characters to ASCII manually. I always forget to do that before it's too late.

**Additional context**

https://github.com/pazz/alot/issues/1 was supposed to handle this encoding issue, which still seems to exist in alot 0.8.

pazz:
wow, you've found #1 :D 
These things are actually handled by pythons email module, so I doubt that the messages are not RFC compliant. But I'd be open to adding some config setting to enforce more strict encodings than what is considered default in 2019, by tweaking the calls to the email module where appropriate.

pacien:
Migrating to Python 3.6's `email.message.EmailMessage` (issue #1409) will probably solve this as a nice side effect:

> email.header: Internationalized headers
> […] In the current API encoding and decoding of headers is handled transparently by the dictionary-like API of the EmailMessage class. […]

PR #1485, opened 2020-04-02
envelope: set policy when constructing email

stregouet:
`default` policy ensure headers are correctly encoded (i.e. will not
leak non-ascii char in headers)

(related to #1482)

### DIFF
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -14,6 +14,7 @@ import re
 import tempfile
 import textwrap
 import traceback
+import sys
 
 from . import Command, registerCommand
 from . import globals
@@ -135,7 +136,7 @@ class SaveCommand(Command):
         mail = envelope.construct_mail()
         # store mail locally
         path = account.store_draft_mail(
-            mail.as_string(policy=email.policy.SMTP))
+            mail.as_string(policy=email.policy.SMTP, maxheaderlen=sys.maxsize))
 
         msg = 'draft saved successfully'
 
@@ -243,7 +244,8 @@ class SendCommand(Command):
 
             try:
                 self.mail = self.envelope.construct_mail()
-                self.mail = self.mail.as_string(policy=email.policy.SMTP)
+                self.mail = self.mail.as_string(policy=email.policy.SMTP,
+                                                maxheaderlen=sys.maxsize)
             except GPGProblem as e:
                 ui.clear_notify([clearme])
                 ui.notify(str(e), priority='error')
--- a/alot/db/envelope.py
+++ b/alot/db/envelope.py
@@ -298,6 +298,8 @@ class Envelope:
         if uastring:
             headers['User-Agent'] = [uastring]
 
+        # set policy on outer_msg to ease encoding headers
+        outer_msg.policy = email.policy.default
         # copy headers from envelope to mail
         for k, vlist in headers.items():
             for v in vlist:
--- a/tests/db/test_envelope.py
+++ b/tests/db/test_envelope.py
@@ -112,3 +112,19 @@ class TestEnvelope(unittest.TestCase):
         })
         self.assertEqual(envlp.body_txt,
                          'Some body content: which is not a header.')
+
+    @mock.patch('alot.db.envelope.settings', SETTINGS)
+    def test_construct_encoding(self):
+        headers = {
+            'From': 'foo@example.com',
+            'To': 'bar@example.com',
+            'Subject': 'Test email héhé',
+        }
+        e = envelope.Envelope(account=test_account,
+                              headers={k: [v] for k, v in headers.items()},
+                              bodytext='Test')
+        mail = e.construct_mail()
+        raw = mail.as_string(policy=email.policy.SMTP)
+        actual = email.parser.Parser().parsestr(raw)
+        self.assertEqual('Test email =?utf-8?b?aMOpaMOp?=', actual['Subject'])
+
--- a/tests/db/test_envelope.py
+++ b/tests/db/test_envelope.py
@@ -20,6 +20,7 @@ import os
 import tempfile
 import unittest
 from unittest import mock
+import sys
 
 from alot.db import envelope
 from alot.account import Account
@@ -124,7 +125,6 @@ class TestEnvelope(unittest.TestCase):
                               headers={k: [v] for k, v in headers.items()},
                               bodytext='Test')
         mail = e.construct_mail()
-        raw = mail.as_string(policy=email.policy.SMTP)
+        raw = mail.as_string(policy=email.policy.SMTP, maxheaderlen=sys.maxsize)
         actual = email.parser.Parser().parsestr(raw)
         self.assertEqual('Test email =?utf-8?b?aMOpaMOp?=', actual['Subject'])
-


pacien:
This possibly solves #1434 and #1484 in addition to #1482.

stregouet:
@pazz do you know why travis is not reporting status/running build ?

pazz:
> @pazz do you know why travis is not reporting status/running build ?

Travis integration in github is a troublemaker at the moment. I believe its due to github restructuring "apps". I don't really have the time to look into it and am thinking of removing it alltogether.

pazz:
I've set up travis from scratch and it seems to run now.
There are failing tests though..

pazz:
Ah, they fail only on python <3.8 which makes sense.
In fact, perhaps we should find a way to make this work on python 3.6 and 3.7..
https://travis-ci.com/github/pazz/alot/builds/158944184

stregouet:
cool! thanks for your time pazz. I will try to find a solution for python 3.6 and 3.7

stregouet:
hi,
sorry for delayed response…
I cannot find travis-ci log anymore, probably because I'm too late… Anyway I run test in with `   maxking/mailman-ci-runner` docker image (used in [mailman](https://gitlab.com/mailman/mailman) project). But I only get failure with python3.6 (not python3.7 as suggested in previous [comment](https://github.com/pazz/alot/pull/1485#issuecomment-610327080))

I finally find where is difference with python3.8: it's in function `email._header_value_parser._refold_parse_tree` at the beginning `maxlen` variable  default to `sys.maxsize` in python3.7 and above but in python3.6 it is `float("+inf")` which break arithmetic and string slicing.

I can see 2 options to fix this:
1. monkeypatching `_refold_parse_tree` in python3.6 but this function is quite big…
2. use optionnal parameter `maxheaderlen` in `as_string` method and force it to `sys.maxsize`

what do you think? do you see another option? (not to forget initial goal of this PR: fixing encoding of headers when sending mail)

stregouet:
@pazz did you have time to think about my proposal?
(not really sure if I need to ping you, I'm not used to github…)

pazz:
Sorry, not yet. I will look into this at the end of the week for sure though.

pazz:
Could you rebase this to master? It seems that travis is choking on some sphinx-versioning stuff that I remember having fixed some time ago.

Also, the following test failure is reported by travis (now, after triggering it again) for python 3.7:

```
1318ERROR: test_construct_encoding (tests.db.test_envelope.TestEnvelope)
1319----------------------------------------------------------------------
1320Traceback (most recent call last):
1321  File "/opt/python/3.7.1/lib/python3.7/unittest/mock.py", line 1191, in patched
1322    return func(*args, **keywargs)
1323  File "/home/travis/build/pazz/alot/tests/db/test_envelope.py", line 127, in test_construct_encoding
1324    raw = mail.as_string(policy=email.policy.SMTP)
1325  File "/opt/python/3.7.1/lib/python3.7/email/message.py", line 158, in as_string
1326    g.flatten(self, unixfrom=unixfrom)
1327  File "/opt/python/3.7.1/lib/python3.7/email/generator.py", line 116, in flatten
1328    self._write(msg)
1329  File "/opt/python/3.7.1/lib/python3.7/email/generator.py", line 195, in _write
1330    self._write_headers(msg)
1331  File "/opt/python/3.7.1/lib/python3.7/email/generator.py", line 222, in _write_headers
1332    self.write(self.policy.fold(h, v))
1333  File "/opt/python/3.7.1/lib/python3.7/email/policy.py", line 183, in fold
1334    return self._fold(name, value, refold_binary=True)
1335  File "/opt/python/3.7.1/lib/python3.7/email/policy.py", line 205, in _fold
1336    return value.fold(policy=self)
1337  File "/opt/python/3.7.1/lib/python3.7/email/headerregistry.py", line 258, in fold
1338    return header.fold(policy=policy)
1339  File "/opt/python/3.7.1/lib/python3.7/email/_header_value_parser.py", line 144, in fold
1340    return _refold_parse_tree(self, policy=policy)
1341  File "/opt/python/3.7.1/lib/python3.7/email/_header_value_parser.py", line 2650, in _refold_parse_tree
1342    part.ew_combine_allowed, charset)
1343  File "/opt/python/3.7.1/lib/python3.7/email/_header_value_parser.py", line 2727, in _fold_as_ew
1344    first_part = to_encode[:text_space]
1345TypeError: slice indices must be integers or None or have an __index__ method
```

stregouet:
just rebased on master

I think I got why travis is failing on 3.7 but test on my laptop is not! from traceback above I guess, travis is running  3.7.1, but my docker is using 3.7.7. Do you know why travis is not using last 3.7 release? 

(_header_value_parser.py has changed between version [3.7.1](https://github.com/python/cpython/blob/v3.7.1/Lib/email/_header_value_parser.py#L2727) and version [3.7.7](https://github.com/python/cpython/blob/v3.7.7/Lib/email/_header_value_parser.py#L2727))

pazz:
Quoting stregouet (2020-05-06 14:20:33)

> just rebased on master
> 
> I think I got why travis is failing on 3.7 but test on my laptop is not! from
> traceback above I guess, travis is running 3.7.1, but my docker is using 3.7.7.
> Do you know why travis is not using last 3.7 release?
> 
> (_header_value_parser.py has changed between version 3.7.1 and version 3.7.7)

OK, goot do know.
No, I'm not sure why travis uses the versions it does. But it is not really an issue
for us because we should really support most recent versions.
Given that for most distributions it is not too long ago that they switched away from python 2.7 as a default, I believe dropping support for 3.6 may be a bit premature.
(or is there a good reason for it except that it makes our life a little harder?)

pazz:
I wonder if it is totally unreasonable to drop support for old python versions:
By old I mean those that are not in use as default in current [Fedora (stable is `32` and has python 3.8.2-2)](https://apps.fedoraproject.org/packages/python3) and [Debian (stable is `buster`, has python 3.7.3-1)](https://packages.debian.org/buster/python3) and by extension everything based on it.

stregouet:
it seems reasonable!
unfortunately there is still this "bug" with `float("+inf")` in python 3.7.3 for [_header_value_parser.py](https://github.com/python/cpython/blob/v3.7.3/Lib/email/_header_value_parser.py#L2594)… 

So my test is still failing with python3.7.3

pazz:
Sorry, just to clarify:
Are you saying that you are working on this or that you want to sit it out
and re-submit this PR once we drop support for python 3.7.3 and lower?


Quoting stregouet (2020-05-09 13:52:22)

stregouet:
I just want to help fixing issue with header encoding ;)

Suggested solution in this PR is facing issue with python <= 3.7.3 so we need to either drop support for older version or find a workaround 

As you wrote python 3.7.3 is default version for debian stable so it seems not reasonnable to drop support for it. 
I've made a [suggestion](https://github.com/pazz/alot/pull/1485#issuecomment-620075367) for a workaround, just waiting for your opinion on this ;)

pazz:
Right, thanks.
So if I understand your two suggestions correctly then number two should be the way to go:
if indeed `email.as_string` accepts some parameter that we can set and thereby make it work for all python versions then this sounds like the cheapest, and still correct, option.

stregouet:
ok, thanks for your reply

just pushed another commit, tell me if it seems enough, especially if explanation in commit message is clear

pazz:
> This possibly solves #1434 and #1484 in addition to #1482.

@pacien can you comment on whether this indeed fixed your earlier problem #1434 ?

pacien:
I'm not sure of what changed but I'm afraid I could not reproduce my issue (#1434) in 0.9 with or without this patch.

pazz:
OK thanks @pacien. this at least meanst that things did not get worse :)

stregouet:
cool! again thanks for your time :)
@pazz do you have any idea when this change will be released?

pazz:
Quoting stregouet (2020-05-11 12:46:26)

>    cool! again thanks for your time :)

Thanks to you for pushing this issue (and project)!

>    [1]@pazz do you have any idea when this change will be released?

It is already part of the master branch.
I plan to make a new release soonish as some people were pushing for it (#1500)..

Ornanovitch:
Hi,

I just installed alot few days ago and am configuring it. Manjaro should provides 0.9.1 in a few days.

When I reply to messages, my `from` is received like this by the recipient :
`=?utf-8?b?SsOpcsO0bWUgVmlhbCA8cHJvZEBkb3plY29tcGFnbmllLmZyPg==?=`

Does this fix has a chance to resolve this or should I open a new issue ?

Ornanovitch:
> Hi,
> 
> I just installed alot few days ago and am configuring it. Manjaro should provides 0.9.1 in a few days.
> 
> When I reply to messages, my `from` is received like this by the recipient :
> `=?utf-8?b?SsOpcsO0bWUgVmlhbCA8cHJvZEBkb3plY29tcGFnbmllLmZyPg==?=`
> 
> Does this fix has a chance to resolve this or should I open a new issue ?

Upgraded this morning, I confirm it fixes everything !

PR #1495, opened 2020-04-19
Make notmuch db location logic match upstream

### LINKED ISSUE #1396: Support notmuch database paths without leading /
Avoid storing one's system username in one's dotfiles repo.

"Database paths without a leading / are relative to $HOME":
https://github.com/notmuch/notmuch/commit/95a9c73c72ff61d5514010a1b9bc11f94cbe988c
https://notmuchmail.org/pipermail/notmuch/2019/027762.html
https://github.com/afewmail/afew/pull/233

Tried applying the above patch to notmuch, but that didn't suffice.

kirelagin:
* Allow relative database.path in notmuch config.
* Fallback to $MAILDIR or $HOME/mail.

Fixes #1396.

------

I am sorry, I have no idea how to write tests in Python, in particular, how to define helper constants in them. I suspect I should have put them not at the toplevel, but in the class, but I don’t know how to do that, please, help.

### DIFF
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -15,6 +15,7 @@ from alot.ui import UI
 from alot.commands import *
 from alot.commands import CommandParseError, COMMANDS
 from alot.utils import argparse as cargparse
+from alot.utils.notmuch import find_db
 
 from twisted.internet import asyncioreactor
 asyncioreactor.install()
@@ -120,8 +121,7 @@ def main():
         settings.set('colourmode', options.colour_mode)
 
     # get ourselves a database manager
-    indexpath = settings.get_notmuch_setting('database', 'path')
-    indexpath = options.mailindex_path or indexpath
+    indexpath = options.mailindex_path or find_db(settings)
     dbman = DBManager(path=indexpath, ro=options.read_only)
 
     # determine what to do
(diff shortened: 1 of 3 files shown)


kirelagin:
_Actually, I wish this code was in the bindings or in the C functions that bindings go through._

pazz:
I agree: this is something that the notmuch bindings should expose and we should not re-implement if it already exists as a (c) library function.

 I don't have time to look into this at the moment, but perhaps you could write to the notmuch list if this is not already in the bindings and ask why not, or propose a change to the bindings before you come back here to propose a PR that simply uses the bindings? I believe that'd be the cleanest solution.

kirelagin:
This problem is that this logic is implemented in `notmuch_config_open` and the only use of this function is directly in `main`. As far as I can tell, there is nothing in bindings for working with the configuration, so exposing this logic from the C code will require quite a bit of restructuring. I wouldn’t hope that this is something that will be trivial to do.

pazz:
I've been thinking about this again and it seems to be quite a big patch for a minimal change.
In fact we can assume that the notmuch config does set the `path` variable, so there is really no need to set an explicit default value.
Secondly, why not sumply replace [this line](https://github.com/pazz/alot/pull/1495/commits/15dfbc1bdd0d7194dc4a381b1a29f8625b3278c7#diff-0f91e8e819ac14d63e8cde46e7694887L123) with

```python
indexpath = os.path.abspath(settings.get_notmuch_setting('database', 'path'))
```

PR #1505, opened 2020-05-09
commands/thread/open: add "open" command to open attachments

### LINKED ISSUE #1494: Open attachment with a custom command like xdg-open

It'd be nice to let xdg-open decide which application should be used when opening some attachment in alot.

Currently, alot only tries to use the `~/.mailcap` mapping, which has to be manually kept in sync with other MIME databases (mimeapps.list, desktop entries, …).

An option could be added to delegate the application choice to another program, such as xdg-open.

It's not possible to add a global fallback in .mailcap such as `*/*; xdg-open "%s"` because `*/*` isn't a valid wildcard as per RFC1524 and isn't accepted by the Python lib.

**Additional context**

Since b1c93c4d0c1eeacd64a195f16861bcb73910e739, having `text/*; xdg-open "%s"` in .mailcap causes all text messages, not only attachments, to be opened externally. It's not possible to keep that entry to open log files attachments for instance.

pacien:
This allows the user to open an attachment file with a program of their choice.

GitHub: closes #1494


alot/commands/thread.py
```diff
@@ -934,72 +934,130 @@ async def apply(self, ui):
...
+        self.cmd = cmd
+        self.thread = thread
+        self.spawn = spawn
 
     async def apply(self, ui):
-        logging.info('open attachment')
-        mimetype = self.attachment.get_content_type()
+        try:
+            logging.info('open attachment')
+            attachment = OpenAttachmentCommand._get_attachment(ui)
```
pazz, line 975: isn't this the `OpenAttachmentCommand` class? why not simply call `_get_attachment`?

pazz:
I'd rather not call this command `open` because it may be confused with unfolding a message. How's `view` ? `open-attachment` ?

alot/commands/thread.py
pacien: Python wants the class name for static methods even within the class itself.

pacien:
Amended to rename the command to "open-attachment".

alot/commands/thread.py
pazz: aha. didn't know that. thanks.

alot/commands/thread.py
```diff
@@ -934,72 +934,130 @@ async def apply(self, ui):
...
+            external_handler = \
+                self._get_user_handler(attachment) if self.cmd \
+                else self._get_mailcap_handler(attachment)
+
+            await ui.apply_command(external_handler)
+
+        except RuntimeError as error:
+            ui.notify(str(error), priority='error')
+
+    def _get_user_handler(self, attachment):
```
roosemberth, line 985:
```suggestion
    def _get_ext_cmd_handler(self, attachment):
```

What's the relationship with a user at this level? When I read user, I expect something about profiles or accounts; regardless of how this will be used later.

alot/commands/thread.py
```diff
@@ -934,72 +934,130 @@ async def apply(self, ui):
...
+        temp_file_name, destructor = self._make_temp_file(attachment)
+        return ExternalCommand(self.cmd + [temp_file_name],
+                               on_exit=destructor,
+                               thread=self.thread,
+                               spawn=self.spawn)
+
+    def _get_mailcap_handler(self, attachment):
```
roosemberth, line 992:
```suggestion
    def _get_mailcap_cmd_handler(self, attachment):
```

pazz:
I feel like I have implemented a very similat logic in `alot.commands.globals.EditCommand`.
Perhaps we should rather subclass that to recycle code?

alot/commands/thread.py
```diff
@@ -934,72 +934,130 @@ async def apply(self, ui):
...
+        'help': 'run in a new terminal window',
+    }),
+])
 class OpenAttachmentCommand(Command):
+    """opens an attachment with a given shell command
+       or according to mailcap"""
 
-    """displays an attachment according to mailcap"""
-    def __init__(self, attachment, **kwargs):
+    def __init__(self, cmd=(), thread=False, spawn=False, **kwargs):
```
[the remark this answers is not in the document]
pacien: Amended.

alot/commands/thread.py
pacien: Amended.

alot/commands/thread.py
pacien: Amended.

alot/commands/thread.py
pacien: Amended.

alot/commands/thread.py
```diff
@@ -934,72 +934,130 @@ async def apply(self, ui):
...
+            return focus.get_attachment()
+        elif (getattr(focus, 'mimepart', False) and
+              isinstance(focus.mimepart, Attachment)):
+            return focus.mimepart
         else:
-            ui.notify('unknown mime type')
+            raise RuntimeError('not focused on an attachment')
+
+    @staticmethod
+    def _make_temp_file(attachment, prefix='', suffix=''):
```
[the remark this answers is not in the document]
pacien: The ExternalCommand in charge of executing the command using the temporary file does not accept context managers. It's not either possible to use a "with" block on top of it because it can spawn an independant thread.

pacien:
The EditCommand and OpenAttachmentCommand implementations do not seem to share much in common: the way the file path is substituted in the command differs with mailcap templates.

PR #1509, opened 2020-05-13
Fix MessageSummaryWidget's child TagWidget rendering.

### LINKED ISSUE #1433: In the thread view, the focused message's summary is rendered incorrectly in case it has multiple tags.
**Describe the bug**
In the thread view, the focused message's summary is rendered incorrectly in case it has multiple tags – only one of the tags has the correct colors, the other ones have their unfocused coloring.

tags

**Software Versions**
- Python version: 3.6
- Alot version: 0.8.1

**To Reproduce**
1. open a message with multiple tags;
2. see that only one tag has the correct colors.

**Additional notes**
The same problem in the search view is handled by overriding `render` method:
https://github.com/pazz/alot/blob/244180d52f4ace8181bb288e77402c766ad620c1/alot/widgets/search.py#L68-L71

Here is a quick and dirty fix which works for me. I am not familiar enough with Urwid and Alot to fix in properly.
```diff
--- a/alot/widgets/thread.py
+++ b/alot/widgets/thread.py
@@ -53,6 +53,14 @@ class MessageSummaryWidget(urwid.WidgetWrap):
                 cols.append(('fixed', tag_widget.width(), tag_widget))
         line = urwid.AttrMap(urwid.Columns(cols, dividechars=1), attr,
                              focus_att)
+
+        def _render_wrap(size, focus=False):
+            for c in cols:
+                if isinstance(c, tuple):
+                    c[2].set_map('focus' if focus else 'normal')
+            return urwid.AttrMap.render(line, size, focus)
+        line.render = _render_wrap
+
         urwid.WidgetWrap.__init__(self, line)
 
     def __str__(self):
```

mmartin:
Fixes #1433

### DIFF
--- a/alot/widgets/thread.py
+++ b/alot/widgets/thread.py
@@ -60,6 +60,19 @@ class MessageSummaryWidget(urwid.WidgetWrap):
                 cols.append(('fixed', tag_widget.width(), tag_widget))
         line = urwid.AttrMap(urwid.Columns(cols, dividechars=1), attr,
                              focus_att)
+
+        # In case when an e-mail has multiple tags, Urwid assumes that only
+        # one of the summary line's TagWidget-s can be focused. Which means
+        # that when the summary line is focused, only one TagWidget is rendered
+        # as focused and the rest of them as unfocused. This forces to render
+        # all TagWidget-s as focused. Issue #1433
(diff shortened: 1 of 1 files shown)


pazz:
cheers :+1:

Issue #1516, opened 2020-05-26
Case sensitivity with abook using notmuch address


mrichar1:
I've set up alot to use `notmuch address` as the abook entry following the docs:

```
command = 'notmuch address --format=json --output=recipients date:1Y..'
regexp = '\[?{"name": "(?P<name>.*)", "address": "(?P<email>.+)", "name-addr": ".*"}[,\]]?'
shellcommand_external_filtering = False
ignorecase = True
```

When I 'search' by typing a partial name and hitting tab, it seems to be case-sensitive in what it searches for, despite setting `ignorecase`.

For example, I have several contacts called 'Bruce', with the following data in notmuch:

```
BROWN Bruce <bruce.brown@example.com>
Bruce Smith <Bruce.Duncan@example.com>
BRUCE David <David.Bruce@example.com>
```

When searching, `bruce` matches only the first one; `Bruce` matches all 3, and `BRUCE` matches only the last one. 

**Software Versions**
- Python version: 3.7.5
- Notmuch version:  0.29.1
- Alot version: github master

pazz:
Thanks for reporting this.
In case you want to give it a try and submit a fix: the problem is that 
[the constructor of AdressBook Subclasses](https://github.com/pazz/alot/blob/master/alot/addressbook/external.py#L39) first set the value of `self.reflags` (by calling the super method, which interprets the `ignorecase` flat) and then, 3 lines below, overwrites this value with the default of the `reflags` parameter (0).

mrichar1:
Fixed in #1517

PR #1517, opened 2020-05-27
Ensure default ignorecase for external abooks.

mrichar1:
Fix for #1516

### DIFF
--- a/alot/addressbook/external.py
+++ b/alot/addressbook/external.py
@@ -37,7 +37,8 @@ class ExternalAddressbook(AddressBook):
         AddressBook.__init__(self, **kwargs)
         self.commandline = commandline
         self.regex = regex
-        self.reflags = reflags
+        if reflags:
+            self.reflags = reflags
         self.external_filtering = external_filtering
 
     def get_contacts(self):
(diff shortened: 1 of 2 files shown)


mrichar1:
So if I'm reading it through right the logic (after this patch) is:

1. Parent __init__ is called and sets self.reflags to `re.IGNORECASE` default) or `0` (depending on `ignorecase`).
2. If chils classes' `reflags` param is truthy (defaults to `0`) then self.reflags is overwritten with this new value.

Do you mean that if say `ignorecase=True` AND `reflags` is truthy then they should be combined, i.e.  `self.reflags|=reflags`

Happy to update it to do this, or some other variation, as you see fit. Also happy to update the docs to explain whichever is settled on.

pazz:
Yes of course, you are right. I somehow thought that ignore case had preference but I was confused :)
I think leaving the code as is (with a preference to `reflags`) is fine.  In fact, `reflags` is never actually used anywhere: It is not read from the config file nor used from within alot as far as I can see. So we could even remove this parameter entirely.

mrichar1:
OK, I'll update the docstrings to explain the ordering just now...

Issue #1520, opened 2020-06-01
search query slowness due to athor replacement


meskio:
**Describe the bug**
If [thread_authors_replace_me](https://alot.readthedocs.io/en/latest/configuration/config_options.html#thread-authors-replace-me) is set every search buffer reload will need to read every single email from disk. This makes the search very slow mostly in the case of a lot of encrypted emails.

**Software Versions**
- Python version: 3.8.3
- Notmuch version: 0.29.1
- Alot version: git master (0.9.1 + 2348014eb6d654ef123c6c353d2dc4306f226305)

**To Reproduce**
1. Set `thread_authors_replace_me` to true
2. do a search including a lot of encrypted email. It will take some seconds to load while gpg is working.

I'm personally happy to keep `thread_authors_replace_me` to false in my setup, if I'm the only one affected by it we can close the issue.

paretje:
I have the same issue. It doesn't need to be encrypted though, signed messages trigger this as well. Quite a lot of time is spent loading the email's from disk, and in gpg (getting the key and verifying signatures). None of this seems necessary, and is a regression from previous versions. I might have a look later on.

These are the top calls:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    472/1    0.003    0.000   40.620   40.620 {built-in method builtins.exec}
        1    0.000    0.000   40.620   40.620 <string>:1(<module>)
        1    0.000    0.000   40.620   40.620 /usr/lib/python3.8/runpy.py:197(run_module)
        1    0.000    0.000   40.619   40.619 /usr/lib/python3.8/runpy.py:64(_run_code)
        1    0.000    0.000   40.619   40.619 /home/kevin/vcs/once/alot/alot/__main__.py:4(<module>)
        1    0.000    0.000   40.188   40.188 /home/kevin/vcs/once/alot/alot/__main__.py:86(main)
        1    0.000    0.000   37.591   37.591 /home/kevin/vcs/once/alot/alot/ui.py:47(__init__)
        1    0.000    0.000   37.590   37.590 /usr/lib/python3/dist-packages/urwid/main_loop.py:276(run)
        1    0.000    0.000   37.590   37.590 /usr/lib/python3/dist-packages/urwid/main_loop.py:374(_run)
        1    0.000    0.000   37.589   37.589 /usr/lib/python3/dist-packages/urwid/main_loop.py:1328(run)
[...]

PR #1523, opened 2020-06-08
Lazy loading of message mime_part

paretje:
Closes: #1520

### DIFF
--- a/alot/buffers/thread.py
+++ b/alot/buffers/thread.py
@@ -57,7 +57,7 @@ class ThreadBuffer(Buffer):
         info['thread_tags'] = self.translated_tags_str()
         info['intersection_tags'] = self.translated_tags_str(intersection=True)
         info['mimetype'] = (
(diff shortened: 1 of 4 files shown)


pazz:
can you explain why this should fix the slowness issue? it looks to me that you're just caching calls to those methods that read the email from disk.
I thought the problem was that the message is read from disk in the first place, instead of getting the relevant info from notmuch?

paretje:
Currently `Message.mime_part` is set in `Message.__init__`. As a result, when `Thread.get_authors` calls `Thread.get_messages`, the  message contents are read from disk. Here, I defined `Message.get_mime_part` and `Message.set_mime_part`, and the mime part is no longer set during init. As a result, the mime part is only loaded on the first call of `Message.get_mime_part`. As `Message.get_author` doesn't need the mime part, this fixes the slowness issue.

paretje:
I'm sorry for the delay. I updated the commit message.

PR #1524, opened 2020-06-08
db/utils: encode 8bit emails as utf-8

### LINKED ISSUE #1314: Special characters not displays correctly

meskio:
The encoding as 'raw-unicode-escape' was producing unreadable characters
for non-ascii letters.

Closes #1314

### DIFF
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -429,11 +429,7 @@ def remove_cte(part, as_string=False):
     # decoding into a str is done at the end if requested
     elif '8bit' in cte:
         logging.debug('assuming Content-Transfer-Encoding: 8bit')
-        # Python's mail library may decode 8bit as raw-unicode-escape, so
-        # we need to encode that back to bytes so we can decode it using
-        # the correct encoding, or it might not, in which case assume that
-        # the str representation we got is correct.
-        bp = payload.encode('raw-unicode-escape')
+        bp = payload.encode('utf-8')
 
     elif 'quoted-printable' in cte:
         logging.debug('assuming Content-Transfer-Encoding: quoted-printable')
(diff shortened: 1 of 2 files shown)


meskio:
I have being heavily using it for the last week reading emails in different languages and encodings and reading a lot of encrypted and signed emails and I think this fixes the issue without problems.

pazz:
Any chance you could add unit tests from the demo messages posted in #1314?
It seems that this issue was introduced by a changed behaviour of the email lib and having tests to cover this would be helpful if that happens again..

meskio:
I'll give it a try.

meskio:
After some testing around I found out the problem comes from passing the `encoded` part over to the mailcap. I modify the existing `test_simple_utf8_file` to use `cat` in mailcap and this fails in master but not here.

The other changes are just cosmetic to avoid having a warning in the test for not closing the file.

Does it sound good? Should I rebase the two commits in one?

pazz:
Could you rebase? The conflict looks like this has made it into master somehow?

meskio:
It looks like someone has already commited the same fix:
https://github.com/pazz/alot/commit/37395809db473fb9a4157084a5b1ea3165914556

I'm happy the problem is fixed.

Issue #1522, opened 2020-06-04
Content-Transfer-Encoding badly interpreted when charset in Content-Type is between quote


guijemont:
Before you submit a bug report, please make sure that the issue still exists on the master branch!

**Describe the bug**
I have a bunch of emails with headers that contain:
```
Content-Type: text/plain; charset="utf-8"
Content-Transfer-Encoding: quoted-printable 
```
Though in all the examples I could find, several headers were in-between these two. For all these emails, non-ascii characters appear incorrectly. Looking at the debug log, I see for these messages that the `Content-Transfer-Enconding` is misinterpreted:
```
DEBUG:utils:Content-Transfer-Encoding: "8bit"
DEBUG:utils:assuming Content-Transfer-Encoding: 8bit
DEBUG:utils:command: more /tmp/pb54ils2
DEBUG:utils:parms: ('text/plain=', 'charset=UTF-8')
```

Worthy of notes, I have found another `quoted-printable` email that is correctly displayed, with the headers looking like:
```
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: quoted-printable
```
Note that unlike in the previous case, the charset is not between quotes.

In that case, non-ascii characters are displayed properly, and the debug log yields:
```
DEBUG:utils:Content-Transfer-Encoding: "quoted-printable"
DEBUG:utils:assuming Content-Transfer-Encoding: quoted-printable
```

**Software Versions**
- Python version: 3.8.2-0ubuntu2
- Notmuch version: 0.29.3-1ubuntu2
- Alot version: 2348014eb6d654ef123c6c353d2dc4306f226305

**To Reproduce**
Steps to reproduce the behaviour:
1. open a correctly formed email, with headers set as above, that contains non-ascii characters (ones that would be transformed by `quoted-printable`
2. look at accentuated characters, see that they are malformed

**Error Log**
See description of the bug.

pazz:
Thanks for reporting this.
Would you mind sending a few problematic anonymized mails our way? Ideally in the form of a PR that adds (failing) unit tests. 
Email etiquette is unfortunately not very rigurous when it ocmes to encoding issues and fidly header syntax. Alot mostly uses pythons email module in order to keep standard compliant, but of course, lots of malformed mails make the rounds..

guijemont:
> Would you mind sending a few problematic anonymized mails our way? Ideally in the form of a PR that adds (failing) unit tests.

I'll see if I find time to do that over the week-end.

[...]

PR #1477, opened 2020-03-13
Allow optional text/plain mailcap entry.

ryneeverett:
My use case is limiting width:

    text/plain; fmt -w 80 -s %s

PR #1526, opened 2020-06-14
Fix utf8 encoding with a text/plain mailcap entry.

ryneeverett:
This seems to essentially revert 777823f414aab5dfa130174dc7c80cda8036d13f,
the reasoning of which I don't yet follow.

This may resolve the issue in #1522.

### DIFF
--- a/alot/db/utils.py
+++ b/alot/db/utils.py
@@ -429,11 +429,7 @@ def remove_cte(part, as_string=False):
     # decoding into a str is done at the end if requested
     elif '8bit' in cte:
         logging.debug('assuming Content-Transfer-Encoding: 8bit')
-        # Python's mail library may decode 8bit as raw-unicode-escape, so
-        # we need to encode that back to bytes so we can decode it using
-        # the correct encoding, or it might not, in which case assume that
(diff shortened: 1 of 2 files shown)


pazz:
I think that the email module's behaviour has changed since that ugly work-around was written, so good riddance. I'm surprised that our tests did not pick it up! Time for new tests?

ryneeverett:
This issue only exists if one has a text/plain mailcap entry (which apparently is commonly included in distributions). This comes down to the logic of the change I made in #1477. If there is no text/plain mailcap entry then the behavior does not change at all. If there is a text/plain mailcap entry, after #1477 the mail gets encoded and written to a temporary file where it is read by the mailcap program. This encoding step is where the bug lies.

As was pointed out in #1522, the mock I added on the plaintext test prevents this issue from emerging by emulating the behavior as if there were no text/plain entry.

Not sure if you noticed but this commit does add a test of utf8 with a text/plain entry.

pazz:
Thanks for the explanation. And yes, I did not notice the test :)
I'll push this as is. Cheers,

PR #1528, opened 2020-06-16
valign message to top of screen

### LINKED ISSUE #1491: Fit message on screen when setting focus with "move"

The "move" command sets the focus on the title line of a message in thread mode. If this line is already visible at the bottom of the screen, but not the message itself, the view isn't scrolling by itself to show the message.

The view should be scrolled to fit the message, or at least its beginning, in the viewport.

A dirty workaround that doesn't work for the last message in a thread is to rebind a key to "move next; move previous".

**Additional context**

/

mrichar1:
I think this is all that is needed to fix #1491. Tests seem to pass fine, but there's a chance it introduces unexpected behaviour. Not sure to what extent the test suite can accommodate checking things like screen positioning etc?

I'll see about running this branch for a bit (with `urwidtrees` from `master`) and see if I can create any bugs, especially with things like folding etc.

### DIFF
--- a/alot/buffers/thread.py
+++ b/alot/buffers/thread.py
@@ -180,7 +180,7 @@ class ThreadBuffer(Buffer):
     def set_focus(self, pos):
         "Set the focus in the underlying body widget."
         logging.debug('setting focus to %s ', pos)
-        self.body.set_focus(pos)
+        self.body.set_focus(pos, valign='top')
 
     def focus_first(self):
         """set focus to first message of thread"""
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
     install_requires=[
         'notmuch>=0.27',
         'urwid>=1.3.0',
-        'urwidtrees>=1.0',
+        'urwidtrees>=1.0.3',
         'twisted>=18.4.0',
         'python-magic',
         'configobj>=4.7.0',


pazz:
FYI: I've just made a release over at urwidtrees (with my other github handle but can't be bothered to rewrite the commits).
So If you'd amend the dependencies to urwidtrees >= 1.0.3 here, we can push this PR.

pazz:
The travis fail is expected for now, as urwidtrees is not updated on pip.

mrichar1:
Just saw the comment about this not working - weirdly it isn't working for me either - though I did have it working before making the commit... let me have a dig and see what I missed or did wrong between testing and committing! :thinking:

mrichar1:
OK - so it looks like this is my fault for not testing the behaviour fully with all the different actions!

It seems that if you open a thread (with all messages unfolded),then  `select` (default `enter` key) does the right thing and collapses the message and focuses the next one to the top. However if you use `move next` (default `space` key) then this doesn't happen. (I'm running from the git repo, without a local config file, but If you can double check and confirm this is the current behaviour for you too that'd be useful!)

I'll have a dig and see if I can figure out what happens in a `move` that is different in the code (and also consider if there are any other actions which need updated too).

PR #1533, opened 2020-07-02
Add commands for managing queries in namedqueries mode

brenns10:
I wanted to use namedqueries as my main startup mode. I hadn't used namedqueries before and wanted some easy management tools for them. So this adds the following commands:

* `query-rename NEWNAME`
* `query-duplicate NEWNAME` (newname defaults to oldname_copy, so this can be bound to other commands)
* `query-remove` is a shortcut for `removequery` global command, which deletes the currently highlighted query (after a prompt)
* `query-refine` opens a prompt where you can edit the `savequery` command for the currently highlighted command. This is useful because it allows you to edit both the query name, and the query itself. This is effectively both an "edit" command, as well as a "edit and save as" command.

This branch is only preliminary - I wanted to check if there were any general objections to the approaches, naming, etc. If not, then I'd like to add documentation and tests.

By the way, great project! I'm enjoying using it and hope to be contributing more.

### DIFF
--- a/alot/commands/namedqueries.py
+++ b/alot/commands/namedqueries.py
@@ -4,7 +4,12 @@
 import argparse
 
 from . import Command, registerCommand
+from .globals import ConfirmCommand
+from .globals import FlushCommand
+from .globals import PromptCommand
+from .globals import RemoveQueryCommand
 from .globals import SearchCommand
+from ..db.errors import DatabaseROError
 
 MODE = 'namedqueries'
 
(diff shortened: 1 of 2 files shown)


alot/commands/namedqueries.py
```diff
@@ -28,3 +33,89 @@ async def apply(self, ui):
...
+        ui.current_buffer.rebuild()
+
+
+@registerCommand(MODE, 'query-remove')
+class NamedqueriesRemoveCommand(Command):
+
+    def afterwards(self):
+        self.ui.current_buffer.rebuild()
+
+    """remove the selected namedquery"""
```
pazz, line 113: Codacy found an issue: [String statement has no effect](https://app.codacy.com/manual/patricktotzke/alot/pullRequest?prid=5804068)

PR #1535, opened 2020-07-06
Focus valign part 2 (#1528 continued)

mrichar1:
So `thread.py` was using `self.body.set_focus` 'directly' in various places. Changing all references to use `self.set_focus` as a wrapper ensures every call uses `valign='top'` and it focuses as expected for the other commands (`move next`, `move next|previous sibling` etc).

There's a minor annoyance remaining in that align-to-top doesn't happen for the last message(s) if their total length is less than the window height. The fix I suspect would be to add some padding to the end somehow, but this is probably a separate issue/feature.

### DIFF
--- a/alot/buffers/thread.py
+++ b/alot/buffers/thread.py
@@ -184,10 +184,10 @@ class ThreadBuffer(Buffer):
 
     def focus_first(self):
         """set focus to first message of thread"""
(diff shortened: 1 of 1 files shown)


pazz:
thanks guys

Issue #1536, opened 2020-07-15
Concatenate two searches, p.e. first all with "tag:flagged", then "tag:inbox NOT tag:flagged"


MaxGyver83:
**Is your feature request related to a problem? Please describe.**
Sometimes I flag an e-mail as a reminder to myself that I have to do something about it. But often I forget these e-mails as soon as they leave the first page when more e-mails arrive. Of course, I can filter for `tag:flagged` but then I need a reminder that reminds me to filter for my reminders. So it would be nice to have flagged e-mails always on top.

**Describe the solution you'd like**
An easy and generic solution would be an option to concatenate two (or maybe even more) searches. For example `:search tag:flagged, tag:inbox` (I don't know if a comma is a appropriate delimiter here). That way *alot* could do two (or more) *notmuch* searches and just concatenate the resulting lists.

**Describe alternatives you've considered**
An alternative, less generic solution could be an option like `show_flagged_emails_always_on_top = true`.

pazz:
Very nice idea!
Don't wait for me to implement it any time soon but I'd definitely push something like that if someone sends a PR.

PR #1537, opened 2020-07-15
Anable concatenation of search queries (split by " / ")

MaxGyver83:
Implement  #1536 

This PR enables the concatenation of two or more search queries. They will be split at ' / ' (a slash surrounded by spaces).

Example:

```
:search tag:flagged" / "tag:inbox NOT tag:flagged
```

This gives you first all flagged e-mails, then all e-mails tagged:inbox not repeating the flagged ones.

Possible problem: I don't know how you would search for a string containing `' / '`.


MaxGyver83:
Hi @pazz , I don't think that Travis CI failed because of my PR. The job log says:

```
./configure: 1558: ./configure: WITH_PYTHON_DOCS: parameter not set
```

pazz:
This was an issue with notmuch master. It is fixed now.
Could you squash the two commits?


Quoting Max (2020-07-16 19:40:19)

MaxGyver83:
> Could you squash the two commits?

Sure, they are squashed now.

pazz:
@lucc: opinions?

mmartin:
This is really wonky. It breaks my hooks (search returns double messages or no messages at all) and the result count is wrong.

alot/walker.py
```diff
@@ -6,6 +6,11 @@
 import urwid
 
 
+def chain(*iterables):
```
lucc, line 9: Is there a reason we can not use `itertools.chain` for this?

alot/walker.py
```diff
@@ -92,3 +97,6 @@ def _get_next_item(self):
 
     def get_lines(self):
         return self.lines
+
+    def append(self, iterableWalker):
```
lucc, line 101: What happens if this is called after somebody started iterating over the IterableWalker already? Is that a case we have to think about? Even if this method is trivial it might be worth it to add a docstring to mention such constrains (if there are  some).

alot/buffers/search.py
```diff
@@ -61,20 +61,26 @@ def rebuild(self, reverse=False):
...
-            self.ui.notify('malformed query string: %s' % self.querystring,
-                           'error')
-            self.listbox = urwid.ListBox([])
-            self.body = self.listbox
-            return
-
-        self.threadlist = IterableWalker(threads, ThreadlineWidget,
-                                         dbman=self.dbman,
-                                         reverse=reverse)
+        querylist = self.querystring.split(' / ')
```
lucc, line 64: What happens if one of the queries is empty? Currently the interface stops me if I start a completely empty query. So what happens with this code when I query for `foo / `?

alot/walker.py
MaxGyver83: Good point. I just didn't know it existed. I'm using `itertools.chain` now.

MaxGyver83:
> This is really wonky. It breaks my hooks (search returns double messages or no messages at all) and the result count is wrong.

@mmartin: Thanks for testing this PR. I have fixed the wrong result count. Could you give an example for a search that returns double messages?

mmartin:
@MaxGyver83 It's this one: https://github.com/pazz/alot/wiki/Contrib-Hooks#auto-refresh-search-buffer

alot/buffers/search.py
MaxGyver83: In this case the trailing space will be stripped. So you search for `foo /`, the query won't be split and get the same result as without this PR.

When you search for `foo / ''`, nothing will be stripped. Your query will be split and you get the result for `foo` (probably a non-empty list) plus the results for `''` (an empty list). I guess that's what you expect. Do you think, the user should see a warning here? Like `empty subquery string`!?

MaxGyver83:
> @MaxGyver83 It's this one: https://github.com/pazz/alot/wiki/Contrib-Hooks#auto-refresh-search-buffer

@mmartin: I could reproduce that behavior. Fixed in 5cf9bdf00493c29e3af9d879308e924630a20a98.

alot/walker.py
MaxGyver83: Good question. I'll try to test that case somehow.

alot/buffers/search.py
lucc: Well if it does not fail dramatically I am fine :)

MaxGyver83:
I couldn't find any search related tests. Probably because you need real e-mails for that.

MaxGyver83:
I don't know if slash is a good query divider. Maybe the word `THEN` is even better because it matches better the notmuch keywords. For example: `search tag:flagged THEN tag:inbox AND NOT tag:flagged`.

pazz:
I agree that adding tests should be good. 

Conceptually, what bugs me about this series is that it diverges from notmuch's query syntax, at least it looks like it does, which is something that I've so far tried to avoid.

One way to get around that would be not to have a single "meta query" but something like the 'refine' command, which allows to stepwise append result lists for individual, valid notmuch queries.
One could then concatenate such commands to have the same effect as proposed here..

MaxGyver83:
> One way to get around that would be not to have a single "meta query" but something like the 'refine' command, which allows to stepwise append result lists for individual, valid notmuch queries.
> One could then concatenate such commands to have the same effect as proposed here..

Sure, this sounds good to me. If I understand you right, you prefer something like `:search tag:flagged; append tag:inbox AND NOT tag:flagged`. Or maybe `append-search`. I'll update my PR. Can you already think of a command name that you like?

MaxGyver83:
I started working on the requested change and it looks like it's a big one!

Adding a new command to `commands/search.py` is easy. But the next logical step would be to replace `self.querystring` in `buffers/search.py:SearchBuffer` with a list of querystrings. So the new `append` command could add a new query to that list and then rebuild the current buffer. But unfortunately the querystring is used in many places, p.e. when checking if there is already an open buffer with the same query or when using the refine command. We need to define a new behavior for these, too. For example, we check instead if there exists already a buffer with the same list of queries and the `refine` command will only be applied to the last query in that list.

For me it's okay to finish this task but first I want to check if you are okay with such extensive changes to the code. @pazz: What do you think about this?

Issue #1539, opened 2020-08-07
notmuch module name change in notmuch-0.30


GuillaumeSeren:
Hey,
I was working on the alot-0.9.1 package for gentoo,
and we discovered this issue.

https://github.com/gentoo/gentoo/pull/16680

It is explained in the release doc of notmuch
https://notmuchmail.org/news/release-0.30/
```
Introduce new CFFI-based python bindings in the python module named "notmuch2". Officially deprecate (but still support) the older "notmuch" module.
```

It look like the 'still support' mean an immediate module name change,
as you can see in this issue https://bugs.gentoo.org/736204

So if you confirm the issue,
I can work on a patch.

pazz:
Wait, no: "notmuch2" is indeed the new CFFI-based module but that is not downward compatible.
ATM alot still needs the old "notmuch" module.

The notmuch2-port of alot is here: https://github.com/pazz/alot/pull/1511
(It is somewhat stalled) but needs pushing fwd.
P

GuillaumeSeren:
Hey @pazz 
thank you for the quick response.

I am not sure that the module is still in the source of the notmuch-0.30 ( but I may be wrong),
it is discussed on the issue about it here https://bugs.gentoo.org/736204

For alot-0.9.1I would like to not hard fix the dependency to <=notmuch-.30.

ff2000:
> I am not sure that the module is still in the source of the notmuch-0.30 ( but I may be wrong),

The old bindings are still in the git tree:
https://git.notmuchmail.org/git?p=notmuch;a=tree;f=bindings/python;h=2143d63b3687aaee8f32fe901f20c05b836a3cd6;hb=HEAD

I also checked the source tarball: the old "notmuch" bindings are still there.

GuillaumeSeren:
@ff2000 thank you for the update.
That is really interresting I will investigate where the problem is.

mjg:
For the record, this is solely a packaging problem.

The new bindings have a new name *on purpose*, they can coexist happily with the legacy bindings. Packagers can choose to package either - the best choice seems to be individual subpackages so that depending packages can depend on the needed binding only.

For a short time, gentoo shipped the new binding only, but this got fixed. The Fedora notmuch packager shipped both subpackages right from the start, of course :redhat-smiley:

So, there is no name-change to react upon and no alot bug here (porting to notmuch2 bindings notwithstanding).

GuillaumeSeren:
Hey,
[...]

PR #1511, opened 2020-05-18
Port to the `notmuch2` bindings

lenormf:
I migrated the code to use the `notmuch2` bindings. Keep in mind:

* I haven't tested the code beyond running `setup.py test`
* named queries are not supported (no sure how they're handled by the new bindings)
* I don't send emails much from Alot (there might be dead paths in my execution flow)
* more commits will be pushed on this branch as I fix bugs

I'll run this for some time to fix things I might have missed. Ultimately, the commit history is not very relevant and can be squashed, when more testing has been done.

PR #1540, opened 2020-08-08
Narrow down notmuch version in dependency

lucc:
This should be fixed by #1511. It might make issues like #1539 more clear, **I made this PR in the assumtion that we really require <0.30, that is what I gatherd from the comments there.**  If we do not really require <0.30 we should obviously ditch this PR.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ setup(
             ['alot = alot.__main__:main'],
     },
     install_requires=[
-        'notmuch>=0.27',
+        'notmuch>=0.27,<0.30',
         'urwid>=1.3.0',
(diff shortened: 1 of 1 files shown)


pazz:
I have just looked at  notmuch master.

1. "our" old notmuch bingings are still shipped, and the python module is called `notmuch`.
2. the new bindings also come with notmuch per default and are part of the repo, and the module is called `notmuch2`.

So for now we do not need any changes (until they remove the old bindings, which I'd veto). Once we push #1511, we should obviously update our dependencies, and let upstream know that they can deprecate the old bindings.

PR #1561, opened 2021-02-08
fix move first and last

paretje:
After rebuilding the threadlist, we still need to actually update the focused thread. Currently, in a buffer with more than 200 results you need to run the command twice to actually change the selected thread.

Would you accept a PR to make this limit of 200 results configurable? I personally find it annoying that the ordering of the threads is incorrect because of this, at least according to the original order, as `oldest-first` is not the reverse of `newest-first`. I currently use `move last; refresh` as a best-effort workaround.

### DIFF
--- a/alot/buffers/search.py
+++ b/alot/buffers/search.py
@@ -115,6 +115,7 @@ class SearchBuffer(Buffer):
             self.body.set_focus(0)
         else:
             self.rebuild(reverse=False)
+            self.body.set_focus(0)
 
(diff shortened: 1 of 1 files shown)


pazz:
Thanks for this!
I will certainly accept a PR to make this limit configurable, sure.
Just note that if you're introducing new config options you will have to re-generate the docs and check in the updated docs as well
(this should be addressed some time.. note to self).

PR #1562, opened 2021-02-09
configurable limit on move last

paretje:
As discussed in #1561, I added an option to configure the maximum number of results before using the reverse order heuristic for `move last`. I kept the default at `200`, although `search_threads_rebuild_limit` is unlimited by default, as we do have the heuristic approach as a fallback.

I also disabled restoring focus after the building the reversed threadlist, as that is limited by `search_threads_rebuild_limit`, making this heuristic pointless by default, as you still load the whole threadbuffer, while getting an inexact order. And even when `search_threads_rebuild_limit` is set to a different value, it would still be unnecessarily slow.

### DIFF
--- a/alot/buffers/search.py
+++ b/alot/buffers/search.py
@@ -46,7 +46,7 @@ class SearchBuffer(Buffer):
         info['result_count_positive'] = 's' if self.result_count > 1 else ''
         return info
 
-    def rebuild(self, reverse=False):
+    def rebuild(self, reverse=False, restore_focus=True):
         self.isinitialized = True
         self.reversed = reverse
         selected_thread = None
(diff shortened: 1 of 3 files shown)


paretje:
Regarding the Codacy failure: the `SearchBuffer.rebuild` function signature already differed before, supporting the `reverse` keyword argument. I guess I could rename `rebuild` to `_rebulld`, and defining `rebuild` to call `_rebuild`, and use the internal `_rebuild` when we want some specific behaviour while performing a jump.

pazz:
Thanks! I think this is a great solution. I was not able to talk codacity into ignoring the issue. IMO that issue is unproblematic because if ever called from outside contexts, the one or zero-parameter invocation still works as expected because you added a sensible default value to the new parameter.

PR #1565, opened 2021-03-19
fix synchronisation of tags to Maildir flags

paretje:
Since the move to notmuch2, tag changes were no longer synchronised to Maildir flags.


pazz:
Great! Many thanks!

PR #1566, opened 2021-03-19
Clarify exclude tags

paretje:
Clarify documentation on `exclude_tags` to make it more clear the notmuch `exclude_tags` setting is not used in notmuch. And cleanup of some old unused code loading the notmuch setting.

### DIFF
--- a/alot/buffers/search.py
+++ b/alot/buffers/search.py
@@ -57,14 +57,10 @@ class SearchBuffer(Buffer):
         if self.threadlist:
             selected_thread = self.get_selected_thread()
 
-        exclude_tags = settings.get_notmuch_setting('search', 'exclude_tags')
-        if exclude_tags:
-            exclude_tags = [t for t in exclude_tags.split(';') if t]
-
         try:
             self.result_count = self.dbman.count_messages(self.querystring)
             threads = self.dbman.get_threads(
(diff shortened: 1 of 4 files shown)


pazz:
Can you elaborate on the doc change? The alot setting is independent of the one for notmuch: you could have exclude tags only for alot but not notmuch or vice versa?

paretje:
Well, I got confused I guess. Somehow it wasn't clear to me that I should set this. But maybe I should drop the commit, in the end it is the result of me wondering why exclude tags didn't work any more, as alot used to use the notmuch `exclude_tags` (as well?).

pazz:
As far as I can see, the code that your first commit reverts actually makes alot use/respect the exclude tags defined in notmuch's config doesn't it?  I do not remember notmuch renaming or moving this setting into the index (like they did for the named queries at some point).

paretje:
The code to load the setting was still there, it was passed to `get_threads`, but it wasn't used. Fixing that would create another issue, as the actual number of results would differ from the count. If notmuch's `exclude_tags` is respected, we should do so consistently. I'm willing to implement that.

pazz:
That sounds like a good plan to me.
Thanks very much :)


Quoting Kevin Velghe (2021-03-25 11:49:36)

PR #1567, opened 2021-03-21
New buffer "directories"

mwtkrayer:
I added a new buffer "directories" for myself which allows browsing mail by underlying directory structure in the notmuch database directory. It is heavily based on the "namedqueries" buffer, but constructs "path:" queries. Unfortunately, the directory structure seems not to be easily deducible from the notmuch database itself, so I rely on `os.walk(self.path)` to construct it. I am exclusively working with "Maildir" for which this approach works fine, however, i did not test it for other storage formats.

### DIFF
--- a/alot/__main__.py
+++ b/alot/__main__.py
@@ -21,7 +21,7 @@ asyncioreactor.install()
 
 
 _SUBCOMMANDS = ['search', 'compose', 'bufferlist', 'taglist', 'namedqueries',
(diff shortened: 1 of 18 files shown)


mjg:
You could use xapian to list all terms with prefixes`P` resp. `XFOLDER:`. On the other hand, notmuch's own shell completion simply walks the filesystem, too.

PR #1568, opened 2021-03-23
fix fold and unfold commands

### LINKED ISSUE #1496: `fold *` sometimes doesn’t actually fold
**Describe the bug**

When in thread view, running `fold *` sometimes does not fold all messages.

**Software Versions**
- Python version: 3.7.7
- Notmuch version: 0.29.3
- Alot version: 0.9, master

**To Reproduce**

1. Unpack somewhere this [maildir](https://github.com/pazz/alot/files/4502615/alotrep.tar.gz) (it contains 1 thread of 3 messages).
2. Make a new notmuch config with a path to this maildir.
3. `export NOTMUCH_CONFIG=/path/to/config && notmuch new`.
4. Only tag the middle message: `notmuch tag -inbox '*' && notmuch tag +inbox id:serokell/serokell.nix/pull/2/review/396149515@github.com`
5. Run `alot` with empty config.
6. Open thread.
7. Type `:fold *<enter>`.

Expected: all messages fold.
Actual: the second message folds, two others unfold. If you type `:fold *<enter>` again, all fold.

**Error Log**
[logfile.gz](https://github.com/pazz/alot/files/4502621/logfile.gz)

paretje:
Explicitly expand or collapse every message, like we do in `ThreadBuffer.unfold_matching` instead of only expanding or collapsing the matching messages.

Closes #1496

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -548,6 +548,12 @@ class ChangeDisplaymodeCommand(Command):
         self.mimepart = mimepart
         Command.__init__(self, **kwargs)
 
+    def _matches(self, msgt):
+        if self.query is None or self.query == '*':
+            return True
+        msg = msgt.get_message()
+        return msg.matches(self.query)
+
     def apply(self, ui):
         tbuffer = ui.current_buffer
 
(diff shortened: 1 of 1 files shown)


pazz (approved):
Looks safe to me and if it fixes an open issue I'm happy to push

PR #1574, opened 2021-04-25
use notmuch exclude_tags as fallback

paretje:
This changes alot's exclude tag setting to use the notmuch exclude_tags setting as a fallback when the alot setting isn't set, as discussed in #1566.


pazz:
Looks good to me.
The only thing that is somewhat out of touch with alot's code base is the use of `@property` tags.
I think these are newer than alot itself. In any case, I don't think we use those consistently yet and that may be confusing
for newcomers to this code. But all in all, I'd be happy to use them more so I'm not complaining :)
Thanks!

PR #1581, opened 2021-07-25
Immediately remove unread tag from one-message threads

po1vo:
If 'auto_remove_unread' is enabled, immediately remove unread tag from one-message threads.

### DIFF
--- a/alot/buffers/thread.py
+++ b/alot/buffers/thread.py
@@ -107,11 +107,18 @@ class ThreadBuffer(Buffer):
 
         if settings.get('auto_remove_unread'):
             logging.debug('Tbuffer: auto remove unread tag from msg?')
+
             msg = self.get_selected_message()
             mid = msg.get_message_id()
-            focus_pos = self.body.get_focus()[1]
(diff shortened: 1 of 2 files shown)


lucc:
Hi @po1vo is there a reason you closed #1582 and moved that commit here? It looks like these are two unrelated things, so it might be easier to discuss them in two PRs.

For the unread stuff: I see what you do, but can you explain why this is needed?

po1vo:
> Hi @po1vo is there a reason you closed #1582 and moved that commit here? It looks like these are two unrelated things, so it might be easier to discuss them in two PRs.

Sorry for the mess, i committed to the wrong branch. Regarding #1582, i realized my solution could be better since i found that notify() from alot/ui.py relies on escape key.

> For the unread stuff: I see what you do, but can you explain why this is needed?

It's just annoying to have to navigate within an email body to have alot mark a single email read. I realize it was done this way to wait until a user actually navigates to an unread email within a thread, but for single-email threads that's totally unnecessary.

PR #1585, opened 2021-07-28
Added an option to unfold only messages matching a query

po1vo:
When defined this option will make only matching messages unfold when opening a thread. Saves time rendering, avoids hacks like `select; fold *; ...` and improves ui/ux.

Sample use:
`thread_unfold_matching = "tag:unread"`


pazz (approved):
To me this looks safe and I agree that it could improve the user experience.
Thanks for the patch and apologies for the long delay

PR #1589, opened 2021-10-20
Message summary subjects

kbingham:
This series tackles presenting threads where a multi-patch series is sent with a cover letter, for instance when sent with git-send-mail.

The default presentation in the thread view hides the details required to identify the patch titles:
alot-without-message-summaries

With this development the individual patches can be seen to show their titles on the summary.

alot-with-message-summaries

To prevent adding a lot of noise, replies do not further repeat the message title. The implemenation here so far is just to check if 'subject.startswith(('Re:')) and ignore those ones which does not seem fool-proof (and may be entirely locale dependant based upon the senders) so I'm looking for advice on a better way to implement that.

I had wondered if comparing the Subject against a parent subject, and only printing if they differ would be better, but the prefixes that get added  by replies would still affect that too.

### DIFF
--- a/alot/db/message.py
+++ b/alot/db/message.py
@@ -26,7 +26,7 @@ class Message:
     It it uses a :class:`~alot.db.DBManager` for cached manipulation
     and lazy lookups.
     """
-    def __init__(self, dbman, msg, thread=None):
+    def __init__(self, dbman, msg, thread=None, parent=None):
         """
         :param dbman: db manager that is used for further lookups
         :type dbman: alot.db.DBManager
@@ -34,11 +34,15 @@ class Message:
         :type msg: notmuch2.Message
         :param thread: this messages thread (will be looked up later if `None`)
         :type thread: :class:`~alot.db.Thread` or `None`
+        :param parent: this message's parent (if it was a reply, and we have the
+                                              message it was a reply to)
+        :type parent: :class:`~alot.db.Message` or `None`
         """
         self._dbman = dbman
         self._id = msg.messageid
         self._thread_id = msg.threadid
         self._thread = thread
+        self._parent = parent
         try:
             self._datetime = datetime.fromtimestamp(msg.date)
         except ValueError:
@@ -114,6 +118,10 @@ class Message:
                 self._email = email.message_from_string(
                     warning, policy=email.policy.SMTP)
         return self._email
+
+    def get_parent(self):
+        """returns parent Message if this is a reply or None otherwise"""
+        return self._parent
 
     def get_date(self):
         """returns Date header value as :class:`~datetime.datetime`"""
--- a/alot/db/message.py
+++ b/alot/db/message.py
@@ -50,6 +50,11 @@ class Message:
         self._mime_tree = None  # will be read upon first use
         self._tags = msg.tags
 
+        try:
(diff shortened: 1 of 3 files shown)


pazz:
I'm not sure which implementation of "get_subject" is better; perhaps @mjg has an opinion. I am actually surprised that this isn't already there (is that patch you mention not in master?).

Ideally, the message summary would be configurable in the theme just like threadline is in search mode.
At the very least I believe we should have config settings to enable/disable whether subjects are displayed at all and if so, if they are re-interpreted as you suggest.

mjg:
The thing is that `git send-email` and some people do the "wrong thing": they specify both `In-Reply-To` and `References` when indeed the patches are not a reply to the cover letter, for example. OTOH I don't think there's a widely accepted standard for saying "related but not a reply".

Just for fun: A bank which I sent an e-mail "Re: some topic" to apparantly translated "Re" into "Reklamation" (complaint) automatically and redirected the e-mail to that department...

So, I think your best option to check whether e-mail 2 is a "real reply" to e-mail 1 is by checking whether e-mail 1's subject is *contained* in that of e-mail 2 (plus maybe a safety check on difference of length being smallish). This catches all kinds of "Re/RE/AW", and forwards typically do not set the headers, so they're not in the mix anyways.

For an extra safety measure, replace the matched subject by `...`, so that message 2 would display the subject as `Re: ...` or `AW: ...` or `Why do you write an e-mail with subject ... to me`? which loses none of the extra information but still is much shorter than the full subject. I think even I would like that.

Or let the user define a hook or regex for the match.

kbingham:
Now that I've fixed alot to run on master with my config, I've rebased this. Probably still a bit more refinement needed, but this really helps my work-load a lot.

kbingham:
Small updates here. The method of getting the subject has been simplified, as the subject headers are indexed by not-much - so from my understanding it's more efficient to simply get it when constructing the Message object, and retain it at that point.

I've also picked up a fix in this series, while testing as I sent myself a mail with no subject to validate this. And that hit a failure in existing code paths. So that's included as the last patch.

Open questions are, if this implementation for stripping the parent subject is satisfactory to all, and if this should be any kind of a config option... or if this should be default behaviour.

pazz:
Sorry for the slow progress on this; you  can see that maintaining this project has low-priority for me personally but I really appreciate people continuing to use it (like myself) and pushing progress!
I have fired up a copilot review just in case, because I doubt I can give you much useful feedback myself. The feature looks very nice though and I am happy to push it if all is fine..
Thanks :)

pazz:
@kbingham please consider the copilot review comments and implement or discard them as you see fit. I am happy to go ahead and push afterwards :)

kbingham:
> Sorry for the slow progress on this; you can see that maintaining this project has low-priority for me personally but I really appreciate people continuing to use it (like myself) and pushing progress!

No problem at all - I certainly know what it's like to maintain an open source project.

alot "fixed" so many of my workflow issues dealing with all the mailing lists I have to handle, and I even got others in my team using it, so if there's anything we can do to help support you let me know.

> @kbingham please consider the copilot review comments and implement or discard them as you see fit. I am happy to go ahead and push afterwards :)

Thanks - I'll work through those when I can and update here.

Issue #769, opened 2015-05-29
alot refreshes search list during execution of command list


MacGyverNL:
I want to be able to delete mail, and I also want to be able to undelete mail and have it marked as undeleted so that I can check them and process them later. In order to have those tags mutually exclusive, I have added the following to my .notmuch-config:
```
[bindings]
    [[search]]
        delete = "tag deleted; untag undeleted,unread,inbox;"
        insert = "tag undeleted; untag deleted;"
```

Now, in normal conditions, this works. But I have also redefined
```
initial_command = search tag:inbox AND NOT tag:killed AND NOT tag:deleted
```
and
```
[bindings]
    I = search tag:inbox AND NOT tag:killed AND NOT tag:deleted
```

Now, this together makes that whenever I press delete on a selected mail in the initial search, it gets tagged deleted, the search gets updated which makes this mail disappear and the next mail selected, and then that next mail gets untagged undeleted, unread, inbox, making it disappear as well. This was not what I was expecting to happen.

In my opinion, the expected behaviour for this should always be to keep the same mail selected while executing the commands, so I would propose to only refresh the search after complete execution of a command list. I realize that people could want to have command lists which update the search halfway inbetween, but in these cases they could just insert a `refresh` command at the desired locations.

pazz:
Quoting Pol Van Aubel (2015-05-29 08:38:55)
>    I want to be able to delete mail, and I also want to be able to undelete
>    mail and have it marked as undeleted so that I can check them and process
>    them later. In order to have those tags mutually exclusive, I have added
>    the following to my .notmuch-config:

alot config you mean?

>    In my opinion, the expected behaviour for this should always be to keep
>    the same mail selected while executing the commands, so I would propose to
>    only refresh the search after complete execution of a command list. I
>    realize that people could want to have command lists which update the
>    search halfway inbetween, but in these cases they could just insert a
>    refresh command at the desired locations.


The thing is that the refresh-command will actually rebuild all widgets in the
buffer, and since the focus position is a property of these widgets, it will be
lost. But yea: it'd be good to "fix" this behaviour of refresh..

MacGyverNL:
Or add another command like `refresh-search` in the search context and leave `refresh` as-is.

MacGyverNL:
[...]

PR #1592, opened 2021-11-28
Partially addresses 'alot refreshes search list during execution of command list' #769 

naegling:
Both the OP and my own issue involved atomically adding and removing tags that could effect the contents of search.  This patch adds a new command `applytags` with arguments formatted as in `notmuch tag` command.

The OP's example:
```
[bindings]
    [[search]]
        delete = "tag deleted; untag undeleted,unread,inbox;"
        insert = "tag undeleted; untag deleted;"
```
could be configured as:
```
[bindings]
    [[search]]
        delete = "applytags +deleted -undeleted -unread -inbox;"
        insert = "applytags +undeleted -deleted;"
```


naegling:
Well, that's annoying.  I just noticed that the first argument to applytags has to be an added tag.  Otherwise, its interpreted as a (invalid) command line option.

pazz:
hehe yes, lols. never thought of that :D

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
kbingham: I don't think this file should be added.

kbingham:
This is a really nice idea, I've also hit the tags affecting the wrong thread after a refresh.

.vscode/launch.json
naegling: I've removed the .vscode file accidentally added.

naegling:
> Well, that's annoying. I just noticed that the first argument to applytags has to be an added tag. Otherwise, its interpreted as a (invalid) command line option.

Suppressing the option interpretation looks both invasive and hacky.  Currently, I see 2 options for more consistent syntax. to remove the dick tag and add the jane tag:
1. use a standalone + as a noop.  `applytags + -dick +jane`.
2. introduce an option name.  `applytags -ops -dick +jane`. The option name could be optional, so that `applytags +jane -dick` would still work as before.

Comments Welcomed!

PR #1593, opened 2021-12-10
commands: thread: Reference existing message when forwarding

kbingham:
When forwarding a mail, the outgoing mail, and any follow up
conversation becomes a disjointed thread.  This removes the original
context, and makes it harder to find and track while the conversation
splits in two directions.

Add appropriate reference headers to the envelope when constructing a
forwarded mail to ensure that the conversations continue to be related
and therefore presented in the same thread with notmuch and alot.

Signed-off-by: Kieran Bingham <kieran.bingham@ideasonboard.com>

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -359,6 +359,10 @@ class ForwardCommand(Command):
                 subject = fsp + subject
         envelope.add('Subject', subject)
 
+        # Set forwarding reference headers
+        envelope.add('References', '<%s>' % self.message.get_message_id())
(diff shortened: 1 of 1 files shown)


kbingham:
I think the 'references' header probably fixes my issue and keeps it in the correct thread, but looking at other mails I had received as forwards, I saw other clients adding the X-Forwarded-Message-Id: header, so I've introduced that as well. This may or may not be desired though.

pazz (approved):
I agree with this: joining these threads in the UI is better from a user perspective in my opinion and the patch you suggest has this effect for messages going forward.

Issue #1588, opened 2021-10-17
change colorscheme at runtime ?


teto:
**Is your feature request related to a problem? Please describe.**
Maybe I've missed it but I would like to change the colorscheme at runtime rather than edit the config file and restart alot.

**Describe the solution you'd like**
I would like to be able to run
`:theme my_theme`
Ideally with some autocompletion

dylanfprice:
For anyone interested, now that this is implemented I am automatically switching themes with OS light mode/dark mode: https://github.com/pazz/alot/wiki/Contrib-Hooks#automatic-dark-mode

PR #1603, opened 2022-06-06
feat: add a 'theme' command

teto:
to be able to change the theme dynamically see https://github.com/pazz/alot/issues/1588

When running the command in development mode, I get:
`Could not read toto and/or /home/teto/alot/alot/settings/../defaults/theme.spec`

What is weird too is that starting alot in development mode takes 5 sec while the system one starts <1 sec . The system alot took longer the first time though so I wonder if alot generates some cache along the way ? Same happened with neomutt, took a long time starting the first time I ran it in a while then successive starts were much faster. I've got 136675 messages in DB and ran a notmuch compact.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -35,6 +35,7 @@ from ..db.envelope import Envelope
 from ..settings.const import settings
 from ..settings.errors import ConfigError, NoMatchingAccount
 from ..utils import argparse as cargparse
+from ..settings.theme import Theme
 
 MODE = 'global'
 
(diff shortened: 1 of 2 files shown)


alot/commands/globals.py
```diff
@@ -1088,6 +1089,28 @@ def apply(self, ui):
             ui.notify('Error when reloading config files:\n {}'.format(e),
                       priority='error')
 
+@registerCommand(MODE, 'theme',     arguments=[
+    (['theme_name'], {'help': 'Name of the theme'})], help='Select your own them')
+class ThemeCommand(Command):
+
+    """move in widget"""
+    def __init__(self, theme_name=None, **kwargs):
```
pazz, line 1097: this does not do anything so could be omitted? the docstring is also not correct is it?

pazz (changes requested):
It looks safe but needs cleaning up.

pazz:
I have not yet tested this but I agree that this feature would be nice to have.
I do not know why you see delayed startups in dev mode. perhaps printing the extra debug messages takes time due to some file system issue? There is certainly no caching going on as far as I am aware.

teto:
It doesn't work yet for some reason, the error message `Could not read toto and/or /home/teto/alot/alot/settings/../defaults/theme.spec` seems wrong: the file exists so it's probably that it cant "load" it instead. Gonna troubleshoot

teto:
hum I was looking at how to pass CI but I notice none of the current PRs pass CI: is that even doable ?
I was hoping to run the checks locally too but this seems to be all SAAS ?

Issue #1560, opened 2021-01-08
Edit mail in a different terminal tabs than the main alot windows


psic:
**Describe the bug**
I don't know if it is really a bug, a misconfiguration by myself or ... a feature request.
I'd like to write my email (create or respond) in a different (new one for each mail) terminal tab than the main tab of alot. It's pretty easy to setup for opening or creating a mail in a new tab with  the header, and edit the mail with vim or nano. I try with tmux, screen , kitty and other terminal. Each time,  whatever the terminal I used, when I close the editor (and the mail is really in the temp file) and normally come back to main windows, the previously edited mail is empty/not showed.

Subsidiary question, is there any way to ask for configuration tips ? any mailing list ? reddit sub or else ?

**Software Versions**
- Python version:3.6.9
- Alot version: 0.9.1

lucc:
Can you post your config (especially `terminal_cmd`, `editor_in_thread`, `editor_spawn`, `editor_cmd`/`$EDITOR`) so that we can better understand your question.

I would assume that you need to set `terminal_cmd` to something like `myTerminal --open-new-tab --forground --execute`. The foreground part is important. If the terminal command that alot executes returns right after it was executed and the terminal window/tab runs in a background process then alot thinks you are done editing already (but you did not even start yet).

Did you have success with new terminal windows (instead of tabs)? In this case you might be able to see the main alot terminal and your editor terminal side by side and spot some hints on what is going on in which order.

lucc:
There is the `#alot` IRC channel on freenode.net, I do not know of any other forums or mailing lists.

psic:
Your assumption is right, but I was aware about the `foreground` part.
`editor_in_thread` and `editor_spawn` are always set to True.
`editor_in_thread=True`
`editor_spawn=True`
`editor_cmd="nano -Y alot -m +4"`
To start nano with an alot custom style, with the cursor on the fourth line, just under the header.
`terminal_cmd="screen -S alot -X screen " ` for screen using st as terminal
`terminal_cmd="kitty @ launch --type=tab"` for the kitty terminal
`terminal_cmd="tmux new-window "` for tmux using st as terminal
No problem to work with terminal in new window.

lucc:
[...]

Issue #460, opened 2012-05-31
editor_in_thread is not honored


xunam:
Hi,

There seems to be a problem with the config option "editor_in_thread". When I set my config as

```
editor_in_thread = True
editor_spawn = False
editor_cmd = "gvim -f"
```

alot opens GVim in blocking mode, effectively ignoring the "editor_in_thread" option.

I tried to run the terminal version of Vim in a spawned terminal using

```
editor_in_thread = False
editor_spawn = True
editor_cmd = vim
```

then alot spawns a terminal which runs Vim, and alot is not blocked. However, I immediately get the envelope view with an empty message (only including the signature) and the editor shows me an empty file (not including any header), as if the temporary file had vanished. This behaviour also does not depend on the value of "editor_in_thread". [edit: this is a problem with how `gnome-terminal` behaves, as described below]

In the end, I cannot edit a message and view other messages at the same time, although that would be a really interesting feature...

This applies to the current master version.

PR #1605, opened 2022-09-13
Fix external tmux command without display

MacGyverNL:
I spent entirely too long today fixing my own instance of #1560, because I'm not running alot in X11 and #460 and 2a7d1fd2dcc7354b06faa0da5574324b3578eba1 make alot simply fall back to non-spawning behaviour silently.

Initially I special-cased tmux outside of X11 in alot/commands/globals.py, but then I found out that I'm supposed to write a hook. So I did that, but I'd like to document it publicly.

There's also a bug where the `thread` result of that hook is not respected because `self.in_thread` gets set directly.

- Ensure the `thread` result of the `touch_external_command` hook is not overridden.
- Provide an example `touch_external_command` hook and corresponding shell script for tmux outside of X11. (Thanks to @cjpbirkbeck in https://github.com/pazz/alot/issues/1560#issuecomment-907222165)
- Log a warning if spawn is requested outside of X11 but no hook is set, rather than fall back silently to non-threaded mode.

### DIFF
--- a/alot/commands/globals.py
+++ b/alot/commands/globals.py
@@ -221,7 +221,7 @@ class ExternalCommand(Command):
             logging.debug('calling hook: touch_external_cmdlist')
             res = touchhook(cmd, shell=shell, spawn=spawn, thread=thread)
             logging.debug('got: %s', res)
-            cmd, shell, self.in_thread = res
+            cmd, shell, thread = res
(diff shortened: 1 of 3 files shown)


pazz:
Looks good to me. thanks!

Issue #777, opened 2015-07-14
Handle calendar content of an email


joeda:
Calendar entries are not handled at the moment. They look like this

```
--_000_8829A0A5DFB6FB4195BC72E26DEACC8601C1136AD89DKITMSXasdfe_
Content-Type: text/calendar; charset="utf-8"; method=REQUEST
Content-Transfer-Encoding: base64

base64encodedstuff
--_000_8829A0A5DFB6FB4195BC72E26DEACC8601C1136AD89DKITMSXasdfe--
```

in an email and contain iCalendar-data [1]. They could either be handled as a regular attachment for users to import them into their calendar or decoded and displayed, even though I think the former solution is preferrable since alot is not a PIM.

[1] https://en.wikipedia.org/wiki/ICalendar

jake-low:
+1 on this.

You can't currently do this using `.mailcap` because alot always chooses exactly one of the `text/*` parts of a multipart message to process (last I checked the source code at least).

The simplest fix for this in my opinion is to give alot an option that when set to true would show all unrendered parts of a multipart message as attachments. That way you could save `text/calendar` parts to your calendar directory.

lucc:
#894 might be related, sadly I am a little stuck on the widgets stuff. I hope to pick it up soon though.

PR #1607, opened 2022-09-14
Another attempt at handling text/calendar sensibly

MacGyverNL:
This is an attempt at fixing the following combination of long-standing issues:

1. Outlook sends calendar invites as `text/calendar` without a content disposition, so it doesn't show up as an attachment.

2. alot doesn't handle `text/calendar` as a mimepart without explicitly navigating to it via the mimetree.

3. *If* a user sets their mailcap to `text/calendar; cat '%s'; copiousoutput`, the calendar contents get displayed as expected, and may be processed using a pipeto command.

The combination of 1 and 2 make it so that `text/calendar` invites from outlook are effectively hidden by default in alot.

The existence of 3 makes it so that we can't simply turn `text/calendar` into attachments.

An early attempt at fixing this by allowing `text/calendar` to be a preferred mimetype was made in <https://github.com/pazz/alot/pull/640> but that seems to have been lost. However, we have to assume that the demonstrated workflow based on decoding with mailcap and using `pipeto` is in use.

<https://github.com/pazz/alot/pull/752> handles this situation by having multiple mailcap lines with different flags. This was merged but then removed in 7a883108b1bc6d8b59ac5a14ebf6265312135be2

Another idea, handling all unrendered parts as attachments, was proposed in <https://github.com/pazz/alot/issues/777>. Since we now have a mimetree, this is not *strictly* necessary, but it *would* be nice to be able to save the inline text/calendar as a .ics-file.

Combining these ideas, this proposal does three things:

1. Introduce an alot-specific x-alot-openattachment field for mailcap, that alot prefers when *opening* an attachment. If it is not present, alot will fall back to the standard ('view') field for opening the attachment.

2. Every `text/calendar` part without a content-disposition gets one forced on it, effectively turning it into an attachment. The resulting attachment is visible in the attachments list and can be handled as any other attachment: saved, and handled by mailcap. However, *opening* such an attachment via the attachment widget *will not display it* in the current proposal. Since these weren't attachments before, there's no need to handle them backwards-compatible with issue 3 above.

3. When a part gets selected via the mimetree, alot checks whether there is a separate `x-alot-openattachment` field defined in the mailcap. If there is, alot will first attempt to display the attachment as mimepart, and then open the attachment as per 2. I expect this to be the most controversial change.

Example mailcap entry:
```text/calendar; cat '%s'; x-alot-openattachment=tmux-horizontal-split-blocking.sh khal import '%s'; copiousoutput```

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -601,6 +601,11 @@ class ChangeDisplaymodeCommand(Command):
                     mimepart = get_body_part(message.get_email(), mimetype)
                 elif self.mimepart is True:
                     mimepart = ui.get_deep_focus().mimepart
+                if isinstance(mimepart, Attachment):
+                    # We're here because we're processing this attachment using a
+                    # mailcap view entry to actually display within alot.
+                    # Therefore, extract the mimepart from the Attachment here.
+                    mimepart = mimepart.part
                 mt.set_mimepart(mimepart)
                 ui.update()
             if self.mimetree == 'toggle':
@@ -919,7 +924,7 @@ class SaveAttachmentCommand(Command):
             focus = ui.get_deep_focus()
             if isinstance(focus, AttachmentWidget):
                 attachment = focus.get_attachment()
-                filename = attachment.get_filename()
+                filename = attachment.get_filename() or ''
                 if not self.path:
                     msg = 'save attachment (%s) to ' % filename
                     initialtext = os.path.join(savedir, filename)
@@ -958,7 +963,8 @@ class OpenAttachmentCommand(Command):
             afterwards = None  # callback, will rm tempfile if used
             handler_stdin = None
             tempfile_name = None
-            handler_raw_commandstring = entry['view']
+            handler_raw_commandstring = entry.get('x-alot-openattachment', '') or entry['view']
+            logging.info("Got {}".format(handler_raw_commandstring))
             # read parameter
             part = self.attachment.get_mime_representation()
             parms = tuple('='.join(p) for p in part.get_params())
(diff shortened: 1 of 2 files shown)


MacGyverNL:
Having thought about this some more, the more sensible solution is probably to define an `x-alot-attachmenttotext` for mailcap and changing the mime-display codepath, so that mailcap's default handler is the "proper" system handler. I'll rewrite this to take that approach.

MacGyverNL:
So the current proposal is as follows:

- Assume that the default entry is set up to handle attachments *as attachments*,
i.e. open them in whatever application is supposed to handle them.

- Define x-alot-attachmenttotext that provides a way to render
the attachment as text. If defined, opening the attachment through the
mime tree will both show it as text *and* open it. If not defined,
opening the attachment through the mime tree will only open it.

- When opening the attachment through the attachment widget, it's still
only opened even if an x-alot-attachmenttotext handler is defined.

There is *no* way to *only* render the attachment (at the moment). I
don't really have a usecase for this myself, but perhaps we could gate
the mimetree "open attachment" action behind a prompt if the attachment
is rendered using an x-alot-attachmenttotext handler?

Example mailcap entry:
`text/calendar; tmux-horizontal-split-blocking.sh khal import '%s'; x-alot-attachmenttotext=cat '%s'; copiousoutput`

MacGyverNL:
At the time I didn't feel up to messing with that part of the code, iirc.

I'm afraid that work circumstances have steered me back to using Thunderbird and plain Maildir for my mail, so I no longer daily-drive alot. For me it's therefore become a non-issue.

Issue #1531, opened 2020-06-21
ANSI codes in replies to HTML mails


mjg:
**Describe the bug**
Replying to HTML messages produces ANSI codes

**Software Versions**
- Python version: 3.7.7
- Notmuch version: 0.29
- Alot version: 530addcb

**To Reproduce**
Steps to reproduce the behaviour:
1. Use the suggested `text/html`-entry in `mailcap` (elinks) to produce ANSI codes 
2. reply to HTML message
3. observe the quoted text:
```
XYZ venit, vidit, dixit 2020-06-21 14:26:49:
> ^[[38;5;0m^[[48;5;0m^[[38;5;15m   ^[[38;5;250mLieber Herr Gruber,
```

I might be missing some additional config to convert ANSIfied HTML back to txt or so, but ANSIfied bofy text certainly is *not* intended to be quoted in replies as is.

pazz:
Ah yes, that makes sense.
What you want in this case are two different html->text converters. One for displaying in alot's UI and one for creating the reply template.
I'm not sure how to implement this nicely (in terms of user configs)..
Suggestions welcome!

johslarsen:
I have the same problem. My far from perfect, but still usable workaround is to stip away ANSI escape codes hook for quoting text. This can be achieved with a hook like:
```
ANSI_ESCAPE = re.compile(r'\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/ ]*[@-~])')
def text_quote(message):
  ANSI_ESCAPE.sub('', message)
```

mjg:
Thanks for the hint.

My workaround is to avoid the feature alltogether since `interpret_ansi_background = false` does not catch those cases where some combination of stupid styles and elinks mode produces blinking text (for the whole e-mail body), which is completely nuts. Yes, I know this is foreground, but for me there are too many issues (startup performance, blinking foreground, escapes in replies) with ANSItext and too little time to hunt them down (in config or code).

I'm just wondering if there's anyone **not** having a problem with replies to HTML e-mails when using ANSI conversion - this appears to be by design, not by coincidence in specific configurations.

pazz:
I agree that the conversion from html to plaintext and back is error-prone, which is why I also don't use this feature. Unfortuntely we cannot assume that all html emails have sensible plaintext alternatives to fall back on.
I am also quite annoyed by the inlined links in text created by elinks.
[...]

PR #1611, opened 2023-04-06
Fix some issues regarding ANSI CSI handling

guludo:
When replying to an email containing ANSI CSI escapes, those sequences are currently included in the reply's quoted message. This PR makes sure they are removed. While working on it, I realized some other issues in the generation of urwid attributes and added fixes for those as well.


guludo:
Force-pushed to fix style.

guludo:
Went through existing issues and realized that this fixes #1531.

pazz (approved):
Very nice! I always wondered where these extra codes came from but never properly investigated. Thanks for the very clean code and commit messages. I agree with all changes made here!

PR #1612, opened 2023-04-12
commands/search/tag: Use only thread id in testquery

guludo:
This is specially necessary if one is chaining a series of tagging commands.

As an example, without this fix, in the chain "untag inbox,unread --no-flush; tag foo,bar", the second command would not work properly if the current search buffer's query string was "tag:inbox and tag:unread".


pazz (approved):
Looks good to me. Thanks!

PR #1615, opened 2023-04-14
alot/settings: Use notmuch config to read configuration

guludo:
The main benefit from this is that we get the defaults directly from
notmuch and do not risk using inconsistent configuration values.

A good example of this is getting the database path. Getting the path
directly from the notmuch config file is not enough, because the user
might be using the default location with no explicit entry in the file.
Since notmuch does is a bit of logic notmuch to get the database path
(see section "DATABASE LOCATION" in man page for notmuch-config(1)), it
is better to get it directly from "notmuch config".

Since get_notmuch_setting() now always returns a string, also refactor
the call to settings.get_notmuch_setting('maildir', 'synchronize_flags')
for this new reality.

### DIFF
--- a/alot/db/manager.py
+++ b/alot/db/manager.py
@@ -72,7 +72,7 @@ class DBManager:
             raise DatabaseROError()
         if self.writequeue:
             # read notmuch's config regarding imap flag synchronization
-            sync = settings.get_notmuch_setting('maildir', 'synchronize_flags')
+            sync = settings.get_notmuch_setting('maildir', 'synchronize_flags') == 'true'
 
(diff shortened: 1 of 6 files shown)


guludo:
This PR is addresses https://github.com/pazz/alot/pull/1610#issuecomment-1506483026

pazz:
Looks good. thank you.
I should really fix this travis CI thing, or better replace it by GitHub actions..

lucc:
The change in e500326 broke the test test_parsing_notmuch_config_with_non_bool_synchronize_flag_fails.

PR #1618, opened 2023-04-18
commands/search: Add option --all-folded to select

guludo:
Allow users who prefer to have all messages collapsed by default to have that behavior. For example, with this change, a user can change her default binding of "enter" to execute "select --all-folded".

### DIFF
--- a/alot/commands/search.py
+++ b/alot/commands/search.py
@@ -19,16 +19,24 @@ from ..db.errors import DatabaseROError
 MODE = 'search'
 
 
-@registerCommand(MODE, 'select')
+@registerCommand(
+    MODE, 'select',
+    arguments=[
+        (['--all-folded'], {'action': 'store_true',
+                            'dest': 'all_folded',
+                            'help': 'do not unfold matching messages'}),
+    ],
+)
 class OpenThreadCommand(Command):
 
     """open thread in a new buffer"""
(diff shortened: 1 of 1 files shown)


guludo:
Personal note: I prefer to have all messages in the thread closed and just open those I'm interested in. With this change, I have the binding `enter = select --all-folded` to get the behavior I want.

lucc:
Why not just
~~~
[bindings]
[[search]]
enter = select; fold *
~~~
I even had this  in my config for a while
~~~
enter = select; fold *; unfold tag:unread; move last; unfold
~~~

guludo:
Hm... I just tried `enter = select; fold *` but I end up with some message unfolded.

guludo:
Ping. Is this PR still under consideration or should I give it up? :-)

lucc (approved):
The code looks very simple and I quickly tested it on some thread with several messages.

As noted by @guludo my suggested a alternative of `select; fold *` seems to invert the folding of all messages in the thread so it folds all messages that match the search query but unfolds all others. Only `select; fold *; fold *` seems to fold all messages.

But as these workarounds are quite slow and work unreliable I would accept this PR. @pazz what do you think?

pazz:
Thanks for looking into this.
I do not object, let's go for it.
P

Quoting Lucas Hoffmann (2024-01-02 13:51:30)

PR #1619, opened 2023-05-09
commands/thread: Add option --strip_ansi to pipeto

guludo:
If someone has a mailcap entry that colors the mail content, it is possible that she would want to get the content without the CSI escapes when piping it an external command.

Ideally, there should be a way to have the content untouched by the mailcap filter so we could use it when it makes sense. However, let's use ansi.remove_csi() for the time being.


pazz:
Shouldn't this default to `true`? when would you *ever* want these escape codes when piping stuff into other tools?

guludo:
> Shouldn't this default to `true`?

I agree that would be a better default. I just wanted to make sure I didn't break other people's workflow. 

> when would you _ever_ want these escape codes when piping stuff into other tools?

I guess one use case is when using another program to view the email. One could, for example, pipe it to `less -R`.

Anyways, let me know if you prefer `true` to be the default, then I'll send a new version here.

pazz:
Can you explain the use case for this? I mean, I have never seen these sequences in actual message contents and only ever *added* them for displaying stuff on the terminal.

guludo:
For me, the use case was to apply a patch directly from the mail client.

pazz:
But do the patch files you receive really have any ansi characters in them?
That's what's confusing me here..


Quoting Gustavo José de Sousa (2023-07-26 14:38:50)

pazz:
Thanks for clarifying. I think there is something wrong with the pipeto command then:
I believe it should only send the content of the mime part, not what has been piped through the mailcap entry.
Will have to look into this at some point.
Cheers!

Quoting Gustavo José de Sousa (2023-07-26 18:40:12)

PR #1625, opened 2023-08-24
Support env variables in mail_container path

milhnl:
Hey! 

I noticed that environment variable expansion did not work with `draft_box`. This patch will add that, and for `sent_box` too. (Implemented it in `mail_container`). The change itself is just 1 line, the rest is documentation.

I'd use this to have the same configuration on macOS and Linux, because my mail directory is in different places on these systems. 

Any feedback is of course welcome!


pazz (approved):
Looks safe to me, thanks!

PR #1627, opened 2023-08-28
Add option to remove obsolete drafts

milhnl:
Hi all!

This is an updated version of https://github.com/pazz/alot/pull/1313 (which aimed to fix https://github.com/pazz/alot/issues/1044). I rebased the commit, and deduplicated the code in `SaveCommand` and `SendCommand`. I think that was the most obvious feedback then, but I can imagine there's more.

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -465,9 +465,15 @@ class EditNewCommand(Command):
         tags.difference_update({'inbox', 'sent', 'draft', 'killed', 'replied',
                                 'signed', 'encrypted', 'unread', 'attachment'})
         tags = list(tags)
+
+        draft = None
+        if 'draft' in self.message.get_tags():
+            draft = self.message
+
         # set body text
         mailcontent = self.message.get_body_text()
-        envelope = Envelope(bodytext=mailcontent, tags=tags)
+        envelope = Envelope(bodytext=mailcontent, tags=tags,
+                            previous_draft=draft)
 
         # copy selected headers
         to_copy = ['Subject', 'From', 'To', 'Cc', 'Bcc', 'In-Reply-To',
--- a/alot/commands/envelope.py
+++ b/alot/commands/envelope.py
@@ -112,6 +112,48 @@ class RefineCommand(Command):
         await ui.apply_command(globals.PromptCommand(cmdstring))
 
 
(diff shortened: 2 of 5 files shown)


alot/commands/thread.py
```diff
@@ -465,9 +465,15 @@ async def apply(self, ui):
...
+
+        draft = None
+        if 'draft' in self.message.get_tags():
+            draft = self.message
+
         # set body text
         mailcontent = self.message.get_body_text()
-        envelope = Envelope(bodytext=mailcontent, tags=tags)
+        envelope = Envelope(bodytext=mailcontent, tags=tags,
+                            previous_draft=draft)
```
pazz, line 476: I don't quite see how this is picket up by my editor then in envelope mode?
Am I missing something or is then envelope -> edit command simply ignoring this?

milhnl: Your `$EDITOR`? Is that necessary? `SendCommand` and `SaveCommand` simply pick it up from this field in the envelope.

milhnl:
By the way, I just remembered I saw a pull request today that I think was merged which made the tags to indicate a draft configurable by the user. That is not yet incorporated in this commit.

alot/commands/thread.py
pazz: Sorry, I was confused and only now understood the logic behind this patch.
Question: What if I answer "no" to the prompt to remove the draft, then create another draft and end up with two drafts, then send the last and say "yes" to the prompt. 
As far as I can see, the current logic would only remove the very last draft, not the first one?
If so, is that intentional?

I could not test this myself at the minute because I have problems installing the dev version with poetry.. :D

guludo:
I took a look at adding support for the draft_tags stuff here and it seems there are other places in the current code base that needs the same type of fix already (see the output of `git grep -W -e "'draft'" -e '"draft"' upstream/master`)...

pazz:
Sorry for the late response but what is the status of this?

milhnl:
No worries. I'm busy with some IRL stuff right now, but have an idea on how to fix the `draft_tags` situation. It boils down to also using the full list to select drafts, but it is probably clearer as a commit. I think I can get y'all revised pull request somewhere in the coming month.

Issue #1318, opened 2018-10-04
Mailto via iTerm2 produces Errno 2


hgvhgv:
Before you submit a bug report, please make sure that the issue still exists on the master branch!

**Describe the bug**
I have iTerm2 on macos 10.13.6 set as the default handler for `mailto` links. It opens a new terminal window and runs the following command `/usr/local/bin/alot compose $$RES$$` with `$$RES$$` pulling the recipient's address from the URL. A new window opens with the alot compose prompt, but once I put in a subject and hit enter, I receive this error `[Errno 2] No such file or directory` The debug log doesn't show anything (see below).

Does alot need an interactive shell for this? I am able to get past this error by setting iTerm to take the `mailto` link, open a new interactive shell, and run `/usr/local/bin/alot compose` as "sent text," but the iTerm documentation is bad and I don't think there is a way to pass the URL or stripped recipient address to this "sent text." 

**Software Versions**
- Python version: 2.7.15
- Alot version: .7

**To Reproduce**
Steps to reproduce the behaviour:
1. click on `mailto` link in web
2. fill in subject line and hit enter
3. See error: `[Errno 2] No such file or directory`

**Error Log**
```
DEBUG:ui:fire first command
DEBUG:ui:global command string: "compose test@test.com"
DEBUG:__init__:mode:global got commandline "compose test@test.com"
DEBUG:__init__:ARGS: [u'compose', u'test@test.com']
DEBUG:__init__:cmd parms {'spawn': None, 'sender': None, 'tags': None, 'cc': None, 'omit_signature': False, 'rest': [u'test@test.com'], 'bcc': None, 'to': None, 'template': None, 'attach': None, 'subject': None}
DEBUG:envelope:TEMPLATE: None
DEBUG:ui:Got key (['enter'], [10])
DEBUG:ui:Got key (['enter'], [10])
DEBUG:globals:SUBJECT: ""
DEBUG:globals:No encryption by default, encrypt_by_default=none
INFO:envelope:editable headers: set(['To', 'From', 'Subject'])
DEBUG:globals:using editor_cmd: nvim '+noh' '+set ft=mail' '+set fileencoding=utf-8' '+set enc=utf-8' '+set ff=unix' '+set fo+=w' '+set backupdir=~/.local/share/alot/nvim-backup'
DEBUG:globals:{'spawn: ': True, 'in_thread': True}
DEBUG:globals:{'spawn': True}
[...]

PR #1628, opened 2023-09-27
ExternalCommand: Fix unbound variables.

ryneeverett:
This issue is referenced here: https://github.com/pazz/alot/issues/1318#issuecomment-629711726


pazz:
Thanks

PR #1633, opened 2023-11-25
Fix configobj to relay tuple type errors

brandonpayton:
Some current config tuple checking is of the form:

```python
if not isinstance(value[1], int) or not isinstance(value[2], int):
    VdtTypeError(value)
```

This has two problems. The first is that `isinstance('123', int)` is always `False`, which means the if's block always runs. A `VdtTypeError` would always be raised except that the `raise` keyword is missing.

This PR attempts to fix that issue.

### DIFF
--- a/tests/utils/test_configobj.py
+++ b/tests/utils/test_configobj.py
@@ -19,22 +19,33 @@ class TestForceList(unittest.TestCase):
         forced = checks.force_list('')
         self.assertEqual(forced, [])
 
+
+class TestWidthTuple(unittest.TestCase):
+
     def test_validates_width_tuple(self):
         with self.assertRaises(VdtTypeError):
             checks.width_tuple('invalid-value')
 
-    def test_validates_width_tuple_for_fit(self):
+    def test_validates_width_tuple_for_fit_requires_two_args(self):
         with self.assertRaises(VdtTypeError):
             checks.width_tuple(['fit', 123])
+
+    def test_args_for_fit_must_be_numbers(self):
         with self.assertRaises(VdtValueError):
             checks.width_tuple(['fit', 123, 'not-a-number'])
+
+    def test_fit_with_two_numbers(self):
         fit_result = checks.width_tuple(['fit', 123, 456])
         self.assertEqual(('fit', 123, 456), fit_result)
 
-    def test_validates_width_tuple_for_weight(self):
+    def test_validates_width_tuple_for_weight_needs_an_argument(self):
         with self.assertRaises(VdtTypeError):
             checks.width_tuple(['weight'])
+
+    def test_arg_for_width_must_be_a_number(self):
         with self.assertRaises(VdtValueError):
             checks.width_tuple(['weight', 'not-a-number'])
+
+    def test_width_with_a_number(self):
         weight_result = checks.width_tuple(['weight', 123])
         self.assertEqual(('weight', 123), weight_result)
--- a/tests/utils/test_configobj.py
+++ b/tests/utils/test_configobj.py
@@ -19,10 +19,11 @@ class TestForceList(unittest.TestCase):
         forced = checks.force_list('')
         self.assertEqual(forced, [])
 
-    def test_validates_width_tuple_fit(self):
+    def test_validates_width_tuple(self):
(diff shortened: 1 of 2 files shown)


lucc:
This `width_tuple` function looks quite simple. Do you mind adding some tests to `tests/utils/test_configobj.py`? Preferably also for the error handling that you improved (i.e. asserting that the right exceptions are thrown).

brandonpayton:
I committed some tests but need to break them into smaller pieces and add one or two more.

brandonpayton:
Thanks also for taking a look at this.

tests/utils/test_configobj.py
```diff
@@ -17,3 +18,23 @@ def test_strings_are_converted_to_single_item_lists(self):
     def test_empty_strings_are_converted_to_empty_lists(self):
         forced = checks.force_list('')
         self.assertEqual(forced, [])
+
```
lucc, line 21:
```suggestion


class TestWidthTuple(unittest.TestCase):

```

tests/utils/test_configobj.py
```diff
@@ -17,3 +18,23 @@ def test_strings_are_converted_to_single_item_lists(self):
...
+    def test_validates_width_tuple(self):
+        with self.assertRaises(VdtTypeError):
+            checks.width_tuple('invalid-value')
+
+    def test_validates_width_tuple_for_fit(self):
```
lucc, line 26:
```suggestion
    def test_validates_width_tuple_for_fit_requires_two_args(self):
```

tests/utils/test_configobj.py
```diff
@@ -17,3 +18,23 @@ def test_strings_are_converted_to_single_item_lists(self):
...
+        with self.assertRaises(VdtTypeError):
+            checks.width_tuple(['fit', 123])
+        with self.assertRaises(VdtValueError):
```
lucc, line 29:
```suggestion

    def test_args_for_fit_must_be_numbers(self):
        with self.assertRaises(VdtValueError):
```

tests/utils/test_configobj.py
```diff
@@ -17,3 +18,23 @@ def test_strings_are_converted_to_single_item_lists(self):
...
+            checks.width_tuple(['fit', 123, 'not-a-number'])
+        fit_result = checks.width_tuple(['fit', 123, 456])
```
lucc, line 31:
```suggestion

    def test_fit_with_two_numbers(self):
        fit_result = checks.width_tuple(['fit', 123, 456])
```

tests/utils/test_configobj.py
```diff
@@ -17,3 +18,23 @@ def test_strings_are_converted_to_single_item_lists(self):
...
+        self.assertEqual(('fit', 123, 456), fit_result)
+
+    def test_validates_width_tuple_for_weight(self):
```
lucc, line 34:
```suggestion
    def test_validates_width_tuple_for_weight_needs_an_argument(self):
```

tests/utils/test_configobj.py
```diff
@@ -17,3 +18,23 @@ def test_strings_are_converted_to_single_item_lists(self):
...
+        with self.assertRaises(VdtTypeError):
+            checks.width_tuple(['weight'])
+        with self.assertRaises(VdtValueError):
```
lucc, line 37:
```suggestion

    def test_arg_for_width_must_be_a_number(self):
        with self.assertRaises(VdtValueError):
```

tests/utils/test_configobj.py
```diff
@@ -17,3 +18,23 @@ def test_strings_are_converted_to_single_item_lists(self):
...
+            checks.width_tuple(['weight', 'not-a-number'])
+        weight_result = checks.width_tuple(['weight', 123])
```
lucc, line 39:
```suggestion

    def test_width_with_a_number(self):
        weight_result = checks.width_tuple(['weight', 123])
```

lucc (changes requested):
The code itself looks good. I suggest breaking the test cases up into individual functions so that each function checks exactly "one thing".

brandonpayton:
That is much cleaner. Thank you!

Issue #1418, opened 2019-07-21
allow each threadline to be in multiple lines


Al-Caveman:
currently a threadline looks like this:

    Wed, 17/Ju  (5)     inbox unread todo  Boaty McBoatface  Re: still unchanged..
    Mon, 10/Ju  (2)     inbox  Someone  Re: some really long long long long long lo
    Thu, 3/Ma   (1)     inbox  Boss  yer fired..
    Wed, 1/Ma   (234)   inbox  Wife  Re: bye, sry

where each threadline, consumes only a single line.

my feature request is to allow for a single threadline to be in multiple lines as follows:

    Wed, 17/Ju  (5)     Boaty McBoatface  
    inbox unread        Re: still unchanged..
    todo

    Mon, 10/Ju  (2)     Someone  
    inbox               Re: some really long long long long long long long long long
                        long long long long long long long long long long long long
                        long long long long long long long long long long subject
    
    Wed, 17/Ju  (5)     Boss  
    inbox               yer fired..

    Thu, 3/Ma  (234)    Wife  
    inbox               Re: bye, sry

here you notice two things:

* fields are distributed over multiple lines.
* too long fields, such as the subject, get wrapped accordingly, as long as they don't hit the max limit (otherwise get cropped).

PR #1634, opened 2023-11-25
Support wrapping column text over multiple lines

brandonpayton:
This PR adds text wrapping for regular text fields to help address issue #1418.

It adds text wrapping support wrapping for thread parts like authors, subject, and content but not for tags.

To make a thread part wrappable, use the following config:
```
# width = 'wrap', min-width, max-width, min-lines, max-lines
width = 'wrap', 40, 40, 2, 2
```

This results in thread lines like:
alot-wrapping-example

A few notes:
1. The new tuple validation in configobj takes the same form as proposed in #1633.
2. For readability, it seems like it would be good to support textwrap.wrap()'s [initial_indent](https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.initial_indent) and [subsequent_indent](https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.subsequent_indent) arguments to better delineate the start of wrapped parts and their subsequent lines, but for the sake of simplicity, that is left for a follow-up PR.
3. These changes do not yet include documentation updates, but I am happy to add those here or in a later PR if this feature addition looks good to you.


alot/widgets/search.py
```diff
@@ -162,6 +180,15 @@ def build_text_part(name, thread, struct):
     return width, part_w
 
 
+def pad_content(content, alignment, min_width):
```
brandonpayton, line 183: I missed documenting this new function. Will fix.

brandonpayton:
I added some missing docs, rebased to incorporate width_tuple validation updates, and added more configobj tests. I think this is ready for review.

PR #1640, opened 2023-12-08
make default theme readable

### LINKED ISSUE #1639: Default theme should always be readable
This issue is related to #33

I've noticed alot doesn't work well on black-on-white terminals.

I know there are light themes available and I know how I can fix this for myself but I think alot should use a default theme which works well in dark and in light mode.

I think this is caused by using default foreground and background colors described as following in the urwid manual:

> 'default' may be specified as a foreground or background to use a terminal’s default color. For terminals with transparent backgrounds 'default' is the only way to show the transparent background. **There is no way to tell what the default colors are, so it is best to use default foregrounds and backgrounds together (not with other colors) to ensure good contrast.**

The default theme from alot uses default foregrounds and backgrounds with other colors, probably due to transparent backgrounds.

My suggestion is to make dedicated dark and light themes and make the default theme a bit more boring (limiting to only using default foreground with a default background and make more use of Bold, Underline, Standout "colors".

I'd gladly do create a PR for this, please let me know how you think about this.

I've created a simple test script to test a theme, I will share this as well if you like. It will show you what's barely/un-readable.

light
dark

meeuw:
Fixes #1639

### DIFF
--- a/alot/defaults/default.theme
+++ b/alot/defaults/default.theme
@@ -6,11 +6,11 @@
 
 [global]
     footer = 'standout','','white,bold','dark blue','white,bold','#006'
-    body = 'default','','dark gray','default','g58','default'
+    body = 'default','','default','default','default','default'
     notify_error = 'standout','','white','dark red','white','dark red'
     notify_normal = 'default','','light gray','dark gray','light gray','#68a'
     prompt = 'default','','light gray','black','light gray','g11'
-    tag = 'default','','light gray','black','light gray','default'
+    tag = 'default','','dark gray','black','dark gray','default'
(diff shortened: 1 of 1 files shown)


pazz (changes requested):
I do not mind changes to the default theme and if this improves visibility I'm happy with it. 
The theme tester script is a welcome addition!

My only two concerns before we can merge this are 
1) let's squash this into one commit or, even better, have one for the theme and one for the script.
2) please run a flake8 or some other python style checker on the script, so that we do not drop on silly code-quality measures
Thanks

meeuw:
Sure, no problem, squashed commits and I will move the Python script to another PR.

pazz:
thanks!

pazz (approved):

PR #1641, opened 2023-12-11
Theme tester

meeuw:
Theme tester

I've formatted the code using black and validated using pylint.

Pylint doesn't like the calls to protected members in theme, but theme doesn't have a usable API for this.

If you want I can add this API to theme?


pazz:
Could you run pyflakes or some other auto-linter over this please, just so that codeclimate does not complain too much for silly things.
Also, It'd be good if this was mentioned in the docs somewhere, like "try the script .. to test your theme file..".
Perhaps you can use argparse in the script to give some ad hoc help, although it is pretty obvious how to use it.
Cheers!

meeuw:
:heavy_check_mark: fixed your remarks, pushed codeclimate to the edge...

pazz (approved):
Thanks!

PR #1644, opened 2024-01-30
Fix default hooks config path to conform to XDG

milhnl:
Hey!

A very small patch that makes the default value for the hooks file use `$XDG_CONFIG_HOME` instead of `~/.config`. The previous behaviour was non-conforming.

I did not update the documentation because of two reasons:

* The documentation is currently inconsistent in calling out env variable support. so adding it should probably be in a different commit.
* the table in `docs/source/api/settings.rst` would get a bit wider and thus unreadable in my editor.

### DIFF
--- a/alot/defaults/alot.rc.spec
+++ b/alot/defaults/alot.rc.spec
@@ -112,7 +112,7 @@ edit_headers_blacklist = force_list(default=list(Content-Type,MIME-Version,Refer
 flush_retry_timeout = integer(default=5)
 
 # where to look up hooks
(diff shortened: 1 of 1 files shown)


pazz:
I don't think this is correct..
This hardcoded location is used as fallback *after* we've already tried to read the suggested location. That is, if the  env var is set then we've already read it [here](https://github.com/pazz/alot/blob/0ba39ccc7f53bd593a3ebf9249db6ddd8fdad394/alot/__main__.py#L102).

pazz:
Oh I see. Thanks for the explanation and the fix!

Issue #1630, opened 2023-10-16
Switch from gpg to python-gnupg


meeuw:
**Is your feature request related to a problem? Please describe.**
https://pypi.org/project/gpg/ hasn't been updated since 2018. I have trouble installing it because it fails with the following error:
```
Using gpgme.h from /usr/include/gpgme.h
Deprecated command line option: -py3. Ignored, this option is no longer supported.
build/gpgme.h:16: Error: CPP #error ""unexpected value for __WORDSIZE macro"". Use the -cpperraswarn option to continue swig processing.
error: command '/usr/bin/swig' failed with exit code 1
```

**Describe the solution you'd like**
Switch to python-gnupg which seems to be better maintained.

**Describe alternatives you've considered**
Make gpg optional in alot. (I'm currently not using gpg to sign/encrypt my mail).

**Additional context**
I noticed that there are tests for the crypto module in alot, I can try to create a PR porting everything to python-gnupg and make all the tests green but first I'd like to check if you agree on this.

dkg:
I think you're looking at the wrong place for the `gpg` module in python.  This is maintained by the GnuPG upstream as part of [gpgme](https://dev.gnupg.org/source/gpgme/browse/master/lang/python/), and has had four releases this calendar year alone.  

`python-gnupg`, by contrast,  is developed by someone outside the GnuPG project and [has had only one release this calendar year](https://github.com/vsajip/python-gnupg).

meeuw:
You're right, that works for me, thanks!

It could still make sense to switch to python-gnupg to ease the installation of alot though.

I leave this decision to the maintainer(s) of this project, my offer still stands to port alot to python-gnupg.

pazz:
I am still struggling with this. I agree that it'd be good to depend on the official gnupg bindings but the current (alot) dependencies point to this outdated PIP repository that is not installable on debian as mentioned above.
Two ways forward: Either port to this python-gpgme package as suggested in the OP, or to pester the main gpgme people to take over the PIP entry.. Thoughts?

lucc:
My gut feeling is to prefer the official gpg package. I would thus vote to "pester the gpg folks".

We have a similar problem with the python notmuch bindings which are also non trivial to install and are not [released to PyPI](https://pypi.org/project/notmuch/) for any recent version.  This also came up in #1617.

For me this topic is connected to the question of our build system: are we using setuptools or are we using poetry? We have a pyproject.toml file which specifies
~~~toml
[build-system]
[...]

PR #1645, opened 2024-02-13
regenerate lock file

meeuw:
As requested in #1630 

This is using Fedora Linux with `python3-gpg-1.20.0-5.fc39.x86_64` installed, then I executed `poetry config virtualenvs.options.system-site-packages true` and regenerated my virtualenv (`poetry env remove  alot-XXXXXXX-py3.12`), then ran `poetry lock` and created this PR.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -12,32 +12,23 @@ files = [
 ]
 
 [[package]]
-name = "atomicwrites"
-version = "1.4.1"
-description = "Atomic file writes."
-optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
-files = [
-    {file = "atomicwrites-1.4.1.tar.gz", hash = "sha256:81b2c9071a49367a7f770170e5eec8cb66567cfbbc8c73d20ce5ca4a8d71cf11"},
-]
-
-[[package]]
 name = "attrs"
-version = "22.2.0"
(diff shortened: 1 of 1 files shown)


poetry.lock
```diff
@@ -199,36 +266,53 @@ six = "*"
...
+version = "1.20.0"
+description = "Python bindings to the GPGME API of the GnuPG cryptography library."
+optional = false
+python-versions = "*"
+files = []
+develop = false
+
+[package.source]
+type = "directory"
+url = "../../../../usr/lib64/python3.12/site-packages"
```
lucc, line 314: This looks strange to me. Does it reference the python-gpgme installation from yum?

meeuw: It references python-gpgme from system site-packages yes. I do agree that `../../../..` it very specific to the directory where alot is installed (`/home/user/git/alot` in my case)

Do you have a better suggestion?

lucc: Even worse, it is specific to the python version so this lock file will not work with python 3.11 or 3.13 even if I have gpgme installed with my distribution.

Sadly I have no better idea.  We have to investigate how to install the upstream gpgme with poetry otherwise we might have to ditch one of the two.

meeuw: Or ditch poetry.lock? With the current pypoetry.toml it's impossible to have a sane poetry.lock.

PR #1646, opened 2024-02-17
Properly sanitize control chars in subject/headers/body

lfos:
I recently noticed that some messages (primarily spam) contain [control characters](https://www.compart.com/en/unicode/category/Cc) that result in garbled output in both the search widget and the thread view. This PR addresses this by

1. removing all unwanted control characters (and not just CR) from strings before outputting and
2. applying string_sanitize() to subjects in the search widget.

### DIFF
--- a/alot/db/thread.py
+++ b/alot/db/thread.py
@@ -3,8 +3,10 @@
 # For further details see the COPYING file
 from datetime import datetime
 
+from ..helper import string_sanitize
 from .message import Message
 from ..settings.const import settings
+from .utils import decode_header
 
 
 class Thread:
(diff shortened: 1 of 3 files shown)


lucc:
Can we have a test email for this?

lfos:
@lucc To manually reproduce, just take a random email and append `U+009d` to the subject line; other control characters work too. Note that the exact result may also depend on the terminal you're using; I'm using kitty to reproduce the garbled output. I also added a test case to the commit.

pazz (approved):
The change looks good and safe, I am happy with adding the dependency on unicodedata, which seems to be part of the standard library anyway.
Also thanks for the test case!

PR #1648, opened 2024-02-19
editnew: Use raw body text instead of rendered version

guludo:
Rendering mail content via mailcap makes sense when viewing email, but not when reusing the email for composing a new message, as it is desirable to rather edit the "original" content and, if mailcap rendering produces ANSI CSI escapes, the user would be left with a bunch of garbarge characters.

Make sure that the editnew command do not render the body when creating the envelope.

### DIFF
--- a/alot/commands/thread.py
+++ b/alot/commands/thread.py
@@ -466,7 +466,7 @@ class EditNewCommand(Command):
                                 'signed', 'encrypted', 'unread', 'attachment'})
         tags = list(tags)
         # set body text
-        mailcontent = self.message.get_body_text()
+        mailcontent = self.message.get_body_text(render=False)
         envelope = Envelope(bodytext=mailcontent, tags=tags)
 
         # copy selected headers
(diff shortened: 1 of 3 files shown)


pazz:
Without having reviewed the code:
IS this what we actually want here?

I ask because IIRW, many mail clients sent out html mails with the plaintext alternative empty or not complete.
So by default I would (personally) be able to respond to the rendered variant.
With the current master version, you can respond to the plaintext alternative by first displaying the plaintext version in thread mode
(togglemimepart) and then replying...

thoughts?

Quoting Gustavo José de Sousa (2024-02-19 18:06:30)

pazz:
I think adding an option to èditnew`and perhaps a config switch for the default value of that option would make sense here, yes.

pazz:
I'm going over old/stale commits towards making a release soonish.
Would you like to repackage this with the extra option and include in a release?
I think I now tend to agree with you that this is a good idea if editnew is used for drafts.

PR #1653, opened 2024-05-02
Add github action to check the nix build

lucc:
This adds CI jobs to run the nix build of alot and also a second nix target with a ci job to build the sphinx docs.

Some test are fixed or removed, see commit messages.  The last question about the fixed test is https://github.com/pazz/alot/pull/1653#issuecomment-2094563736

The plan behind this is that we have *some* CI that is green until the pip/poetry installation of alot is fixed again and our other CI jobs can be trusted.  And also to check that the nix code works correctly because I personally rely on it.

### DIFF
--- a/alot/db/manager.py
+++ b/alot/db/manager.py
@@ -51,9 +51,7 @@ class DBManager:
         if exclude_tags is not None:
             return exclude_tags
 
-        exclude_tags = settings.get_notmuch_setting('search', 'exclude_tags')
-        if exclude_tags:
-            return [t for t in exclude_tags.split(';') if t]
+        return settings.get_notmuch_setting('search', 'exclude_tags')
 
     def flush(self):
         """
@@ -72,7 +70,7 @@ class DBManager:
             raise DatabaseROError()
         if self.writequeue:
             # read notmuch's config regarding imap flag synchronization
-            sync = settings.get_notmuch_setting('maildir', 'synchronize_flags') == 'true'
+            sync = settings.get_notmuch_setting('maildir', 'synchronize_flags')
 
             # go through writequeue entries
             while self.writequeue:
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,4 +1,4 @@
-name: Run code checks
+name: Run code checks with different build tools
 
 on:
   - push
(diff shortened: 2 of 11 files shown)


pazz:
Is this work in progress or ready to merge? I trust you know what you are doing wrt github actions as I don't :)

lucc:
Sorry this is still WIP as some tests are still failing in the nix sandbox.

pazz:
cool, no worries :)

lucc:
@pazz & @guludo what do you think about 1359aba? I just removed the broken test, it would also be possible to add some validation code to the read_notmuch_config function to throw that exception again. We are only using the following values from the notmuch config:

- `maildir.synchronize_flags`, bool as the string "true" or "false" in the notmuch config
- `database.path`, a path as a string in  the notmuch config
- `search.exclude_tags`, a list of strings as a ";"-separated string in the notmuch config

so we could also parse these into special values inside `read_notmuch_config`.

guludo:
> @pazz & @guludo what do you think about [1359aba](https://github.com/pazz/alot/commit/1359abaaad5dffa6b5d07edc070c39d24d475161)?

Looks good to me.

> I just removed the broken test, it would also be possible to add some validation code to the read_notmuch_config function to throw that exception again.

That would make sense. Ideally, `notmuch config` would raise an error, since the documentation says that valid values are either true or false, but that does not seem to be the case.

lucc:
From my side this is ready for review (and merge).

@pazz is it ok to ignore the codeclimate check?

alot/db/manager.py
```diff
@@ -72,7 +70,7 @@ def flush(self):
             raise DatabaseROError()
         if self.writequeue:
             # read notmuch's config regarding imap flag synchronization
-            sync = settings.get_notmuch_setting('maildir', 'synchronize_flags') == 'true'
+            sync = settings.get_notmuch_setting('maildir', 'synchronize_flags')
```
pazz, line 73: lol

pazz (approved):
I do not have experience with nix, flake nor github actions but the changes look good and to me, especially the work around the gpg issue. If this fixes the current problems with the CI pipeline then let's go for it! Many thanks!

Issue #1608, opened 2022-09-14
Crash while trying to attach word .doc file: module 'magic' has no attribute '_libraries


zevv:
Trying to attach a .doc document throws the following error:

```
INFO:envelope:editable headers: ['From', 'To', 'Subject']
INFO:globals:calling external command: ['vim', '+5', '/tmp/alot.ydvdzjr9.eml']
INFO:globals:open command shell
INFO:envelope:attaching: ['/tmp/attachment.doc']
ERROR:ui:Traceback (most recent call last):
  File "/usr/share/alot/alot/ui.py", line 723, in apply_command
    cmd.apply(self)
  File "/usr/share/alot/alot/commands/envelope.py", line 66, in apply
    envelope.attach(path)
  File "/usr/share/alot/alot/db/envelope.py", line 173, in attach
    part = helper.mimewrap(path, filename, ctype)
  File "/usr/share/alot/alot/helper.py", line 458, in mimewrap
    not libmagic_version_at_least(513)):
  File "/usr/share/alot/alot/helper.py", line 414, in libmagic_version_at_least
    magic_wrapper = magic._libraries['magic']
AttributeError: module 'magic' has no attribute '_libraries'
```

python3-magic comes from debian upstream and does not seem to have the `_libraries`; I made a local workaround by changing the offending line into

```
    if hasattr(magic, 'open') and hasattr(magic, '_libraries'):
```

boutil:
For reference, this bug has been also reported in Debian by @zevv .

boutil:
I am reading the comment around line 336 of alot/helper.py. The source package python-magic in Debian  based on the code of  `https://github.com/ahupp/python-magic` is not using directly the bindings from the `file` source directly. Rather, the compatibility layer has been now part of upstream python-magic module.
The change in the Debian package occured in January 2021, with version 2:0.4.20-1, according to the changelog of Debian's python3-magic package.

Would it make sense to ask a version >= 0.4.20 for the magic python module, and drop the logic in gess_mimetype?

PR #1663, opened 2024-07-17
removed obsolete/faulty libmagic workaround

paumr:
This fixes: #1608

This workaround was introduced in 2013: d093628919b35ee5dda77ca1e4f4e4f951259fd3
Since then the libmagic bug was fixed, additionally it should have been propagated through all release channels of all major distros.

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -7,7 +7,6 @@ from datetime import timedelta
 from datetime import datetime
 from collections import deque
 import logging
-import mimetypes
 import os
 import re
 import shlex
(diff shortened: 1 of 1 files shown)


pazz (approved):
I agree that we can and should remove this now.

paumr:
Thanks for the quick reaction! 

Would you mind triggering a new release?

pazz:
I will try to remember but that takes me a while:
I will want to go over outstanding PRs to see if there's anything pressing,
then go over my release flowchart :)
Thanks,
P


Quoting paumr ***@***.***) at 2024-07-17 14:09:28

PR #1670, opened 2024-08-01
nix: install support files

lucc:
This makes our  nix derivation usable not only as a development environment but also for installation in a nix based system and daily use. 

(I am installig the current master of alot on my home computer through this way and would like to have these goodies)


pazz (approved):
sure, why not..

PR #1671, opened 2024-08-06
nix: parse dependencies from pyproject file

lucc:
@pazz this is the result of my comment https://github.com/pazz/alot/issues/1666#issuecomment-2270396622 . This makes the nix code a little more complicated but removes the need to specify our dependencies twice.  As I know nix quite well I like it but I am unsure if it raises the bar for others to look at the nix code (do we care about that?). What do you think?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,12 +8,19 @@
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = nixpkgs.legacyPackages.${system};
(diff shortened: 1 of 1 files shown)


pazz (approved):
I think it is a great idea to avoid duplication of the dependencies.
Otherwise inevitably these will diverge and possibly create issues.

Perhaps you could add one larger comment before the affected lines explaining that the following replaces explicit dependencies and instead extracts them from the setuptools config?

Issue #1597, opened 2022-02-15
Not compatible with notmuch2 0.35 bindings


inglor:
Before you submit a bug report, please make sure that the issue still exists on the master branch!

**Describe the bug**
notmuch 0.35 released a few days ago and the test suite fails due to breaking changes on the `_database.py`. I think the commit which introduce these changes is https://github.com/notmuch/notmuch/commit/8b737af28bc377db3e661a5744f3b7479b7ce485

**Software Versions**
- Python version: 3.10
- Notmuch version: 0.35
- Alot version: 0.10, master

**To Reproduce**
Steps to reproduce the behaviour:
1. execute test suite with notmuch2 bindings 0.35

**Error Log**
Test suite error:
```
ERROR: test_save_named_query (tests.db.test_manager.TestDBManager)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/build/alot/src/alot-0.10/alot/db/manager.py", line 92, in flush
    db = Database(path=self.path, mode=mode)
  File "/usr/lib/python3.10/site-packages/notmuch2/_database.py", line 160, in __init__
    raise errors.NotmuchError(ret, msg)
notmuch2.NoConfigError: Error: cannot load config file.


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/build/alot/src/alot-0.10/tests/db/test_manager.py", line 49, in test_save_named_query
    self.manager.flush()
  File "/build/alot/src/alot-0.10/alot/db/manager.py", line 175, in flush
    raise e
  File "/build/alot/src/alot-0.10/alot/db/manager.py", line 94, in flush
    raise DatabaseLockedError()
alot.db.errors.DatabaseLockedError
```

oliversturm:
On Arch Linux I have seen this problem for a long time, and worked around it by downgrading to notmuch 0.34. I didn't run test suites though - I simply can't use alot because it comes up with a blank screen at all times. Now I'm running on a Mac where alot and notmuch are installed via brew (see versions below), and it's not trivial to downgrade.

For clarity and completeness, here are steps I take:

1. Run `alot -n <mynotmuchconfig> -l ./alot.log`
2. The screen goes completely blank. I can use `o` to search, `d` to leave the (invisible) result screen, `q` to quit alot, but I must do it all blind.
3. I find this output at the top of the log file:

```
INFO:ui:setup gui in 256 colours
ERROR:ui:Traceback (most recent call last):
[...]

PR #1673, opened 2024-08-07
notmuch2 copatibility

pazz:
This change causes alot to pass on the "notmuch-config" option (path to notmuch's configuration file) to any call to `Database(..)` in the underlying database.

#1597


lucc (approved):
LGTM and I can run it successful.

PR #1675, opened 2024-08-10
Reacticate CI jobs after switch to pyproject and setuptools

lucc:
I want to reactivate the unittests on the ubuntu runners and find a replacement for the `petry check` jobs.  We can even think  about pycodestyle or mypy or the like.

@pazz I added this to the 0.11 milestone but I don't know how quickly you want to release that and how quickly I can finish this PR. You can just drop it from 0.11 if you want as we have at least some working CI with the nix flake.


lucc:
@pazz I did not install gpg in ci because it was even more complicated than notmuch. Instead I patch the pyproject dependencies during ci, removing gpg and then add a mock of the package and skip all other tests that depend on the gpg module.

This is suboptimal but for me it is still acceptable because we get the python test matrix back and the gpg related tests are still run in the nix ci job.

What do you think?

pazz:
> What do you think?

Sounds like a sensible temporary solution, yes!

pazz (approved):

PR #1678, opened 2024-08-13
Fix build on RTD with dynamic version

lucc:
I think this should do it.

@pazz  I took the liberty to also push this branch to the "rtd" branch as that still looks activated on RTD. So I hope to get a build of this branch on RTD (maybe you can check and if neccesary activate the dversion or dversion2 branch on RTD).


pazz:
thanks

PR #1680, opened 2024-09-05
Keep ESC control characters in the message body

### LINKED ISSUE #1679: ANSI escape sequences are not interpreted in the message body
Earlier versions of alot interpreted ANSI escape sequences in the messages bodies for e.g. HTML coloring. This feature was broken by 11ea16f1651a3e8321dfafb5ef0d8b060becc14a, because it also strips ESC `\x1b` control characters. Leaving just the coloring commands like e.g. `[1;30m`. Returning true for `\x1b` along side `\n` in the `unicode_printable` function brings back the old behavior.

Whether that feature was a good idea to begin with is debatable, so if you want it like this I don't mind, but then at least [this](https://github.com/pazz/alot/blob/master/docs/source/configuration/theming.rst#ansi-escape-codes) part of the documentation should be removed. And I am guessing a lot of `alot/widgets/ansi.py` might be superfluous now

**Software Versions**
- Python version: 3.12.5
- Alot version: 0.11

johslarsen:
In order to bring back support for ANSI colored HTML renderings etc.

Fixes #1679

### DIFF
--- a/alot/helper.py
+++ b/alot/helper.py
@@ -44,10 +44,11 @@ def split_commandstring(cmdstring):
 def unicode_printable(c):
     """
     Checks if the given character is a printable Unicode character, i.e., not a
-    private/unassigned character and not a control character other than tab or
-    newline.
+    private/unassigned character and not a control character other than tab,
+    newline or ESC (for ANSI escape sequences, e.g. HTML coloring).
     """
-    if c in ('\n', '\t'):
+    ESC = '\x1b'
+    if c in ('\n', '\t', ESC):
         return True
     return unicodedata.category(c) not in ('Cc', 'Cn', 'Co')
 


alot/helper.py
```diff
@@ -47,7 +47,8 @@ def unicode_printable(c):
     private/unassigned character and not a control character other than tab or
     newline.
```
lfos, line 48: Thanks! Can you please update the comment too, adding ESC control characters?

lfos:
I addition to the minor code comment, I am not sure this is a great idea; I added some thoughts to the original discussion in #1679.

lfos:
As an alternative to what I suggested in #1679, we could try to allowlist certain escape characters; e.g., those that change color and basic formatting. That said, we'd need to be quite careful with what we allow to ensure there's no safety issue.

lfos:
Proposed secure alternative to this PR: #1697

PR #1685, opened 2024-12-08
Update flake for new mailcap dependency

lucc:
This should fix the nix build which is broken due to #1684.

The standard-mailcap package is not yet available in nixpkgs so this still breaks for python 3.13 on nix but at least the dependency parsing works again and the derivation can be used with python <=3.12.

---------

@pazz what do you think about requiring green CI for pull requests? I would quite like that.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,10 +15,14 @@
         # get a list of python packages by name, used to get the nix packages
         # for the dependency names from the pyproject file
         getPkgs = names: builtins.attrValues (pkgs.lib.attrsets.getAttrs names pkgs.python3Packages);
-        # extract the python dependencies from the pyprojec file, cut the version constraint
-        dependencies' = pkgs.lib.lists.concatMap (builtins.match "([^>=<]*).*") pyproject.project.dependencies;
+        # extract the python dependencies from the pyprojec file, cut the version constraints
+        dependencies' = pkgs.lib.lists.concatMap (builtins.match "([^>=<;]*).*") pyproject.project.dependencies;
         # the package is called gpg on PyPI but gpgme in nixpkgs
-        dependencies = map (x: if x == "gpg" then "gpgme" else x) dependencies';
+        renameGPG = x: if x == "gpg" then "gpgme" else x;
+        # mailcap has been removed from the stdlib in py3.13 and needs to be
+        # fetched from pypi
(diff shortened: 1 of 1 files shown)


pazz:
You are right, I did not consider the CI status when pushing this las PR.
In principle I think it is a good idea to require a green status, yes.
About flake/nix in particular I worry though that you are among the only ones who care and could reasonably fix this, and perhaps requiring such fixes is creating unecessary hurdles?

lucc:
@pazz we could leave the nix stuff out of the required checks but I would prefer if we can see in a pr that it gets broken. If nobody can/wants to figure out how to fix it they can at least ping me.

lucc:
The problem about the current situation is, that we can not run all our test cases on a default ci machine (ubuntu) because the python gpg module is not installable (#1630).  We only run 289 test cases on the ubuntu runner vs 344 in the `nix build`

https://github.com/pazz/alot/actions/runs/12234986811/job/34125317235#step:10:346

https://github.com/pazz/alot/actions/runs/12234986810/job/34125316256#step:4:1283

pazz:
You are right. let's do it!

lucc:
@pazz I think the first step is the settings in the alot project page here on github (and I think only you can edit that).

You would need to create a protection rule for the master branch and then define who can push to it or what is required from a merge request.

pazz:
I think I did this now

2024-12-11T08:56:01,883875071+00:00

Tangentially related: I would really like to move this repo under the notmuch org and give you full access in the long run.
Not sure if this would preserve all issues/PRs.. thoughts?

pazz:
Do you think it'd be a good idea to do this?


Quoting Lucas Hoffmann ***@***.***) at 2024-12-11 09:30:07

lucc:
I don't know if it is a good idea.  That depends on what  your goals with moving to the notmuch org would be?

- more visibility for alot?  --> might not change with an alot-org
- shared responsibility?  --> can be done with an alot-org (might also be doable in your repo right now)
  - we could add a [`CODEOWNERS` file](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) (you could also put me down as the code owner of the nix code)
- clear distinction between alot team and other contributors? 
  - easier in alot-org than notmuch-org I think, in notmuch-org the alot team and the notmuch team might get mixed up by outsiders --> outsiders might not know who to talk to or who is responsible for what (I assume that the alot team and the notmuch team are currently distinct sets of people)
  - CODEOWNERS again

PR #1686, opened 2024-12-09
Run test on python 3.13

lucc:
As we have a temporary fix for #1632 now we should run our tests on python 3.13.

I am still working on packaging the standard-mailcap package on nixpkgs. If that is finished I would like to use python 3.13 for the nix code as well. This PR will be a draft until then.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


lucc:
Waiting for https://github.com/NixOS/nixpkgs/pull/365443.

pazz (approved):
Looks good to me. Thanks for modernising my old scripts! :)

PR #1687, opened 2024-12-22
fix default-theme to correctly highlight focused attachments

paumr:
e90da9b3da225683c0ab6e8890ebda916d3c4d97 broke the highlighting of focused_attachments.

This PR resets the faulty value in question to its prior color.

### DIFF
--- a/alot/defaults/default.theme
+++ b/alot/defaults/default.theme
@@ -32,7 +32,7 @@
     arrow_heads = '','','dark red','','#a00',''
     arrow_bars = '','','dark red','','#800',''
     attachment = 'default','','light gray','dark gray','light gray','dark gray'
-    attachment_focus = 'underline','','light gray','light green','light gray','light green'
+    attachment_focus = 'underline','','dark gray','light green','dark gray','light green'
     body = 'default','','default','default','default','default'
     body_focus = 'default','','default,standout','default','default,standout','default'
     header = 'default','','white','dark gray','white','dark gray'
--- a/alot/defaults/default.theme
+++ b/alot/defaults/default.theme
@@ -32,7 +32,7 @@
     arrow_heads = '','','dark red','','#a00',''
     arrow_bars = '','','dark red','','#800',''
     attachment = 'default','','light gray','dark gray','light gray','dark gray'
-    attachment_focus = 'underline','','light gray','light green','light gray','dark gray'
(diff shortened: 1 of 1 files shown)


lucc:
In what sense is it broken? The commit you mention is from #1640, @meeuw what was the motivation for the change from green to gray at this point specifically?

I checked on xterm's default theme and it might be a little hard to read there:
2024-12-22-224526_243x42_scrot


@paumr can you please also check your change with the `extra/theme_test.py` script, preferably on different terminals (light and dark).

paumr:
@lucc thanks for looking into this. Your terminal probably only supports 16 colors, in which case everything works as expected.

As you can see here: https://alot.readthedocs.io/en/latest/configuration/theming.html#colour-attributes
The configuration format is:
```
name = mono fg, mono bg, 16c fg, 16c bg, 256 fg, 256 bg
```

means #1640 kept the 16c highlighted-background at light green, but set the 256c background to "dark gray" (the same color as the non-highlighted setting). I assume this was a copy and paste error.

If you want to reproduce this issue open alot in a 256color terminal. Then you won't see any attachment highlighting.

lucc:
Ok now I understand what you say about #1640  braking the theme (having a typo) at this point.

My point about it being hard to read after your proposed change is the contrast between the fg and bg of the selected attachement:

2024-12-23-175001_248x40_scrot

I consider that light gray to be hard to read on the light green.

Do you mind changing the fg or bg a bit so that it is easier to read the text?

lucc:
PS for other to understand the bug in the current master: the current theme does not distinguish between selected and not selected attachement:

2024-12-23-180003_602x18_scrot
2024-12-23-180026_469x26_scrot

That is bad UI.

paumr:
Ok, as requested I changed the `fg` color to `dark grey` to increase the contrast a bit.
I hope this change doesn't break too many themes/styles.

meeuw:
I don't remember changing green to gray, my main rationale was to change the colors to make them both usable for black on white as well as white on black terminals.

I actually think it's a typo (from me)

lucc (approved):

lucc:
@meeuw  can I ask you  to review this? Otherwise I would merge it.

meeuw (approved):
Sure! I did some quick testing and it's readable on both black-on-white as white-on-black terminals.

Thanks for fixing!

PR #1692, opened 2025-05-29
Python 3.13

meeuw:
Add tests for Python 3.13

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.12"]
+        python-version: ["3.8", "3.13"]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


lucc (changes requested):
The whitespace changes in the generated docs files break the formatting in the rendered html.  Somehow python 313 seems to do different things so we might need to change the generation scripts.

lucc:
@meeuw I looked into the whitespace problem.  It is due to a change in docstring handling: https://docs.python.org/3.13/whatsnew/3.13.html#other-language-changes

While working on a fix I noticed that I already had a PR open for adding python 3.13 to CI (#1686). I am sorry that I previously forgot and forced you to do the work again. 

I picked your python 3.13 commit but dropped your docs regeneration and added my whitespace fixes to #1686.  

I will now close this in favor of #1686.

meeuw:
No worries, this was only a small change!

PR #1698, opened 2025-08-24
db/utils: correctly handle non-UTF8 8bit encodings

johslarsen:
The 8bit encoded mail parts should be encoded consistently with the charset they specify in their Content-Type header. Otherwise subsequent decoding of non-ASCII characters is likely to produce the wrong string.

Currently our usage of [Message.get_payload] returns a string decoded from the charset in the message header. From discussion in pazz#1522, pazz#1526 and 777823f it seems that it its behavior (or our usage of it) may have changed over the years. In the 8bit case this code essentially tries to undo the get_payload decoding. Arguably it would be better if the email library provided a proper bytestring or its internal ASCII with surrogateescape representation, but this is what we have available.

[Message.get_payload]: https://github.com/python/cpython/blob/96b7a2eba423b42320f15fd4974740e3e930bb8b/Lib/email/message.py#L297

### DIFF
--- a/tests/db/test_utils.py
+++ b/tests/db/test_utils.py
@@ -35,6 +35,12 @@ def set_basic_headers(mail):
     mail['From'] = 'bar@example.com'
 
 
+def message_from_binary_file(filename):
+    with (Path("tests/static/mail") / filename).open("rb") as file:
+        return email.message_from_binary_file(
+            file, _class=email.message.EmailMessage)
+
+
 class TestGetParams(unittest.TestCase):
 
     mailstring = '\n'.join([
@@ -725,12 +731,16 @@ class TestExtractBodyPart(unittest.TestCase):
         self.assertEqual(actual, expected)
 
     def test_simple_japanese_file(self):
-        with open('tests/static/mail/japanese.eml', 'rb') as file:
-            mail = email.message_from_binary_file(file)
+        mail = message_from_binary_file('japanese.eml')
         actual = utils.extract_body_part(mail)
         expected = "MA-EYESご利用者各位\n\nBIRD-BOの河和です。お疲れ様です。\n"
         self.assertEqual(actual, expected)
 
+    def test_simple_iso_8859_1_mail(self):
+        mail = message_from_binary_file("iso-8859-1-qp.eml")
+        actual = utils.extract_body_part(mail)
+        expected = "müßig\nfleißig\n"
+        self.assertEqual(actual, expected)
 
     def test_text_plain_with_attachment_text(self):
         mail = EmailMessage()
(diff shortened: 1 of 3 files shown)


johslarsen:
> Did you investigate if it is easier (and maybe less manual juggling of encodings) to use the new api?

Interesting. None of the deprecation was mentioned in the Python 3.13 source code only on the documentation page. Had not looked into it, but skimmed it now.

Even more interesting is that the current `EmailMessage` is just a subclass of the deprecated Message:
```py
class Message:
class MIMEPart(Message):
class EmailMessage(MIMEPart):
```
So the deprecation is really having those methods public I guess, and some complicated policy logic that defaults to this Compat32 one if `Message`s are created directly.

> This is a legacy method. On the EmailMessage class its functionality is replaced by get_content() and iter_parts()

The `get_content` method seems to be a fancy mime-based factory lookup. Maybe it is meant as an extension point, but the default (and only content manager python ships with the library) [raw_data_manager](https://docs.python.org/3/library/email.contentmanager.html#email.contentmanager.raw_data_manager) just results in calls to `get_payload(decode=True)`. The decode flag makes it decode quoted-printable and base64, etc. so more of our `remove_cte` implementation. Using it for decoding might be a good idea, though. For `8bit` transfer encoding it seems to just rencode the text into the raw bytes instead of decoding it to the specified charsets, which is more direct than our current implementation.

The `iter_parts` calls `get_payload` in this mode where it returns a list of `Message`s (probably `MIMEPart` subclasses), which is fairly similiar to how we iterate over multipart messages already, although in the `MIMEPart`s we are still calling `get_payload` instead of `get_content`.


I tried just changing `get_payload` here to `get_content`, then tests starts failing:
* 4 times `AttributeError: 'Compat32' object has no attribute 'content_manager'`
* 3 times `AttributeError: 'Message' object has no attribute 'get_content'. Did you mean: 'get_content_type'?`
Guessing we would need to to port everything from Message to EmailMessage for this to work.

I tried replacing most of our `remove_cte` implementation with `get_payload(decode=True)`, then all of the tests pass (including the new one here), and everything I manually tested in my mail archive seems to work, but I did not test attachments etc. exhaustively, and I am not sure the test suite do it either. I will push a second commit with this change. If we merge that I still think the initial commit here would be a good idea to include in case it turns out that having email do the decoding comes with unintended consequences.

lucc:
@johslarsen thanks for your contribution, this was a long standing issue.

I have decided to skip some of the issues I mentioned above in https://github.com/pazz/alot/pull/1698#issuecomment-3419953472 because they might already be fixed and I did not want to delay this PR any longer.

PR #1710, opened 2026-01-24
Include package defaults pyproject.toml

Hellseher:
Include `alot/defaults/` directory containing no `*.py` files ignored by setuptools.

Reported in https://codeberg.org/guix/guix/issues/5797

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,9 +51,9 @@ Issues = "https://github.com/pazz/alot/issues"
 
 [tool.setuptools.packages.find]
 include = ["alot*"]
-include-package-data = true
 
-[tool.setuptools_scm]
+[setuptools.package-data]
+alot = ["defaults/*"]
 
 [tool.mypy]
 python_version = "3.9"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,7 @@ Issues = "https://github.com/pazz/alot/issues"
 
 [tool.setuptools.packages.find]
 include = ["alot*"]
+include-package-data = true
 
 [tool.setuptools_scm]
 


lucc:
Where are you getting your wheels from? If I run `python3 -m build` on v0.12 the wheel includes the defaults dir:
~~~sh
git checkout master # or v0.12
python3 -m build
python3 -m zipfile --list dist/alot-0.12-py3-none-any.whl | grep defaults
~~~

I see no benefit in these changes or rather I can not reproduce the issue from guix that you linked.  I tried to read the definition of the alot derivation in mail.scm but it seems to fetch the source from github via git (but I only use and know nix and not guix so I might misinterpret what I read)

Hellseher:
Let me try to reproduce it on my end, may you specify Python exact version, setuptools version and other  system wide dependencies version exact, please.

Hope you are aware about this one https://setuptools.pypa.io/en/latest/userguide/datafiles.html
as project specifies in `pyproject.toml`
``` toml
[build-system]
requires = ["setuptools", "setuptools-scm"]
build-backend = "setuptools.build_meta"
```

lucc:
The python version I use is from the nix flake in the alot project.  Currently that is Python 3.12.10 and more precisely /nix/store/x5vzp7g27a6lkma7arvnrdg8x0xd8m8n-python3-3.12.10-env/bin/python .

I interpret the docs you link to also  say that the files should currently already be included.
See https://setuptools.pypa.io/en/latest/userguide/datafiles.html#configuration-options  under 1.2:
> They are being tracked by a revision control system such as Git, Mercurial or SVN, AND you have configured an appropriate plugin such as [setuptools-scm](https://pypi.org/project/setuptools-scm)

And the directory you are missing is tracked by git:
~~~console
$ git ls-files alot/defaults
alot/defaults/abook_contacts.spec
alot/defaults/alot.rc.spec
alot/defaults/default.bindings
alot/defaults/default.theme
alot/defaults/theme.spec
~~~

PR #1713, opened 2026-04-04
Replace glob1 with glob

### LINKED ISSUE #1712: Python 3.15 removed glob1 (used in alot/account.py)
**Describe the bug**
Python 3.15 removed glob1 (used in alot/account.py)

https://docs.python.org/3.15/whatsnew/3.15.html

> Removed the undocumented glob.glob0() and glob.glob1() functions, which have been deprecated since Python 3.13. Use glob.glob() and pass a directory to its root_dir argument instead. (Contributed by Barney Gale in gh-137466.)

meeuw:
Fixes #1712 

As glob.glob1 is undocumented and it's not so clear to me why it was used in the first place I'm not 100% sure if this fix is right.

### DIFF
--- a/alot/account.py
+++ b/alot/account.py
@@ -299,10 +299,9 @@ class Account:
         if isinstance(mbx, mailbox.Maildir):
             # this is a dirty hack to get the path to the newly added file
             # I wish the mailbox module were more helpful...
-            plist = glob.glob1(os.path.join(mbx._path, 'new'),
-                               message_id + '*')
+            plist = glob.glob(os.path.join(mbx._path, 'new', message_id + '*'))
             if plist:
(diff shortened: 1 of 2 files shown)


lucc (approved):

lucc:
@pazz I did a quick check and this seems to work.

Was there an original reason to use glob1 instead of glob or pathlib.Path.glob?

pazz:
Quoting Lucas Hoffmann ***@***.***) at 2026-04-07 06:15:18

> @pazz I did a quick check and this seems to work.


\o/

 
> Was there an original reason to use glob1 instead of glob or pathlib.Path.glob?

I don't remember, sorry.

PR #1717, opened 2026-05-10
sys.stderr replace mock with io.StringIO

meeuw:
Python 3.15 does some formatting on sys.stderr for argparse and this doesn't work with a Mock object (it tries call fileno and expects an integer).

Use io.StringIO instead, I think it's only used for supressing stderr.


pazz:
It looks safe to me but I'll fire up a copilot review just in case as I'm no expert. thanks for pushing this project :)

lucc (approved):
lgtm

PR #1711, opened 2026-02-24
Implement `theme` command.

dylanfprice:
Fixes #1588. Took inspiration from #1603.

### Testing

I switched back and forth between a few different themes I have and visually it seemed to work great. I also was able to switch back to the default theme with `:theme default.theme`.

I switched to the same theme I was already using and verified that it didn't rebuild the buffer (e.g. the thread I was viewing didn't reset its collapsed state).

I tried loading `:theme doesntexist` and saw the error message in the UI.

I sabotaged one of my theme files and tried to load it. I saw an error message in the UI and this in the logs:
```
ERROR:theme:Theme file `/home/dylan/.config/alot/themes/solarized-light` failed validation: key "footer" in section "global" is missing.
```

I added a few unittests for the `get_theme` function I wrote.

### Docs

~~I took a stab at updating the docs, but didn't try to build it or anything.~~ I didn't realize the command docs were generated, fixed.

PR #1720, opened 2026-06-05
Set ui global body color on update.

dylanfprice:
I ran into a bug with my recently merged PR #1711. When switching the theme using the `theme` command, the global body color is not updated.

I didn't notice this until I switched to a dark theme and also changed my terminal background color to match. The background color of alot remained what it was in the light theme.

The issue is that the global body attribute is only set in the `UI.__init__` constructor. I fixed it by setting the global body attribute again in the `UI.update` method.

Testing:\
I did basic manual testing that the `theme` command now changes the background color. Doesn't look like there are any unit tests for the `UI` class so I didn't attempt to add any.

### DIFF
--- a/alot/ui.py
+++ b/alot/ui.py
@@ -647,6 +647,9 @@ class UI:
         # get the main urwid.Frame widget
         mainframe = self.root_widget.original_widget
 
+        global_att = settings.get_theming_attribute('global', 'body')
+        self.root_widget.set_attr_map({None: global_att})
+
         # body
         if self.current_buffer:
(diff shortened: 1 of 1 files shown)


pazz:
does the above commit fix the issue for you?

dylanfprice:
@pazz yes

[Here](https://github.com/pazz/alot/blob/master/alot/ui.py#L94) is where the global body colors are set in `UI.__init__`. I just did the same in `UI.update`.

pazz:
Thanks for the quick fix

